PR #1, opened 2018-11-21
Initial version of bundle-size bot

danielrozenberg:


danielrozenberg:
Missing from this PR:
* Lots of things
* Tests
* eslint rules, Travis checks, LGTM, etc...

.gitignore
```diff
@@ -0,0 +1,3 @@
+**/node_modules/**
+.env
+package-lock.json
```
rsimha: Future cleanup note: If you're using `yarn`, this file should not be created. If you're using `npm`, this file should be checked in.

bundle-size/app.js
```diff
@@ -0,0 +1,219 @@
...
+
+const {dbConnect} = require('./db');
+const path = require('path');
+
+const EMPTY_SHA = '0'.repeat(40);
+
+async function getBuildArtifactsFile(github, filename) {
+  if (filename == 'OWNERS') {
+    // Temporarily, until we add the OWNERS file to the build artifacts repo.
+    return 'choumx\ndanielrozenberg\njridgewell\nrsimha\n';
```
rsimha: Curious: Why not create an owners file right away and read from it?

bundle-size/app.js
```diff
@@ -0,0 +1,219 @@
...
+    owner: 'ampproject',
+    repo: 'amphtml-build-artifacts',
+    path: path.join('bundle-size', filename),
+  }).then(result => {
+    return Buffer.from(result.data.content, 'base64').toString();
+  });
+}
+
+function formatBundleSizeDelta(delta) {
+  return 'Δ ' + (delta >= 0 ? '+' : '') + delta.toFixed(2) + 'KB';
```
rsimha: nit: Maybe use the `∇` symbol if the bundle size has reduced?

danielrozenberg: I prefer to stick with Δ, the mathematical notation for "difference" :)

bundle-size/app.js
danielrozenberg: Will do and remove this line before submitting this PR

.gitignore
danielrozenberg: I could use help figuring out the "infra" for this, let's talk offline

PR #7, opened 2018-11-28
Rename getContent to getContents (octokit deprecation warning)

danielrozenberg:
Deprecation warning:
```
DEPRECATED (@octokit/rest): `repos.getContent()` is deprecated, use `repos.getContents()`
```


rsimha:
For posterity, add a snippet with the warning that's being fixed.

danielrozenberg:
Done

PR #13, opened 2018-12-03
Retry reports for 60 minutes when the base SHA is missing

danielrozenberg:
This PR will cause the bundle-size app to retry reporting the bundle size of a PR every minute for 10 minutes, if the base SHA (i.e., branch point from master) is missing from the `ampproject/amphtml-build-artifacts` repo

This isn't going to block the PR checks on Travis for `ampproject/amphtml`, as the retries will happen in the background. The GitHub app will return an HTTP status of `200 OK` if successful immediately, or `202 CREATED` if it's going to retry instead

### DIFF
--- a/bundle-size/app.js
+++ b/bundle-size/app.js
@@ -73,7 +73,7 @@ module.exports = app => {
     }
   }
 
-  function tryReport(retriesLeft, check, bundleSize) {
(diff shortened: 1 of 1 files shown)


danielrozenberg:
Discussed offline - we'll retry the check every minute for 60 minutes, for a total of 1 hour of checks

PR #14, opened 2018-12-03
Add configuration files for CI checks and fix code to conform

danielrozenberg:
Add configuration files for:
* ESLint
* LGTM
* Travis

Also adds a `pr-check.js` file that will, for now, execute only `eslint` (but later will also run tests)

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,46 @@
+{
+  "root": true,
+  "extends": ["eslint:recommended", "google"],
+  "plugins": [
+    "jsdoc"
+  ],
+  "parserOptions": {
+    "ecmaVersion": 2018
+  },
+  "env": {
+    "node": true
+  },
+  "settings": {
+    "jsdoc": {
+      "tagNamePreference": {
+        "returns": "return",
+        "constant": "const",
+        "class": "constructor"
+      },
+      "additionalTagNames": {
+        "customTags": [
+          "deprecated",
+          "export",
+          "final",
+          "package",
+          "restricted",
+          "suppress",
+          "template",
+          "visibleForTesting"
+        ]
+      },
+      "allowOverrideWithoutParam" : true
+    }
+  },
+  "rules": {
+    "arrow-parens": [2, "as-needed"],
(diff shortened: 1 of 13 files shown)


.eslintrc
```diff
@@ -0,0 +1,46 @@
+{
+  "root": true,
+  "extends": ["eslint:recommended", "google"],
+  "plugins": [
+    "jsdoc"
+  ],
+  "parserOptions": {
+    "ecmaVersion": 2018
```
rsimha: I'd imagine this can be `6`, since we're using ES6 code?

.travis.yml
```diff
@@ -0,0 +1,15 @@
+language: node_js
+dist: xenial
+node_js:
+  - "lts/*"
+# notifications:
+#   email:
+#     recipients:
+#       - amp-build-cop@grotations.appspotmail.com
+#     on_success: change
+#     on_failure: change
```
rsimha: These are notifications for the build cop rotation, so might as well remove.

package.json
```diff
@@ -0,0 +1,25 @@
...
+  "private": true,
+  "license": "Apache-2.0",
+  "author": "The AMP HTML authors",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/ampproject/amp-github-apps.git"
+  },
+  "engines": {
+    "node": "^10.14.1"
+  },
```
rsimha: `amphtml` currently allows all LTS versions including 8. Maybe do so here too?

See [this snippet](https://github.com/ampproject/amphtml/blob/master/package.json#L7-L10).

danielrozenberg: Since this is going to run on AppEngine, and AppEngine is currently running the latest LTS, I don't see a reason to have tests running on a lower LTS version

.eslintrc
danielrozenberg: Since this is going to run on node > 10, might as well take advantage of the latest language features :)

.travis.yml
danielrozenberg: Done

PR #17, opened 2018-12-03
Fix LGTM-bot issue with db-config.example.js

danielrozenberg:


rsimha:
For posterity, add a link to the alert.

rsimha:
Yeah :)

I only wish `amphtml-lgtm-bot` wouldn't add a new comment with the same message for each commit.

PR #18, opened 2018-12-05
bundle-size: add tests for `app.js`

### LINKED ISSUE #3: Write tests for bundle-size app

danielrozenberg:
Fixes #3

I would appreciate a thorough review, since the rest of this app is going to rely on these tests being good :)


bundle-size/setup-db.js
```diff
@@ -14,29 +14,43 @@
...
-  return db.destroy();
-});
+if (require.main === module) {
+  const db = dbConnect();
+  setupDb(db).then(() => {
+    log.info('Database table `checks` created.');
+  }).catch(error => {
+    log.error(error.message);
+  }).then(() => {
+    db.destroy();
```
erwinmombay: dumb question on my part, does this destroy the actual db or just the connection?

bundle-size/test/app.test.js
```diff
@@ -0,0 +1,371 @@
...
+        repo: 'amphtml',
+        pull_request_id: 19621,
+        installation_id: 123456,
+        check_run_id: 555555,
+        delta: null,
+      },
+    ]);
+  });
+
+  test('approved review submitted by an OWNERS person', async () => {
```
erwinmombay: I'm a little unclear on what the resulting behavior should be by these actions

bundle-size/test/app.test.js
```diff
@@ -0,0 +1,371 @@
...
+            },
+          });
+          return true;
+        })
+        .reply(200);
+
+    await probot.receive({name: 'pull_request_review', payload});
+  });
+
+  test('approved review submitted by a non-OWNERS person', async () => {
```
erwinmombay: a little unclear on what the resulting behavior should be on non-OWNER vs OWNER

bundle-size/setup-db.js
rsimha: From the line `exports.dbConnect = () => db;` above, seems like this is just a connection?

danielrozenberg: This only destroys the connection

bundle-size/test/app.test.js
danielrozenberg: I changed all the test names to be more descriptive

danielrozenberg:
Good point re skipped message. I'll make a separate PR for this change

PR #29, opened 2018-12-18
Add an OWNERS reviewer to a pull request that goes above the auto-approved bundle size

danielrozenberg:
Fixes #5

PR #33, opened 2018-12-19
Fix how tests import fixtures and wait for network expectations

danielrozenberg:
This PR fixes some issues with the tests for the bundle-size app. Specifically:
* Tests will now receive a deep copy of JSON fixtures, so that they can modify them without changing the expectations of other tests
* Explicitly wait for network expectations, since async behaviours can cause those to seem to succeed in one test A but fail on the test B, despite the failure really occurring in test A
* Split the _delayed_ test cases to two cases, since one of them now includes extra network expectations that the other doesn't (to add an OWNERS reviewer, see #29)

### DIFF
--- a/bundle-size/test/app.test.js
+++ b/bundle-size/test/app.test.js
@@ -15,16 +15,47 @@
 
 const bundleSizeApp = require('../app');
 const {dbConnect} = require('../db');
+const deepcopy = require('deepcopy');
 const nock = require('nock');
 const {Probot} = require('probot');
 const request = require('supertest');
 const {setupDb} = require('../setup-db');
+
+const nockRetryTimeoutMs = 100;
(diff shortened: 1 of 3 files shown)


bundle-size/test/app.test.js
```diff
@@ -15,17 +15,48 @@
...
 const bundleSizeApp = require('../app');
 const {dbConnect} = require('../db');
+const deepcopy = require('deepcopy');
 const nock = require('nock');
 const {Probot} = require('probot');
 const request = require('supertest');
 const {setupDb} = require('../setup-db');
 
+const nockRetryTimeoutMs = 100;
+const nockMaxTimeoutMs = 5000;
```
rsimha: I assume this is equivalent to a test timeout?

PR #34, opened 2018-12-19
Also add an OWNERS reviewer in the case of a missing base size

danielrozenberg:


bundle-size/test/app.test.js
```diff
@@ -543,6 +543,18 @@ describe('bundle-size', async () => {
...
           return true;
         })
+        .reply(200)
+        .get('/repos/ampproject/amphtml/pulls/19603/requested_reviewers')
+        .reply(200, getFixture('requested_reviewers'))
+        .post('/repos/ampproject/amphtml/pulls/19603/requested_reviewers',
+            body => {
+              expect(body).toMatchObject({
+                reviewers: [
+                  expect.stringMatching(/aghassemi|choumx|danielrozenberg/),
```
rsimha: Maybe move this list to a separate place? Or derive it from the actual owners list?

danielrozenberg: It's ok for a test to be extra verbose and explicit. This is derived from a base64-encoded list of OWNERS in a JSON fixture file. Actually deriving it in the test code means re-writing the parsing code that is being tested, so "hard coded" here is better than "test that I can copy-paste correctly" :stuck_out_tongue_closed_eyes:

PR #56, opened 2019-01-31
bundle-size: update an existing check row with new check_run_id, when the head_sha already exists

danielrozenberg:
This fixes the issue where a PR gets reported twice from GitHub (happens on occasion, not 100% sure why but I think it might have to do with how PRs can transition through a state where they were created in the UI but not yet added to the repository)

### DIFF
--- a/bundle-size/app.js
+++ b/bundle-size/app.js
@@ -231,9 +231,10 @@ module.exports = app => {
   app.on(['pull_request.opened', 'pull_request.synchronize'], async context => {
     context.log(`Pull request ${context.payload.number} created/updated`);
 
+    const headSha = context.payload.pull_request.head.sha;
     const params = context.repo({
       name: 'ampproject/bundle-size',
-      head_sha: context.payload.pull_request.head.sha,
+      head_sha: headSha,
       output: {
         title: 'Calculating new bundle size for this PR…',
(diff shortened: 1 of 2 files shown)


bundle-size/app.js
```diff
@@ -242,15 +243,29 @@ module.exports = app => {
...
+              return trx('checks').insert({
+                head_sha: headSha,
+                pull_request_id: context.payload.number,
+                installation_id: context.payload.installation.id,
+                owner: params.owner,
+                repo: params.repo,
+                check_run_id: checkRunId,
+              });
+            } else {
+              return trx('checks')
```
rsimha: Does this refresh an existing check that was initially pending?

danielrozenberg: Yes - GitHub Checks are tied to a commit SHA, so if the same SHA gets re-reported as a new pull request, it just means that the first check was abandoned. e.g.:
first report (abandoned): https://github.com/ampproject/amphtml/pull/20613/checks?check_run_id=56942658
second report (active): https://github.com/ampproject/amphtml/pull/20613/checks?check_run_id=57323551

Note that this is for the same commit SHA, but we saved only the check_run_id of the first report, so we also only reported the results back to it, despite it being outdated

PR #62, opened 2019-02-15
Update bundle-size calculation description text

danielrozenberg:
Followup from changed introduced in https://github.com/ampproject/amphtml/pull/20811

PR #63, opened 2019-02-15
Fix typo introduced in #62

danielrozenberg:


rsimha:
I just realized the terminology used by Travis is "job", not "shard". 

Totally optional to fix, since the current message is self explanatory.

PR #70, opened 2019-02-27
Encourage bundle-shrinking humans with emoji

alanorozco:
Let's celebrate when we shrink that bundle! 🎉

### DIFF
--- a/bundle-size/app.js
+++ b/bundle-size/app.js
@@ -21,6 +21,8 @@ const sleep = require('sleep-promise');
 const RETRY_MILLIS = 60000;
 const RETRY_TIMES = 60;
 
+const HUMAN_ENCOURAGEMENT_MAX_DELTA = -0.1;
+
 const db = dbConnect();
 
 /**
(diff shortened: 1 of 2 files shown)


bundle-size/app.js
```diff
@@ -68,6 +70,44 @@ function formatBundleSizeDelta(delta) {
...
 
+/**
+ * Returns an encouraging result title when the bundle size is reduced.
+ * Otherwise the title is neutral, but not discouraging.
+ *
+ * @param {number} delta bundle size delta.
+ * @param {string} deltaFormatted bundle size delta from `formatBundleSizeDelta`
+ * @return {string} successful title message.
+ */
+function successfulTitleMsg(delta, deltaFormatted) {
```
danielrozenberg: nit: successfulTitleMessage

bundle-size/app.js
```diff
@@ -68,6 +70,44 @@ function formatBundleSizeDelta(delta) {
...
+
+/**
+ * Returns an encouraging result summary when the bundle size is reduced.
+ * Otherwise the summary is neutral, but not discouraging.
+ *
+ * @param {number} delta bundle size delta.
+ * @param {string} deltaFormatted bundle size delta from `formatBundleSizeDelta`
+ * @return {string} successful summary message.
+ */
+function successfulSummaryMsg(delta, deltaFormatted) {
```
danielrozenberg: nit: successfulSummaryMessage

bundle-size/app.js
```diff
@@ -68,6 +70,44 @@ function formatBundleSizeDelta(delta) {
...
+  const sizeDesc = 'gzipped compressed size of `v0.js`';
+  const deltaDesc = `The bundle size change is ${deltaFormatted}.`;
+
+  if (delta <= HUMAN_ENCOURAGEMENT_MAX_DELTA) {
+    return `This pull request *reduces* the bundle size (${sizeDesc}), so no ` +
```
danielrozenberg: I don't mind just having that text in `sizeDesc` and `deltaDesc` repeated twice, since it's easier for readers of the source code to read this instead of jumping up and down, and I feel like readability here is more important than strict DRI

bundle-size/app.js
```diff
@@ -21,6 +21,8 @@ const sleep = require('sleep-promise');
 const RETRY_MILLIS = 60000;
 const RETRY_TIMES = 60;
 
+const HUMAN_ENCOURAGEMENT_MAX_DELTA = -0.03;
```
danielrozenberg: @choumx happy with this value?

danielrozenberg:
Also, tests need fixing

bundle-size/app.js
choumx: Sure.

choumx:
Cute idea. 🎉

PR #81, opened 2019-03-08
Remove emoji from human-encouragement title

alanorozco:
Looks like including emoji breaks the title.

image

So let's just remove it.


alanorozco:
@danielrozenberg I've no way to add reviewers and Github won't show the `undraft` button :( 

PTAL, this is ready for review.

alanorozco:
Actually I'll just reopen.

PR #82, opened 2019-03-08
Remove emoji from human-encouragement title

alanorozco:
Looks like including emoji breaks the title.

image

So let's just remove it.


danielrozenberg:
I can imagine how having "(shrunk!)" in the title can freak out contributors that don't know what this means. Let's leave the `title` without any special text - the `summary` text explains it better :)

alanorozco:
@danielrozenberg PTAL.

PR #87, opened 2019-03-12
Add gcloud project to `npm run deploy` scripts

danielrozenberg:
This will help ensure that we deploy to the correct project, since the `gcloud` tool does not allow per-directory configurations

### DIFF
--- a/bundle-size/package.json
+++ b/bundle-size/package.json
@@ -14,7 +14,7 @@
   },
   "scripts": {
     "dev": "nodemon",
-    "deploy": "gcloud config set project amp-bundle-size-bot && gcloud app deploy",
+    "deploy": "gcloud --project amp-bundle-size-bot app deploy",
     "setup-db": "node ./setup-db.js",
     "start": "probot run ./app.js",
     "test": "jest"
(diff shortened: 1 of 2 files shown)


bundle-size/package.json
```diff
@@ -14,7 +14,7 @@
   },
   "scripts": {
     "dev": "nodemon",
-    "deploy": "gcloud app deploy",
+    "deploy": "gcloud config set project amp-bundle-size-bot && gcloud app deploy",
```
rsimha: This will silently change the global default if you've set it to something else. Instead, I suggest using `--project`, which sets the project just for this invocation.

https://cloud.google.com/sdk/gcloud/reference/#--project

danielrozenberg: Done

bundle-size/package.json
```diff
@@ -14,7 +14,7 @@
   },
   "scripts": {
     "dev": "nodemon",
-    "deploy": "gcloud app deploy",
+    "deploy": "gcloud --project amp-bundle-size-bot app deploy",
```
rsimha: Does this work? According to the [docs](https://cloud.google.com/sdk/gcloud/reference/app/deploy), it appears that the `--project` option should appear after `gcloud app deploy`.

danielrozenberg: Yep, it works. I prefer to put flags before positional arguments (== opposite of how most languages accept arguments, but whatever 😝)

PR #89, opened 2019-03-12
Remove async keyword from test describes

danielrozenberg:
Test describe functions are not promises. This fixes a deprecation error that will turn into a failing error in a future version of Jest: https://travis-ci.org/ampproject/amp-github-apps/builds/505340784#L243-L244

### DIFF
--- a/bundle-size/test/app.test.js
+++ b/bundle-size/test/app.test.js
@@ -57,7 +57,7 @@ async function waitUntilNockScopeIsDone(nocks) {
   nocks.done();
 }
 
-describe('bundle-size', async () => {
+describe('bundle-size', () => {
   let probot;
   let app;
   const db = dbConnect();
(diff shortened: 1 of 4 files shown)


bundle-size/test/app.test.js
```diff
@@ -57,7 +57,7 @@ async function waitUntilNockScopeIsDone(nocks) {
   nocks.done();
 }
 
-describe('bundle-size', async () => {
+describe('bundle-size', () => {
```
rsimha: Curious: If you are removing the `async` keyword, shouldn't you also be making the test synchronous by say, returning a promise? (Assuming the tests were asynchronous to begin with.)

danielrozenberg: The tests themselves do return a promise, it's the test-suite (`describe('...', functionThatDefinesTestSuite)` that should not be

PR #99, opened 2019-03-21
Add subtype to tests slug

danielrozenberg:
As suggested in https://github.com/ampproject/amphtml/pull/21407#discussion_r267410535

Note that I broke the database convention here from snake_case to camelCase, calling this field in the database `sub_type` to calling it `subType`. I'm planning on sending another PR soon to rename all of the database fields to camelCase so that they'll match the JavaScript style for variable names. I don't know why I didn't start that way, and since we don't have any real production data in this database yet I can just drop the tables and start over easily

### DIFF
--- a/test-status/README.md
+++ b/test-status/README.md
@@ -25,10 +25,10 @@ The App has the following API points, which are to be triggered from Travis CI
 runs. If the `TRAVIS_IP_ADDRESSES` environment variable is set, only requests
 from this comma separated list of IP addresses will be processed.
 
-* `POST /v0/tests/:headSha/:type/:status(queued|started|skipped)`
+* `POST /v0/tests/:headSha/:type/:subType/:status(queued|started|skipped)`
   * Creates a new check on the supplied head commit (for `status` = `queued` or
-    `skipped`) or reports that the `type` tests have started running.
-* `POST /v0/tests/:headSha/:type/report/:passed/:failed`
+    `skipped`) or reports that the `type/subType` tests have started running.
+* `POST /v0/tests/:headSha/:type/:subType/report/:passed/:failed`
   * Updates the equivalent check with the number of `passed` and `failed` tests
   * If `failed` is 0, sets the check's conclusion to `success`, which turns the
     check green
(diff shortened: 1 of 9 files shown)


test-status/README.md
```diff
@@ -25,10 +25,10 @@ The App has the following API points, which are to be triggered from Travis CI
 runs. If the `TRAVIS_IP_ADDRESSES` environment variable is set, only requests
 from this comma separated list of IP addresses will be processed.
 
-* `POST /v0/tests/:headSha/:type/:status(queued|started|skipped)`
+* `POST /v0/tests/:headSha/:type/:subType/:status(queued|started|skipped)`
```
rsimha: nit: No colon before `status`, correct?

test-status/api.js
```diff
@@ -20,25 +20,26 @@ const {getCheckRunId, getPullRequestSnapshot} = require('./db');
...
   switch (status) {
     case 'queued':
       Object.assign(params, {
         status: 'queued',
         output: {
-          title: `${type} tests are queued on Travis`,
-          summary: `The ${type} tests are queued to run on Travis. Watch ` +
-            'this space for results in a few minutes!',
+          title: 'Tests are queued on Travis',
+          summary: `The ${type} (${subType}) tests are queued to run on ` +
```
rsimha: Grammar nit: Move the word `tests` next to `${type}` here, and in all other messages in this PR. IMO, it's better to read...
```
The unit tests (local-changes) are queued to run on Travis
```
... than...
```
The unit (local-changes) tests are queued to run on Travis
```

danielrozenberg: Done

test-status/README.md
danielrozenberg: It's how the [URL is defined in express.js](https://expressjs.com/en/guide/routing.html), `status` is the name of the variable, and `(queued|started|skipped)` are its possible values - so a valid URL would be `/v0/tests/abc123/unit/saucelabs/started` where `subType` is assigned the string `'saucelabs'` and `status` is assigned the string `'started'` - the word "status" never appears in the URL

rsimha: Aha, `:status` is a "variable" while `report` in the line below is a "constant". Never mind, carry on.

PR #100, opened 2019-03-21
Rename database fields from snake_case to camelCase

danielrozenberg:
Quick refactor

### DIFF
--- a/test-status/api.js
+++ b/test-status/api.js
@@ -30,7 +30,7 @@ function createNewCheckParams(pullRequestSnapshot, type, subType, status) {
     owner: pullRequestSnapshot.owner,
     repo: pullRequestSnapshot.repo,
     name: `ampproject/tests/${type} (${subType})`,
-    head_sha: pullRequestSnapshot.head_sha,
+    head_sha: pullRequestSnapshot.headSha,
   };
   switch (status) {
     case 'queued':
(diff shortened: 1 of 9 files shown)


test-status/web.js
```diff
@@ -142,19 +142,19 @@ exports.installWebUiRouter = (app, db) => {
     const errors = validationResult(request);
     if (!errors.isEmpty()) {
       response.render('status', Object.assign({
-        short_head_sha: request.short_head_sha,
+        shortHeadSha: request.shortHeadSha,
         is_skipping: true,
```
rsimha: Is `is_skipping` something that needs to be changed too?

test-status/api.js
```diff
@@ -30,7 +30,7 @@ function createNewCheckParams(pullRequestSnapshot, type, subType, status) {
     owner: pullRequestSnapshot.owner,
     repo: pullRequestSnapshot.repo,
     name: `ampproject/tests/${type} (${subType})`,
-    head_sha: pullRequestSnapshot.head_sha,
+    head_sha: pullRequestSnapshot.headSha,
```
rsimha: How come `head_sha` remains in snake case?

test-status/web.js
danielrozenberg: This isn't a db variable, it's a template variable, but might as well! Done.

test-status/api.js
danielrozenberg: This is for the GitHub API, which is not using the same style guide as ours. I thought about making some extra compatibility layer for a hot second and decided it's not worth the hassle 😅

PR #150, opened 2019-06-04
Project Metrics: Metric base class and dummy metric implementation

rcebulko:
The dummy implementation is in place so the infrastructure can be set up before implementing the more time-consuming real metrics which will require various API integrations.


project-metrics/metrics_service/metrics/base.py
```diff
@@ -0,0 +1,99 @@
...
+  @property
+  def name(self):
+    """Metric name to use as DB identifier when storing results."""
+    return self.__class__.__name__
+
+  @property
+  def label(self) -> Text:
+    """Label for the metric (ex. TestCoverageMetric -> "Test Coverage")."""
+    return stringcase.titlecase(
+        self.name[:-6] if self.name.endswith('Metric') else self.name)
```
danielrozenberg: Being smart about the presentation layer usually doesn't end well, this might eventually come back to bite you ;)
(but don't feel like you have to "fix" this now)

project-metrics/metrics_service/metrics/base.py
```diff
@@ -0,0 +1,99 @@
...
+  @property
+  def formatted_result(self) -> Text:
+    """The formatted result value (ex. 3w, 80%, 100PRs)."""
+    return self._format_value(self._result.value) if self._result else '?'
+
+  @property
+  def result(self) -> Optional[models.MetricResult]:
+    """The result of a computation of the metric."""
+    if self._result is None:
+      self._result = self._fetch_result()
```
danielrozenberg: This might cause unexpected behaviour down the line - if this is supposed to show a large table of values and a bunch of them are missing, that will cause a lot of background network or database activity before showing anything, right?

project-metrics/metrics_service/metrics/dummy.py
```diff
@@ -0,0 +1,36 @@
+"""Dummy/example metric.
+
+This metric is not going to stick around. It will live here temporarily as an
+example of how to implement a metric, and it will allow development of
+surrounding infrastructure (serving REST APIs, Cron job configuration, etc.).
+Once one or two real metrics are implemented, this can be removed.
+
+TODO (rcebulko): Remove once other metrics are in place.
```
danielrozenberg: You can rename this to example.py and change the documentation to explain that this is an example, and leave this file in place

project-metrics/metrics_service/metrics/base.py
rcebulko: It does the trick for any new metrics we define out-of-the-box, and if someone is unhappy with the label it can be overridden :)

project-metrics/metrics_service/metrics/dummy.py
rcebulko: Done

project-metrics/metrics_service/metrics/base.py
rcebulko: This ends up being replaced in the next PR, but for now I'll drop this part.

PR #152, opened 2019-06-04
Project Metrics: Metric database interactions and registering

rcebulko:


project-metrics/metrics_service/metrics/base.py
```diff
@@ -32,29 +40,81 @@ def label(self) -> Text:
...
     return self._score_value(
-        self._result.value) if self._result else models.MetricScore.UNKNOWN
+        self.result.value) if self.result else models.MetricScore.UNKNOWN
+
+
+class Metric(MetricDisplay):
+  """Abstract base class for health metrics."""
+
+  __metaclass__ = abc.ABCMeta
+  __active_metrics = {}
```
danielrozenberg: nit: just one underscore (here and in the method below) - two underscores have a complicated meaning in Python

project-metrics/metrics_service/metrics/base.py
```diff
@@ -32,29 +40,81 @@ def label(self) -> Text:
   @property
   def formatted_result(self) -> Text:
     """The formatted result value (ex. 3w, 80%, 100PRs)."""
-    return self._format_value(self._result.value) if self._result else '?'
-
-  @property
-  def result(self) -> Optional[models.MetricResult]:
-    """The result of a computation of the metric."""
-    return self._result
+    return self._format_value(self.result.value) if self.result else '?'
```
danielrozenberg: This wouldn't (or at least, shouldn't) pass PyType checks because `self.result` is not a defined attribute of this class (it's not defined in `__init__`) - there is a reverse expectation of inheritance here. I'm not sure what the best solution here is. Maybe don't split the classes at all - there's no reason to create a deep chain of inheriting classes, the codebase doesn't benefit from it

project-metrics/metrics_service/metrics/base.py
```diff
@@ -32,29 +40,81 @@ def label(self) -> Text:
...
+    metric_results = models.MetricResult.__table__
+    session = db_engine.get_session()
+
+    max_dates = session.query(
+        metric_results.c.name,
+        sqlalchemy.func.max(
+            metric_results.c.computed_at).label('max_computed_at')).group_by(
+                metric_results.c.name).filter(
+                    metric_results.c.name.in_(
+                        cls.__active_metrics.keys())).subquery('latest')
```
danielrozenberg: 1. Can you split `cls._active_metrics.keys` → `active_metrics_names` as temp variable for readability?
2. Some manual formatting here would go a long way towards readability - start every SQL-equivalent section in its own new line (so `.group_by`, `.filter` would start their own lines)

project-metrics/metrics_service/metrics/base.py
```diff
@@ -32,29 +40,81 @@ def label(self) -> Text:
...
+
+    latest_results = session.query(models.MetricResult).join(
+        max_dates,
+        sqlalchemy.and_(
+            metric_results.c.name == max_dates.c.name,
+            metric_results.c.computed_at == max_dates.c.max_computed_at))
```
danielrozenberg: No way of doing this entire query in one go? Or does that impact readability?

project-metrics/metrics_service/metrics/example.py
```diff
@@ -10,8 +10,8 @@
 class IdentityMetric(base.PercentageMetric):
   """A metric which reports the value it is initialized with."""
 
-  def __init__(self, value: float):
-    super(IdentityMetric, self).__init__()
+  def __init__(self, result: models.MetricResult = None, value: float = 0.5):
```
danielrozenberg: Why do these arguments have default values?

project-metrics/metrics_service/metrics/base.py
rcebulko: Done.

project-metrics/metrics_service/metrics/base.py
rcebulko: Done.

project-metrics/metrics_service/metrics/base.py
rcebulko: 1. Done.
2. Done. SQL Alchemy and Google-Format extension don't play very nice together :-1:

project-metrics/metrics_service/metrics/base.py
rcebulko: This is all done in one go, if you mean one DB round-trip query. `max_dates` just defines a subquery. `latest_results` defines the main query, which joins against the subquery table. The query itself isn't executed until we try to resolve it in the iteration below, where it's all sent as one big SELECT.

project-metrics/metrics_service/metrics/example.py
rcebulko: `result` has a default in the base class, and the `value` default was for convenience in invocation/testing. I'll remove it, doesn't add much.

PR #155, opened 2019-06-04
Project Metrics: Define table for Travis builds

rcebulko:


project-metrics/metrics_service/models.py
```diff
@@ -42,3 +42,29 @@ class MetricResult(Base):
   def __repr__(self) -> Text:
     return "<Metric(name='%s', value='%.3g', computed_at='%s')>" % (
         self.name, self.value, self.computed_at)
+
+
+class TravisState(enum.Enum):
+  """A state of a Travis build or job."""
```
danielrozenberg: Are these values taken from Travis? If so, link to docs

project-metrics/metrics_service/models.py
```diff
@@ -42,3 +42,29 @@ class MetricResult(Base):
...
+  __tablename__ = 'travis_builds'
+
+  id = sqlalchemy.Column(sqlalchemy.Integer, primary_key=True)
+  number = sqlalchemy.Column(sqlalchemy.Integer)
+  duration = sqlalchemy.Column(sqlalchemy.Integer)
+  state = sqlalchemy.Column(sqlalchemy.Enum(TravisState))
+  started_at = sqlalchemy.Column(sqlalchemy.DateTime)
+
+  def __repr__(self) -> Text:
+    return "<Build(number='%d', duration='%d', state='%s', started_at='%s')>" % (
```
danielrozenberg: Nit: none of there values are strings, so no need to surround with `' '`

project-metrics/metrics_service/models.py
rcebulko: It's not "documented" anywhere, it's just defined in the Rails model for [builds](https://github.com/travis-ci/travis-api/blob/master/lib/travis/model/build/states.rb#L25) and [jobs](https://github.com/travis-ci/travis-api/blob/master/lib/travis/model/job/test.rb#L21). Adding comment

project-metrics/metrics_service/models.py
rcebulko: Removed.

PR #158, opened 2019-06-05
Project Metrics: Implement presubmit-ignored metric

rcebulko:


project-metrics/metrics_service/metrics/presubmit_ignored.py
```diff
@@ -0,0 +1,40 @@
+"""Presubmit-Ignored metric."""
+
+import db_engine
+from metrics import base
+import models
+
+
+class PresubmitIgnoredMetric(base.Metric):
+  """A metric tracking the number of PRs that submitted with failing builds."""
```
danielrozenberg: That's not what this metric is supposed to do. This metric has to check the GitHub API for pull requests that were merged without passing the required checks, and were force submitted by an admin

rcebulko: Removed in #180

PR #159, opened 2019-06-05
Project Metrics: Implement travis-greenness metric

rcebulko:


project-metrics/metrics_service/metrics/travis_greenness.py
```diff
@@ -0,0 +1,61 @@
...
+    elif percentage < 0.75:
+      return models.MetricScore.MODERATE
+    elif percentage < 0.90:
+      return models.MetricScore.GOOD
+    else:
+      return models.MetricScore.EXCELLENT
+
+  def _count_states(self) -> Dict[models.TravisState, int]:
+    """Counts the number of builds for each relevant build state."""
+    logging.debug('Counting successful builds')
```
danielrozenberg: nit: this can be logging.info - I usually leave logging.debugging for extremely verbose stuff (e.g., if you read thousands of files and want to log each of their names)

project-metrics/metrics_service/metrics/travis_greenness.py
```diff
@@ -0,0 +1,61 @@
...
+    session = db_engine.get_session()
+    count_query = session.query(
+        models.Build.state,
+        sqlalchemy.func.count().label('state_count')
+        ).group_by(models.Build.state
+        ).filter(models.Build.state.in_([
```
danielrozenberg: If I'm not mistaken, there should also be a filter for last 90 days, right?

rcebulko: That scope helper is provided in the pruning PR, and I wanted to avoid rebasing and preventing submission of all the metric PRs while waiting for that; same logic behind cron configs. I plan to add cron configs and date filters in all metrics with one upcoming PR. I could add a TODO to all of the metric files, but I don't think that would help much since it'll be replaced ASAP and would be more effort than it's worth.

project-metrics/metrics_service/metrics/travis_greenness.py
rcebulko: Done.

PR #160, opened 2019-06-05
Project Metrics: Add module to prune old data

rcebulko:
This will be used by a cron server endpoint to periodically prune the `builds` and `releases` tables.

### DIFF
--- a/project-metrics/metrics_service/models.py
+++ b/project-metrics/metrics_service/models.py
@@ -1,12 +1,29 @@
 """Defines ORM models for interactions with the DB."""
 
-from typing import Text
+from typing import Text, Type, TypeVar
 import datetime
 import enum
 import sqlalchemy
 from sqlalchemy.ext import declarative
+import timedelta
 
 Base = declarative.declarative_base()
+
+
+def is_last_n_days(
+  timestamp_column: sqlalchemy.orm.attributes.InstrumentedAttribute, days: int
+  ) -> sqlalchemy.sql.elements.BinaryExpression:
+  """Produces the filter expression for the last N days of a model.
+
+  Args:
+    timestamp_column: model column to filter by.
+    days: number of days back to filter by.
+
+  Returns:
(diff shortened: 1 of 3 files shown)


project-metrics/metrics_service/models.py
```diff
@@ -1,14 +1,31 @@
...
+  """Produces the filter expression for the last N days of a model.
+
+  Args:
+    timestamp_column: model column to filter by.
+    days: number of days back to filter by.
+
+  Returns:
+    A binary expression which can be passed to a query's `.filter()` method.
+  """
+  n_days_ago = datetime.datetime.now() - timedelta.Timedelta(days=days)
```
danielrozenberg: 🤔 timedelta is built into Python?
`datetime.datetime.now() - datetime.timedelta(days=days)` should achieve the same effect?

project-metrics/metrics_service/pruning.py
```diff
@@ -0,0 +1,14 @@
+"""Module to provide pruning for metric data tables."""
+
+import logging
+
+import db_engine
+import models
+
+
+def prune() -> None:
```
danielrozenberg: I wouldn't merge this just yet, but keep it around. Once we have some sense of how fast the data grow, we might decide to keep it around a bit longer (so we can, say, display historic data)
Since there's no personal data involved, there's no rush to delete the data

rcebulko: Pulling the time logic out of this PR; will roll it into the upcoming PR adding time filtering to metrics. Will close this PR afterwards, not to hard to rewrite `models.Build.last_90_days(db_engine.get_session(), negate=True).delete()`

project-metrics/metrics_service/models.py
rcebulko: Didn't know `timedelta` was in `datetime`! Awesome! Will remove the dependency and use the native one when I replicate this code in the upcoming date filtering PR.

PR #161, opened 2019-06-05
Project Metrics: Implement presubmit-latency metric

rcebulko:


project-metrics/metrics_service/metrics/presubmit_latency.py
```diff
@@ -0,0 +1,45 @@
...
+import sqlalchemy
+from typing import Text
+
+import db_engine
+from metrics import base
+import models
+
+
+class PresubmitLatencyMetric(base.Metric):
+  """A metric tracking the average duration of Travis builds."""
```
danielrozenberg: This metric should only measure the Travis builds for pull requests - is that what it does or does it also measure push builds?

rcebulko: This only measures pulls (only PR builds are queried from Travis API)

rcebulko: Added clarifying comment in #180

PR #163, opened 2019-06-05
Project Metrics: Add Codecov API interface

rcebulko:
Also removes unused Travis token parameter to Travis API constructor.


project-metrics/metrics_service/apis/codecov.py
```diff
@@ -0,0 +1,59 @@
...
+    """Fetch the absolute coverage at HEAD.
+
+    Raises:
+      CodecovApiError: if the call to the Codecov API fails.
+
+    Returns:
+      Code coverage percentage in the range [0-100].
+    """
+    status_code, response = self.repo.branch.master.get(limit=1)
+    if status_code is status.HTTP_200_OK:
```
danielrozenberg: Don't use `is` to compare numbers. It only works for numbers below 256 because of a CPython implementation detail, and is not technically part of the language: https://stackoverflow.com/questions/306313/is-operator-behaves-unexpectedly-with-integers

Use `==` instead :)

project-metrics/metrics_service/apis/travis.py
```diff
@@ -39,11 +40,7 @@ def __init__(self, travis_token: Text = None, *args, **kwargs):
     self.setClient(agithub_base.Client(*args, **kwargs))
     self.setConnectionProperties(props)
 
-    if travis_token:
-      self._token = travis_token
-    else:
-      self._token = self._get_travis_token(env.get('GITHUB_API_ACCESS_TOKEN'))
-
+    self._token = self._get_travis_token(env.get('GITHUB_API_ACCESS_TOKEN'))
```
danielrozenberg: Document this change in the PR description

rcebulko: Done.

project-metrics/metrics_service/apis/codecov.py
rcebulko: Done; also updated the comparisons in `TravisApi`

PR #164, opened 2019-06-06
Project Metrics: Add filters, metric registration, and cron configs to existing metrics

rcebulko:
Since these metric implementations were written while metric registration and the cron server were pending CLs, they did not include these things in order to unblock review. This PR adds a helper scope to Builds, configures each metric, and cleans up a couple bugs introduced in a previous "Code review fixes" commit.

### DIFF
--- a/project-metrics/metrics_service/cron.yaml
+++ b/project-metrics/metrics_service/cron.yaml
@@ -10,3 +10,7 @@ cron:
 - description: "Update presubmit-latency metric."
   url: /_cron/recompute/PresubmitLatencyMetric
   schedule: every 24 hours
+
+- description: "Update travis-greenness metric."
+  url: /_cron/recompute/TravisGreennessMetric
+  schedule: every 24 hours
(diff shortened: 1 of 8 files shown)


project-metrics/metrics_service/cron.yaml
```diff
@@ -2,3 +2,15 @@ cron:
 - description: "Update dummy metric."
   url: /_cron/recompute/IdentityMetric
   schedule: every 6 hours
+
+- description: "Update presubmit-ignored metric."
```
erwinmombay: should we be consistent with naming here? you changed `Presubmit-Ignored` to `"Presubmit Ignored` below

rcebulko: In the cron descriptions, I'm using the identifiers from the design doc; in the module comments, I'm using the "prose" metric names. Seemed reasonable, but if it bothers you I can patch it.

PR #165, opened 2019-06-06
Project Metrics: Implement absolute-coverage metric

rcebulko:


project-metrics/metrics_service/metrics/absolute_coverage.py
```diff
@@ -0,0 +1,30 @@
+"""Absolute Coverage metric."""
+
+from apis import codecov
+from metrics import base
+import models
+
+
+class AbsoluteCoverageMetric(base.PercentageMetric):
+  """A metric tracking the percentage of green Travis builds."""
```
danielrozenberg: Description is incorrect

rcebulko: Fixed in #180

PR #166, opened 2019-06-06
Project Metrics: Add JSON API endpoint to server

rcebulko:


project-metrics/metrics_service/server.py
```diff
@@ -28,5 +28,16 @@ def recompute(metric_cls_name):
   return 'Successfully recomputed %s.' % metric_cls_name, status.HTTP_200_OK
 
 
+@app.route('/api/metrics')
```
danielrozenberg: super-nit (no reason to create a new PR for this) - prefixing your API endpoints with a version number (`/api/v0/metrics`) is a good habit for the future that can save you a lot of headaches. For this self-contained project, you'll probably never get blocked by this. Just a good habit to form for when you work on larger projects

project-metrics/metrics_service/server.py
```diff
@@ -28,5 +28,16 @@ def recompute(metric_cls_name):
...
+def list_metrics():
+  try:
+    results = base.Metric.get_latest()
+  except Exception as error:
```
danielrozenberg: Can you narrow this to a more specific exception/list of exceptions? (`IOError` comes to mind)

project-metrics/metrics_service/server.py
rcebulko: Noted, I noticed this in the release dashboard API. Will keep this in mind going forward.

project-metrics/metrics_service/server.py
rcebulko: I want to catch any error and return valid JSON for the client to handle

PR #169, opened 2019-06-07
Project Metrics: Tests for database modules

rcebulko:


project-metrics/metrics_service/.gcloudignore
```diff
@@ -15,5 +15,9 @@
...
 # Python pycache:
 __pycache__/
+.pytest_cache/
+
 # Ignored by the build system
-/setup.cfg
\ No newline at end of file
+/setup.cfg
+
+env.example.yaml
```
danielrozenberg: Extrasupernit: newlines at end of files 🙃

project-metrics/metrics_service/database/models.py
```diff
@@ -82,7 +82,7 @@ class Build(Base):
 
   @classmethod
   def is_last_90_days(cls):
```
danielrozenberg: I gotta admit when I first saw this my mind drifted to "I don't like that there's a hard coded 90 in this method name", but that doing a `method_missing` thing here with dynamic number parsing would be so, so much worse despite how cool it would be 😅
https://dev.to/rpalo/metaprogramming-python---method-missing

project-metrics/metrics_service/database/models_test.py
```diff
@@ -0,0 +1,68 @@
...
+  @classmethod
+  def setUpClass(cls):
+    super(TestBuild, cls).setUpClass()
+
+    engine = sqlalchemy.create_engine('sqlite:///:memory:', echo=True)
+    init_db.init_db(engine)
+    cls.Session = orm.scoped_session(orm.sessionmaker(bind=engine))
+
+    cls.FAKE_NOW = datetime.datetime(2019, 6, 1)
+    cls.FAKE_90_DAYS_AGO = cls.FAKE_NOW - datetime.timedelta(days=90)
```
danielrozenberg: I'm wondering if you really need to fix the dates in place, since they are relative

If you don't mock the `datetime.datetime.now` method and just create the db rows with the real "now" and "now - 90 days", does this fail?

project-metrics/metrics_service/.gcloudignore
rcebulko: :+1:

project-metrics/metrics_service/database/models.py
rcebulko: Yeah I wavered a bit on that too, but since pretty much all of the metrics have the 90 day limit as an integral component of their definition, it seemed reasonable.

project-metrics/metrics_service/database/models_test.py
rcebulko: Done

PR #170, opened 2019-06-07
Project Metrics: Add tests for metrics subpackage

rcebulko:


project-metrics/metrics_service/metrics/absolute_coverage_test.py
```diff
@@ -0,0 +1,50 @@
...
+from test import metric_test_case
+
+
+class TestAbsoluteCoverageMetric(metric_test_case.MetricTestCase):
+
+  def new_metric_under_test(self):
+    return absolute_coverage.AbsoluteCoverageMetric()
+
+  @mock.patch.object(
+      codecov.CodecovApi, 'get_absolute_coverage', return_value=42)
```
danielrozenberg: 🐁

project-metrics/metrics_service/test/metric_test_case.py
```diff
@@ -0,0 +1,72 @@
...
+  @abc.abstractmethod
+  def new_metric_under_test(self):
+    """Returns a new instance of the metric class under test."""
+    pass
+
+  def assertLatestResultEquals(self, result_value):
+    latest_result = base.Metric.get_latest()[self.metric.name].result
+    self.assertEqual(latest_result.value, result_value)
+
+  def assertScores(self, expected_result_scores):
```
danielrozenberg: Instead of accepting a list of value/score pairs, make this accept just one pair of value and score (each as its own argument) and rename this to something like `assertValueHasScore` and call it multiple times. This makes the tests much more explicit, and if one of them fails it'll be more clear which line it is on, rather than having to go look through the list

project-metrics/metrics_service/metrics/presubmit_ignored_test.py
```diff
@@ -0,0 +1,72 @@
...
+    super(TestPresubmitIgnoredMetric, self).tearDown()
+    session = self.Session()
+    session.query(models.Build).delete()
+    session.commit()
+
+  def new_metric_under_test(self):
+    return presubmit_ignored.PresubmitIgnoredMetric()
+
+  def testRecompute(self):
+    default_values = {
```
danielrozenberg: nit: "default" has some connotations. How about common_values/metric_values/partial_values?

project-metrics/metrics_service/metrics/presubmit_ignored_test.py
```diff
@@ -0,0 +1,72 @@
...
+        models.Build(state=models.TravisState.PASSED, **default_values),
+        models.Build(state=models.TravisState.CANCELLED, **default_values),
+        models.Build(state=models.TravisState.PASSED, **default_values),
+        models.Build(state=models.TravisState.ERRORED, **default_values),
+        models.Build(state=models.TravisState.FAILED, **default_values),
+    ])
+    session.commit()
+
+    self.metric.recompute()
+    self.assertLatestResultEquals(2)
```
danielrozenberg: Explain in a one-line comment why it should be 2

project-metrics/metrics_service/metrics/presubmit_latency_test.py
```diff
@@ -0,0 +1,72 @@
...
+        models.Build(duration=1000, **default_values),
+        models.Build(duration=2000, **default_values),
+        models.Build(duration=3000, **default_values),
+        models.Build(duration=4000, **default_values),
+        models.Build(duration=5000, **default_values),
+    ])
+    session.commit()
+
+    self.metric.recompute()
+    self.assertLatestResultEquals(3000)
```
danielrozenberg: One-line explanation

project-metrics/metrics_service/test/metric_test_case.py
```diff
@@ -0,0 +1,72 @@
...
+
+class MetricTestCase(unittest.TestCase):
+
+  __metaclass__ = abc.ABCMeta
+
+  @classmethod
+  def setUpClass(cls):
+    super(MetricTestCase, cls).setUpClass()
+
+    cls.engine = sqlalchemy.create_engine('sqlite:///:memory:', echo=True)
```
danielrozenberg: How heavy is it to create a new engine for every test (that is, in `setUp` as opposed to `setUpClass`) or to at least put the schema init/dropping in the `setUp` and `tearDown`?
Expecting the subclasses to know always do the tearDown correctly is a source of future misery for when we add a new metric 20 months from now and fail to figure out how to write tests then

project-metrics/metrics_service/test/metric_test_case.py
rcebulko: Sure

project-metrics/metrics_service/metrics/absolute_coverage_test.py
rcebulko: :dolphin:

project-metrics/metrics_service/metrics/presubmit_ignored_test.py
rcebulko: Done.

project-metrics/metrics_service/metrics/presubmit_ignored_test.py
rcebulko: Done.

project-metrics/metrics_service/metrics/presubmit_latency_test.py
rcebulko: Done.

project-metrics/metrics_service/test/metric_test_case.py
rcebulko: It approximately doubles the time to run those tests, but since it's doubling from 0.06s to 0.13s I feel like it's worth it :laughing:

PR #171, opened 2019-06-07
Project Metrics: Add tests for API modules

rcebulko:
These are not beautiful, but they make it easy to compare to the API docs and confirm that the right endpoints are being called in the right way.

### DIFF
--- a/project-metrics/metrics_service/apis/codecov_test.py
+++ b/project-metrics/metrics_service/apis/codecov_test.py
@@ -25,11 +25,7 @@ class TestCodecovApi(unittest.TestCase):
 
   def testGetAbsoluteCoverage(self):
     self.mock_request.return_value = (200, {
-        'commit': {
-            'totals': {
-                'c': '13.37'
-            }
-        }
+        'commit': {'totals': {'c': '13.37'} }
     })
     coverage = self.codecov_api.get_absolute_coverage()
 
--- a/project-metrics/metrics_service/apis/codecov_test.py
+++ b/project-metrics/metrics_service/apis/codecov_test.py
@@ -1,0 +1,46 @@
+"""Tests for Codecov API interface."""
+
+from agithub import base as agithub_base
+import unittest
+from unittest import mock
+
+import env
+from apis import codecov
+
+
+class TestCodecovApi(unittest.TestCase):
+
+  def setUp(self):
+    super(TestCodecovApi, self).setUp()
+
+    mock.patch.object(
+        env, 'get', side_effect={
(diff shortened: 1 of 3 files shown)


project-metrics/metrics_service/apis/codecov_test.py
```diff
@@ -0,0 +1,46 @@
...
+
+  def setUp(self):
+    super(TestCodecovApi, self).setUp()
+
+    mock.patch.object(
+        env, 'get', side_effect={
+            'GITHUB_REPO': '__repo__',
+        }.get).start()
+    self.mock_request = mock.patch.object(
+        agithub_base.Client, 'request', autospec=True).start()
```
danielrozenberg: You know what I think about mocking an implementation detail of a 3rd party library but alright ;)

project-metrics/metrics_service/apis/codecov_test.py
```diff
@@ -0,0 +1,46 @@
...
+    self.addCleanup(mock.patch.stopall)
+
+    self.codecov_api = codecov.CodecovApi()
+
+  def testGetAbsoluteCoverage(self):
+    self.mock_request.return_value = (200, {
+        'commit': {
+            'totals': {
+                'c': '13.37'
```
danielrozenberg: nit: comma if you use this format. Without a comma this should go in one line: `'totals': {'c': '13.37'},` ...

project-metrics/metrics_service/apis/travis_test.py
```diff
@@ -0,0 +1,87 @@
...
+    self.mock_request.assert_called_once_with(
+        travis_api.client, 'POST', '/auth/github?github_token=__github_token__',
+        mock.ANY, mock.ANY)
+
+  def testGetTravisTokenError(self):
+    self.mock_request.return_value = (403, 'Unauthorized.')
+
+    with self.assertRaisesRegex(
+        travis.TravisApiError,
+        r'Travis API Exception \(HTTP 403\): Travis Auth API request failed with response: Unauthorized\.'
```
danielrozenberg: You don't need the whole text if you want it to fit in one line ;)

project-metrics/metrics_service/apis/codecov_test.py
rcebulko: Yep yep :)

project-metrics/metrics_service/apis/codecov_test.py
rcebulko: Updated; I use the autoformatter generally, but sometimes I does stuff like this and I have to disable it :unamused:

PR #188, opened 2019-06-18
test-status: add build cop handling

danielrozenberg:
* store active build cop in db
* allow build cop updates through the API
* allow active build cop to use the web ui

### DIFF
--- a/test-status/api.js
+++ b/test-status/api.js
@@ -169,8 +169,8 @@ function createErroredCheckParams(
 }
 
 exports.installApiRouter = (app, db) => {
-  const v0 = app.route('/v0');
-  v0.use((request, response, next) => {
+  const tests = app.route('/v0/tests');
+  tests.use((request, response, next) => {
     request.app.set('trust proxy', true);
     if ('TRAVIS_IP_ADDRESSES' in process.env &&
       !process.env.TRAVIS_IP_ADDRESSES.includes(request.ip)) {
@@ -181,7 +181,7 @@ exports.installApiRouter = (app, db) => {
     }
   });
 
-  v0.post('/tests/:headSha/:type/:subType/:status(queued|started|skipped)',
+  tests.post('/:headSha/:type/:subType/:status(queued|started|skipped)',
       async (request, response) => {
         const {headSha, type, subType, status} = request.params;
         app.log(
@@ -225,7 +225,7 @@ exports.installApiRouter = (app, db) => {
         return response.end();
       });
 
-  v0.post('/tests/:headSha/:type/:subType/report/:passed/:failed',
+  tests.post('/:headSha/:type/:subType/report/:passed/:failed',
       async (request, response) => {
         const {headSha, type, subType, passed, failed} = request.params;
         app.log(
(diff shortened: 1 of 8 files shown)


test-status/db.js
```diff
@@ -83,3 +83,15 @@ exports.getCheckRunResults = async (db, headSha, type, subType) => {
...
+
+/**
+ * Get the GitHub username of the current build cop.
+ *
+ * @param {!Knex} db instantiated database connection.
+ * @return {string} the GitHub username of the current build cop.
+ */
+exports.getBuildCop = async db => {
+  return await db('buildCop')
+      .pluck('username')
```
estherkim: nit: `first('username').then(username => username)`

test-status/setup-db.js
```diff
@@ -47,6 +48,15 @@ function setupDb(db) {
     table.foreign('headSha')
         .references('pullRequestSnapshots.headSha')
         .onDelete('CASCADE');
+  }).createTable('buildCop', table => {
+    table.comment(
+        'Singleton table to store the GitHub username of active build cop');
```
estherkim: Can we ensure that this is true at the db level? Maybe by adding constraints or primary keys to the table.

test-status/api.js
```diff
@@ -114,9 +115,8 @@ function createReportedCheckParams(
           'resulted in test breakage or fix the broken tests.\n\n' +
           'If you believe that this pull request was not the cause of this ' +
           'test breakage (i.e., this is a flaky test) please contact the ' +
```
estherkim: I think we should change the messaging here so that skipping a test is the last thing to recommend, after
1. restarting the Travis job or build
2. rebasing
3. if all else fails then contact build cop to skip test or advise

test-status/api.js
```diff
@@ -279,4 +282,27 @@ exports.installApiRouter = (app, db) => {
...
+        request.body.accessToken != process.env.BUILD_COP_UPDATE_TOKEN) {
+      app.log(`Refused a request to ${request.originalUrl} from ${request.ip}`);
+      response.status(403).end('Access token missing or not authorized');
+    } else {
+      next();
+    }
+  });
+
+  buildCop.post('/update', async (request, response) => {
+    if (!('username' in request.body)) {
```
erwinmombay: what if key exists but field value is empty string? is that possible?

test-status/setup-db.js
danielrozenberg: Not sure how. It's a simple enough app that I don't worry about this breaking (and if it does, recovery is easy enough) but if you know of a way I'll be happy to implement it!

test-status/db.js
danielrozenberg: Done. It's `first('username').then(row => row['username'])` but that's still more readable than using `pluck` :)

test-status/api.js
danielrozenberg: Done. wdyt?

test-status/api.js
danielrozenberg: You're right, I added a check for this, just in case the cron job accidentally tries to update an empty string

test-status/setup-db.js
estherkim: Sure maybe overkill for now :)

test-status/api.js
estherkim: Looks awesome

PR #201, opened 2019-06-26
[pr-deploy] Create initial project & the unzip and move function

estherkim:
To deploy PR branches, we will be taking the zipped (dist) build from `gs://amp-travis-builds`, unzipping it and uploading it to the new bucket (`gs://amp-test-websites-1`, name subject to change) that will serve the static files.

This creates the initial pr-deploy project and includes the method that reads, unzips and writes. The file names are hardcoded for now.

Up next:
- Create a github app that will allow users to trigger this method.
- Add HTML files from  `examples/` and `test/manual/` in the Travis build artifact packages and replace the URLs with the newly served minified build URLs


pr-deploy/src/function/zipper.ts
```diff
@@ -0,0 +1,71 @@
...
+
+ /**
+ * @fileoverview 
+ * Google Cloud Function that takes the minified build and test fixtures from
+ * the AMP Travis Build Storage bucket, unzips it and writes to a test website bucket
+ * that serves the files publicly. Function is triggered when someone clicks 'deploy'
+ * with the amp-pr-deploy GitHub app.
+ */
+
+const unzip = require('unzip-stream');
```
erwinmombay: why are we still using require if we're using ts?

pr-deploy/package.json
```diff
@@ -0,0 +1,27 @@
...
+    "test": "jest"
+  },
+  "devDependencies": {
+    "@google-cloud/storage": "^2.5.0",
+    "@types/jest": "^24.0.15",
+    "@types/node": "^12.0.8",
+    "jest": "^24.8.0",
+    "ts-jest": "^24.0.2",
+    "typescript": "^3.5.2",
+    "unzip-stream": "^0.3.0"
```
erwinmombay: i believe we prefer to remove all `^`

pr-deploy/src/function/zipper.ts
estherkim: Ahh thank you. Didn't know why type definitions weren't loading. Changed it to `import` and now instead of running this like
```
node src/function/zipper.ts
```
you gotta do 
```
ts-node src/function/zipper.ts
```

pr-deploy/src/function/sa-pr-deploy-key.json
```diff
@@ -0,0 +1,12 @@
+{
```
danielrozenberg: This file should probably be `.gitignore`d (and a new private key generated now that it's been on the internets 🙃). For the other apps I added a "<config-file-name>.example.json" file that lists the fields and what should be written in them, and checked that file in instead.

The real config file is "our business"

pr-deploy/src/function/zipper.ts
```diff
@@ -0,0 +1,68 @@
...
+ * Google Cloud Function that takes the minified build and test fixtures from
+ * the AMP Travis Build Storage bucket, unzips it and writes to a test website bucket
+ * that serves the files publicly. Function is triggered when someone clicks 'deploy'
+ * with the amp-pr-deploy GitHub app.
+ */
+
+import unzip from 'unzip-stream';
+import {Storage} from '@google-cloud/storage';
+
+const PROJECT_ID = 'amp-travis-build-storage';
```
danielrozenberg: Can these be moved into a `.env` file? (makes it easier to run locally for development)

pr-deploy/src/function/zipper.ts
```diff
@@ -0,0 +1,68 @@
...
+
+function authenticate_() {
+  process.env.GOOGLE_APPLICATION_CREDENTIALS = PROJECT_KEY_PATH;
+}
+
+unzipAndMove();
+
+module.exports = {
+  unzipAndMove,
+}
```
danielrozenberg: Nit: newline at end of file :P

pr-deploy/src/function/zipper.ts
```diff
@@ -0,0 +1,68 @@
...
+          })
+      );  
+    });  
+};
+
+function authenticate_() {
+  process.env.GOOGLE_APPLICATION_CREDENTIALS = PROJECT_KEY_PATH;
+}
+
+unzipAndMove();
```
danielrozenberg: Should this be called on module import?

pr-deploy/test/reply.test.ts
```diff
@@ -0,0 +1,34 @@
+/**
```
danielrozenberg: I don't understand what this test is testing

pr-deploy/src/function/sa-pr-deploy-key.json
estherkim: Yeah I got lazy here... GitGuardian told me the same thing. Where do you keep the actual key value?

pr-deploy/test/reply.test.ts
estherkim: It was me testing the test framework, I'll remove

pr-deploy/src/function/zipper.ts
estherkim: Yeah I removed it, it was a stray from my local testing

estherkim:
Okiedokes 1 update, I added eslint for typescript. For the vscode eslint extension, you'll have update the settings to this. PTAL!
```
"eslint.validate": [
  "javascript",
  {
    "language": "typescript",
    "autoFix": true
  }
]
```

pr-deploy/src/zipper.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  return new Promise<void>((resolve, reject) => {
+    buildFile.createReadStream()
+      .pipe(unzip.Parse())
+      .on('entry', entry => {
+        const servePath = serveDir + entry.path;
+        const serveFile = serveBucket.file(servePath);
+        entry.pipe(serveFile.createWriteStream()
+          .on('error', error => {
+            console.log(error);
```
danielrozenberg: Nit: in other code we use a library for logging instead of `console.log`, not sure why but maybe @erwinmombay / @rsimha have insights on why that would be and if it's important here?

rsimha: The library we use is `fancy-log`, and it adds timestamps to logging statements.
https://github.com/ampproject/amp-github-apps/blob/06839c8ac5c036fcf6269ee53e8bb277eb2737ab/bundle-size/setup-db.js#L18

PR #214, opened 2019-07-02
[pr-deploy] Create initial probot app and unit tests

estherkim:
The PR Deploy app works like this: 
1. Creates a check with a `queued` status when a pull request is opened
2. Listens to requests on `pr-deploy/owners/:owner/repos/:repo/headshas/:headSha`. When the dist job on Travis is finished and the build package is uploaded, it will send a request to this URL. When a request is received, the app will set the check to `completed`, which enables the `Deploy Me!` action.
3. When a user clicks on the `Deploy Me!` button, the app sets the check to `in_progress` while it unzips and moves the files to the test serving bucket. The button is removed.
4. When the files are ready, the app sets the check to `completed` and updates the output to include the test site URL.

This PR doesn't implement everything and I didn't deploy the app yet, but it mostly implements the 4 steps.


pr-deploy/test/app.test.ts
```diff
@@ -0,0 +1,116 @@
...
+  let server: express.Application;
+
+  beforeEach(() => {
+    github = {
+      checks: {
+        create: jest.fn().mockReturnValue(Promise.resolve()),
+        update: jest.fn().mockReturnValue(Promise.resolve()),
+        listForRef: jest.fn().mockImplementation(
+          params => {
+            if (params.repo == 'newRepo') {
```
danielrozenberg: Adding "if" statements in tests makes them much more difficult to verify in code review, because the reviewer has to simulate the runtime inside their head

My suggestion is that in the `beforeEach` just set `listForRef: jest.fn(),` without the implementation. Then in each individual test case do `github.listForRef.mockReturnValue(…)` and at the end of those tests `expect(github.listForRef).toHaveBeenCalledWith(expect.objectContaining({repo: 'newRepo'}))` (or `toHaveBeenCalledWith({repo: 'newRepo'}) if that's the whole `params` object, or whatever else is relevant for that function)

pr-deploy/test/app.test.ts
```diff
@@ -0,0 +1,116 @@
...
+
+      await app.receive(prSynchronizedEvent);
+      expect(github.checks.update).toHaveBeenCalled();
+      expect(github.checks.create).not.toHaveBeenCalled();
+    });
+
+  test('enables deployment action when post is received', async done => {
+    request(server)
+      .post('/pr-deploy/owners/1/repos/2/headshas/3')
+      .expect(() => { expect(github.checks.update).toHaveBeenCalledTimes(1);})
```
danielrozenberg: Why not have that as a separate line after the `request();`? 🤔

pr-deploy/src/app.ts
```diff
@@ -0,0 +1,91 @@
...
+  });
+}
+
+/**
+ * Listens to the HTTP post route from Travis to know when dist upload is complete.
+ * When received, updates the PR check status to 'complete'
+ * so that the check run action to deploy site is enabled.
+ */
+function initializeRouter(app: Application) {
+  const router: IRouter<void> = app.route('/pr-deploy');
```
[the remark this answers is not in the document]
estherkim: Good call! Changed to `v0/pr-deploy` to follow the same pattern as `v0/tests` and `v0/build-cop`

pr-deploy/test/app.test.ts
estherkim: Makes sense :)

pr-deploy/test/app.test.ts
estherkim: Yeah it's not intuitive! The assertion fails if you do it after the request. It's something with how  `request()` from `supertest` works.

pr-deploy/src/app.ts
```diff
@@ -0,0 +1,91 @@
...
+  router.use(require('express').json());
+  router.post('/owners/:owner/repos/:repo/headshas/:headSha',
```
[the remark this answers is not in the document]
estherkim: I tried it but the test failed and I don't know how to fix~~~ :desert_island: :desert_island: :desert_island:

estherkim:
Thanks @danielrozenberg for the awesome review :heart_eyes:

pr-deploy/tsconfig.json
```diff
@@ -10,12 +10,13 @@
...
       "*": [
-        "node_modules/*",
+        "node_modules/*",        
+        "node_modules/@types/*",
         "src/types/*",
       ]
     }
   },
-  "include": [
+  "include": [
```
danielrozenberg: Extra white-space 😬

PR #218, opened 2019-07-03
Project Metrics: Migrate to GraphQL Travis API

rcebulko:


project-metrics/metrics_service/apis/travis.py
```diff
@@ -26,22 +28,36 @@ def __init__(self, status_code: int, err_msg: Text):
 class TravisApi(agithub_base.API):
   """Travis API interface."""
 
-  def __init__(self, *args, **kwargs):
+  # Default/maximum page size from the API.
+  TRAVIS_PAGE_SIZE = 25
```
danielrozenberg: Place this here makes this an instance field, which is not necessarily what you want. There's some odd behaviours around this (most of them are irrelevant to this PR, but it's best to just place this at the module level)

Examples of unexpected behaviours:

```python
class A(object):
  F = 1

print(A.F)  # "1"

a1 = A()
a2 = A()

a1.F = 2
print(A.F, a1.F, a2.F)  # "1 2 1"

A.F = 3
print(A.F, a1.F, a2.F)  # "3 2 3"
```

project-metrics/metrics_service/apis/travis.py
```diff
@@ -26,22 +28,36 @@ def __init__(self, status_code: int, err_msg: Text):
...
-    self._token = self._get_travis_token(env.get('GITHUB_API_ACCESS_TOKEN'))
+    try:
+      self.api_token = env.get('TRAVIS_API_ACCESS_TOKEN')
+    except KeyError:
+      logging.warn('No Travis API key found; exchanging GitHub token')
+      self.api_token = self._get_travis_token(
+          env.get('GITHUB_API_ACCESS_TOKEN'))
+
     extra_headers['Authorization'] = 'token %s' % self._token
+    extra_headers['Travis-API-Version'] = '3'
```
danielrozenberg: Can add this where you define the `extra_headers`

project-metrics/metrics_service/apis/travis.py
```diff
@@ -26,22 +28,36 @@ def __init__(self, status_code: int, err_msg: Text):
...
 
-    self._token = self._get_travis_token(env.get('GITHUB_API_ACCESS_TOKEN'))
+    try:
+      self.api_token = env.get('TRAVIS_API_ACCESS_TOKEN')
+    except KeyError:
+      logging.warn('No Travis API key found; exchanging GitHub token')
+      self.api_token = self._get_travis_token(
+          env.get('GITHUB_API_ACCESS_TOKEN'))
+
     extra_headers['Authorization'] = 'token %s' % self._token
```
danielrozenberg: Move the whole block that gets the token above where you define `extra_headers`, and add the `Authorization` header as a field in the initializer of `extra_headers`

If you can initialize the whole dict in one go, it makes this a little more readable and makes procedural sense:

`"get info required to init the headers" → "init the headers" → "use headers to init the connection"`

compared with

`"init partial headers" → "get info required to complete the headers" → "complete the headers" → "use headers to init the connection"`

project-metrics/metrics_service/apis/travis.py
```diff
@@ -69,13 +85,25 @@ def _get_travis_token(self, github_token: Text) -> Text:
   @property
   def repo(self) -> agithub_base.IncompleteRequest:
     """Returns a partial Travis request for the repository in env.yaml."""
-    return self.repos[env.get('GITHUB_REPO')]
-
-  def fetch_builds(self, after_number: int = None) -> Dict[Text, Any]:
+    return self['repo'][env.get('GITHUB_REPO').replace('/', '%2F')]
+
```
danielrozenberg: Can you use `encodeURIComponent` here?

project-metrics/metrics_service/database/models.py
```diff
@@ -66,20 +66,25 @@ class TravisState(enum.Enum):
   Based on
   https://github.com/travis-ci/travis-api/blob/master/lib/travis/model/job/test.rb#L21
```
danielrozenberg: Still true?

project-metrics/metrics_service/apis/travis.py
rcebulko: Nope, the `auth` API is V2, so this gets added after the credential exchange. That said, it seems like once you exchange a GitHub token for a Travis token, the Travis token doesn't change. So, it may even be worth removing all this token exchange logic, requiring a Travis token in the YAML, and providing the CLI command to generate a Travis token from a GitHub token. Thoughts?

project-metrics/metrics_service/apis/travis.py
rcebulko: Can't do that for the same reason as above; the V2 API query requires that the client be working/set-up, with headers, then a slightly different set of headers are used to make GraphQL queries.

project-metrics/metrics_service/apis/travis.py
rcebulko: I could! That seemed heavy-weight given the known format of the repo string, since all other characters allowed in the slug don't need any encoding. I don't feel too strongly either way

project-metrics/metrics_service/database/models.py
rcebulko: Updated to the `build` states instead of  `job` states.

rcebulko: (in workspace, waiting to commit until other PR fixes are taken care of)

project-metrics/metrics_service/apis/travis.py
rcebulko: Done

project-metrics/metrics_service/apis/travis.py
danielrozenberg: Magic strings aren't all that bad, and I've been known to use them myself. I think for this case it just makes it a little bit more readable to use a pre-existing function.

"quoting text" has a long history in web for meaning what "escape" usually means in other places, so that's fine. Turns out I was thinking of JS (`encodeURIComponent` is a built-in JS function), so you can't use that here 😝

Yes, use `urllib`…

project-metrics/metrics_service/apis/travis.py
danielrozenberg: Ah I see. Add a comment for this then :)

project-metrics/metrics_service/apis/travis.py
danielrozenberg: Hmm... Yeah? I think that makes sense

In your summary doc write something about how the Travis token was auto-generated and doesn't seem to change, but if it does "here's the code I had before to auto-generate it on the fly"

project-metrics/metrics_service/apis/travis.py
rcebulko: Yeah I'm familiar with the JS function and I think the naming is way better/clearer than `quote`. I'll swap this out then.

project-metrics/metrics_service/apis/travis.py
rcebulko: Done

project-metrics/metrics_service/apis/travis.py
rcebulko: Done

PR #230, opened 2019-07-10
Project Metrics: Add scraping of commits from GitHub

rcebulko:


project-metrics/metrics_service/apis/github.py
```diff
@@ -14,15 +14,16 @@
 GraphQL = Text
 JsonDict = Dict[Text, Any]
 
-PAGING_INFO = '''
+MAX_PAGE_SIZE = 100
```
rcebulko: This is the only change; I let the auto-formatter have its way.

project-metrics/metrics_service/database/db.py
```diff
@@ -21,7 +21,7 @@ def get_engine() -> sqlalchemy.engine.Engine:
               'charset':
                   'utf8mb4',
           }),
-      echo=True)
+        echo=True)
```
rcebulko: Auto-formatter :stuck_out_tongue:

project-metrics/metrics_service/database/models.py
```diff
@@ -120,11 +120,12 @@ def __repr__(self) -> Text:
                                    self.commit_hash)
 
 
-class PullRequestStatus(enum.Enum):
+class CommitStatus(enum.Enum):
```
danielrozenberg: Why is this renamed?

project-metrics/metrics_service/server.py
```diff
@@ -4,27 +4,46 @@
...
+    metric_cls = base.Metric.get_metric(metric_cls_name)
   except KeyError:
+    logging.error('No active metric found for %s.', metric_cls_name)
     return ('No active metric found for %s.' % metric_cls_name,
             status.HTTP_404_NOT_FOUND)
-  metric_cls().recompute()
+  logging.info('Recomputing %s.', metric_cls_name)
+  try:
+    metric_cls().recompute()
+  except Exception as error:
```
danielrozenberg: Can you limit this? Catching `Exception` means also catching `SystemExit`, for example...

rcebulko: The SystemExit exception inherits from BaseException instead of StandardError or Exception so that it is not accidentally caught by code that catches Exception. This try/except allows the cron to catch any type of error, log it appropriately, and return the proper response/status.

project-metrics/metrics_service/database/models.py
rcebulko: It _was_ renamed because it seemed like commits had the status, not the PR; but as determined offline that's a whole separate data model/API endpoint. I'll revert this.

PR #234, opened 2019-07-11
Project Metrics: Add scraping of Travis builds

rcebulko:


project-metrics/metrics_service/database/commit_scraper.py
```diff
@@ -104,7 +104,6 @@ def scrape_since_latest(self):
         db.get_engine().execute(
             models.Commit.__table__.insert().prefix_with('IGNORE'),
             commit_dicts)
-        self.session.commit()
```
danielrozenberg: ?

rcebulko: Commit message was "Remove unnecessary session.commit()". I wasn't using the session to execute, I was using the engine directly; this was an artifact from an alternative approach.

project-metrics/metrics_service/apis/travis.py
```diff
@@ -73,8 +74,7 @@ def fetch_builds(self, page_num: int = 0) -> Sequence[models.Build]:
       TravisApiError: if the call to the Travis Builds API fails.
 
     Returns:
-      The API response containing a `builds` list and associated `commits` list.
-      See https://docs.travis-ci.com/api/#builds
+      A list of the fetched builds.
     """
     params = {
         'event_type': 'pull_request',
```
rsimha: This should have been `push` and not `pull_request`. The goal of measuring the Travis Greenness metric is to keep track of the state of `master` and make sure most if not all builds pass. (Meanwhile, it's normal for a large percentage of `pull_request` builds to fail.)

PR #238, opened 2019-07-12
Project Metrics: Add API endpoint for Shields.io badges

rcebulko:


project-metrics/metrics_service/server.py
```diff
@@ -61,6 +61,36 @@ def list_metrics():
...
+  try:
+    metric = base.Metric.get_latest()[metric_cls_name]
+    response['color'] = [
+        '#EEEEEE',
+        'indianred',
+        'orange',
+        'greenyellow',
+        'lawngreen',
+        'forestgreen',
+    ][metric.score.value]
```
danielrozenberg: Put the colour values in a const at the top of the files and reference it here

rcebulko: Done.

PR #240, opened 2019-07-15
Project Metrics: Implement travis-flakiness metric

rcebulko:


project-metrics/metrics_service/metrics/travis_flakiness.py
```diff
@@ -0,0 +1,68 @@
...
+    if build_count == 0:
+      raise ValueError('No Travis builds to process.')
+    if build_count < 3:
+      raise ValueError('Not enough Travis builds to determine flakiness.')
+
+    flakes = 0
+    build_triples = zip(builds[:-2], builds[1:-1], builds[2:])
+    for prev_build, curr_build, next_build in build_triples:
+      if (curr_build.state != models.TravisState.PASSED and
+          prev_build.state == models.TravisState.PASSED and
```
danielrozenberg: Organize the conditions prev→curr→next, easier to read

project-metrics/metrics_service/metrics/travis_flakiness.py
```diff
@@ -0,0 +1,68 @@
...
+          next_build.state == models.TravisState.PASSED):
+        flakes += 1
```
danielrozenberg: If you reaaaaaaaaaaaaaaaally want to write "smart looking" code here (not required), make that if expression a separate function:

```python
_is_flaky(build_triples):
  prev_build, curr_build, next_build = build_triples
  return (prev_build.state == models.TravisState.PASSED and
      curr_build.state != models.TravisState.PASSED and
      next_build.state == models.TravisState.PASSED)
```

And then replace the for loop with:
```python
  flakes = sum(_is_flaky(build_triples) for build_triple in build_triples)
  # explicit conversion from True/False to 1/0 makes the sum correct
```

rcebulko: Yeah I thought about this but considered it to be over-engineering.

project-metrics/metrics_service/metrics/travis_flakiness.py
rcebulko: Done

PR #243, opened 2019-07-15
Project Metrics: Use wall-clock time for Travis build duration

rcebulko:
Also fixes scoring for presubmit-latency

### DIFF
--- a/project-metrics/metrics_service/apis/travis.py
+++ b/project-metrics/metrics_service/apis/travis.py
@@ -55,11 +55,15 @@ class TravisApi(agithub_base.API):
     started_at = datetime.datetime.strptime(
         json_build['started_at'],
         '%Y-%m-%dT%H:%M:%SZ') if json_build['started_at'] else None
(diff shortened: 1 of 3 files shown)


project-metrics/metrics_service/apis/travis.py
```diff
@@ -55,11 +55,15 @@ def _json_to_build(self, json_build: Dict[Text, Any]) -> models.Build:
     started_at = datetime.datetime.strptime(
         json_build['started_at'],
         '%Y-%m-%dT%H:%M:%SZ') if json_build['started_at'] else None
+    finished_at = datetime.datetime.strptime(
+        json_build['finished_at'],
+        '%Y-%m-%dT%H:%M:%SZ') if json_build['finished_at'] else None
+    duration = finished_at - started_at if started_at and finished_at else None
```
danielrozenberg: A bit confusing... I assume it's equivalent to `(finished_at - started_at) if (started_at and finished_at) else None`? Add these parenthesis to clarify

PR #249, opened 2019-07-15
Project Metrics: Add cherry-pick data model and scraping

rcebulko:


project-metrics/metrics_service/database/cherrypick_scraper.py
```diff
@@ -0,0 +1,115 @@
...
+import time
+import sqlalchemy
+
+from apis import github
+from database import db
+from database import models
+
+SCRAPE_INTERVAL_SECONDS = 5
+CHERRYPICK_ISSUE_TITLE_REGEX = (
+    r'^🌸?\s*Cherry[ -]?pick(?: request)?(?: for)? #?(\d+).*$')
```
danielrozenberg: I'm having some issues with this solution, because it counts the number of requests, not the actual number of cherry-picks. There are also cases where an approved cherry-pick requests creates 2 or 3 cherry picks, on 2 different releases (for the existing canary and prod, for example)

I think we can't get away from doing git tree analysis here:
1. start with all the release tags - they're easy to find - all the tags containing exactly 13 digits (that can be parsed as `YYMMDDHHMMSSX` where X is any digit, if you really want to get technical)
2. For every release tag, find its SHA: `git rev-list -n 1 [[RTV]]` → SHA(rtv)
2. For every release, find where they were split off from canary: `git merge-base [[SHA(rtv)]] origin/canary` → SHA(base)
3. Collect all the commits in between the two: `git rev-list [[SHA(base)]]..[[SHA(rtv)]]` → returns a list of SHAs

Collect all the returned lists (if this was not a cherry-pick, the list will be empty) and dedup repeated SHAs - the total number of distinct SHAs is the number of cherry-picks

e.g., release 1907022322580 has a list of 6 cherry picks, but it's really 3 cherry picks on top of release 1906282133370, which itself is 3 cherry picks on top of canary

PR #255, opened 2019-07-18
Project Metrics: Scrape cherry-picks for each release

rcebulko:


project-metrics/metrics_service/database/cherrypick_scraper.py
```diff
@@ -0,0 +1,75 @@
...
+    Returns:
+      A list of cherrypicks
+    """
+    logging.info('Querying GitHub for commits in release "%s"', release.name)
+    response = self.github.query_repo("""
+      release(tagName: "{release_name}") {{
+        tag {{
+          target {{
+            ... on Commit {{
+              history(first: 20) {{
```
danielrozenberg: Magic number?

rcebulko: Just a (very) loose upper bound on how many commits to consider when finding cherrypicks. Since this is more than the total number of cherrypicks in the last 90 days, it seems unlikely there will every be 20 cherrypicks in one release

danielrozenberg: Fair, but might still be worth documenting it in code and maybe making it a const too

PR #260, opened 2019-07-18
Project Metrics: Enable support for metrics to compute historical metric results

rcebulko:
The accuracy of results will depend on historical data being present, which is yet to be scraped. Historical metric computation will be a one-off backfill once the necessary historical data is scraped.

### DIFF
--- a/project-metrics/metrics_service/database/models.py
+++ b/project-metrics/metrics_service/database/models.py
@@ -1,6 +1,6 @@
 """Defines ORM models for interactions with the DB."""
 
-from typing import Text
+from typing import Optional, Text
 import datetime
 import enum
 import sqlalchemy
--- a/project-metrics/metrics_service/database/models.py
+++ b/project-metrics/metrics_service/database/models.py
@@ -12,21 +12,21 @@ Base = declarative.declarative_base()
 def _is_last_n_days(
     timestamp_column: sqlalchemy.orm.attributes.InstrumentedAttribute,
     days: int,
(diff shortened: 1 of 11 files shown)


project-metrics/metrics_service/database/models.py
```diff
@@ -11,17 +11,22 @@
 
 def _is_last_n_days(
     timestamp_column: sqlalchemy.orm.attributes.InstrumentedAttribute,
-    days: int) -> sqlalchemy.sql.elements.BinaryExpression:
+    days: int,
+    now: Optional[datetime.datetime] = None
```
danielrozenberg: Maybe call it start_time? base_time?

project-metrics/metrics_service/metrics/base.py
```diff
@@ -90,8 +91,11 @@ def get_latest(cls) -> Dict[Text, MetricImplementation]:
   def get_metric(cls, metric_cls_name) -> models.MetricResult:
     return cls._active_metrics[metric_cls_name]
 
-  def __init__(self, result: Optional[models.MetricResult] = None):
+  def __init__(self,
+               result: Optional[models.MetricResult] = None,
+               now: Optional[datetime.datetime] = None):
     self.result = result
+    self.now = now or datetime.datetime.now()
```
danielrozenberg: Rename here too

project-metrics/metrics_service/database/models.py
rcebulko: Done.

PR #262, opened 2019-07-19
Project Metrics: Add support for scraping historical data

rcebulko:


project-metrics/metrics_service/scrapers/release_scraper.py
```diff
@@ -9,6 +10,7 @@
 from database import models
 
 SCRAPE_INTERVAL_SECONDS = 5
+RELEASE_NAME_REGEX = r'^\d{13}$'
```
danielrozenberg: Super-nit: `re.match(r'^something') == re.match(r'something')` - that is, if you use `re.match` you don't need to use `^`
(in Python 3 there's `re.fullmatch` which also implicitly adds the `$` at the end, but not in Python 2...)

rcebulko: Yeah I'm aware, but since it doesn't include the $ and is non-obvious I opted to include the `^` for clarity. Presumably the `re` module ignores it anyway.

PR #272, opened 2019-07-22
Project Metrics: Plot historical result values

rcebulko:


project-metrics/metrics_service/metric_plot.py
```diff
@@ -0,0 +1,80 @@
...
+    self.history = self._get_metric_history()
+
+  def __del__(self):
+    plt.clf()
+
+  def plot_metric_history(self, width=4, height=2) -> io.BytesIO:
+    """Plots 6 months of metric results.
+
+    Args:
+      width: plot width in inches.
```
danielrozenberg: inches???? 😵

project-metrics/metrics_service/metric_plot.py
```diff
@@ -0,0 +1,80 @@
...
+    history = session.query(models.MetricResult).order_by(
+        models.MetricResult.computed_at.asc()).filter(
+            (models.MetricResult.name == self.metric.name)
+            & (models.MetricResult.computed_at > start_date)).all()
+
+    session.close()
+    return history
+
+  def _make_plot(self):
+    result_dates = [result.computed_at for result in self.history]
```
danielrozenberg: Do these need to be lists? When dealing with buffers and lots of data it's worth trying to make these generators (`(…)`) instead of lists (`[…]`)

project-metrics/metrics_service/server.py
```diff
@@ -26,20 +30,51 @@
...
+    content_type: content type of the file
+  """
+  _get_cloud_blob(filename).upload_from_string(data, content_type=content_type)
+
+
+def _get_from_cloud(filename: Text) -> bytes:
+  """Download data from a Google Cloud Storage blob.
+
+  Args:
+    filename: key under which the file in the Cloud Storage bucket is stored\
```
danielrozenberg: Stray \

project-metrics/metrics_service/metric_plot.py
rcebulko: `matplotlib` is weird.

project-metrics/metrics_service/server.py
rcebulko: Whoops

project-metrics/metrics_service/metric_plot.py
rcebulko: Pretty sure matplotlib requires lists/converts generators to lists internally, as all docs talk in terms of sequences not iterables.

PR #289, opened 2019-07-23
[pr-deploy] Update POST route and downgrade to nodejs10 for App Engine standard env

estherkim:
Updates the route URL to `travisbuilds` because that's how zipped build artifacts are named, `amp_dist_${travisBuildNumber}`.

Downgrades to nodejs10 because App Engine's standard environment only supports specific versions of nodejs, unlike the flexible environment which can support any of them. Opted for using the standard environment because it's :heavy_dollar_sign: free.99

Moved repo `owner` and `repo` to env variables because they're the same every time.

### DIFF
--- a/pr-deploy/src/github.ts
+++ b/pr-deploy/src/github.ts
@@ -23,9 +23,10 @@ const ACTION: Octokit.ChecksUpdateParamsActions = {
   description: 'Serves the minified output of this PR.',
   identifier: 'deploy-me-action',
 };
-const CHECK_NAME: string = 'pr-deploy-check';
-const OWNER = process.env.GH_OWNER;
-const REPO = process.env.GH_REPO;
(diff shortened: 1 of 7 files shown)


pr-deploy/src/github.ts
```diff
@@ -24,19 +24,17 @@ const ACTION: Octokit.ChecksUpdateParamsActions = {
   identifier: 'deploy-me-action',
 };
 
-const CHECK_NAME: string = 'pr-deploy-check';
+const check_name = process.env.GH_CHECK;
```
danielrozenberg: Update the readme/example files with these new env variables

estherkim: K added an `.env.example` file, let me update the README entirely in another PR

Issue #316, opened 2019-08-08
Enable prettier for the `amp-github-apps` repo


rsimha:
For consistency, let's use the same `.prettierrc` as `amphtml`.

rsimha:
@rcebulko Reopening this since the code doesn't actually follow the format recommended by prettier. (You'd need to run `eslint` with `--fix` and change the `prettier` entry in `.eslintrc` from a warning to an error.)

rcebulko:
@rsimha On the Prettier PR @erwinmombay suggested making it a warning rather than an error. I'll spin up a linted PR with the changes and loop you both in as reviewers.

PR #328, opened 2019-08-13
Remove conflicts between Prettier and eslint + unify OWNERS bot rules

rcebulko:
Part 1 of 2 to make Prettier/linting a blocking check.

Follow-up PR will contain all linting fixes and will make Prettier an error/blocking check. Wanted to separate this out from all the actual code fixes.

This adds `eslint-config-prettier` to the repo and eliminates rules that conflict between `eslint` and `Prettier`. The OWNERS bot is able to inherit most of the `.eslintrc` config, so it seemed sensible to standardize the rules and let it inherit the rules for the main repo.

Also adds `npm run (start|lint)` as scripts in `package.json`.

Associated with #316


.eslintrc
```diff
@@ -36,8 +40,6 @@
     }
   },
   "rules": {
-    "arrow-parens": [2, "as-needed"],
-    "indent": [2, 2, { "SwitchCase": 1, "VariableDeclarator": 2, "MemberExpression": 2, "ObjectExpression": 1, "CallExpression": { "arguments": 2 } }],
```
rcebulko: These rules are controlled by Prettier. Since we use the same Prettier config as amphtml, I imagine we're probably satisfied with the results. However, maybe we want to copy the `.eslintrc` rules from amphtml over as well? I'll run `--fix` once with and once without the amphtml ruleset, and will take a look at the diff.

owners/.eslintrc
```diff
@@ -1,4 +1,13 @@
 {
+  "extends": [
+    "eslint:recommended",
+    "google",
+    "prettier"
+  ],
+  "plugins": [
+    "jsdoc",
+    "prettier"
+  ],
```
rcebulko: These do not inherit from the root.

owners/.eslintrc
```diff
@@ -37,67 +46,5 @@
...
-    "quotes": [2, "single", "avoid-escape"],
-    "radix": 2,
-    "semi": 2,
-    "space-before-blocks": 2,
-    "space-before-function-paren": [2, "never"],
-    "space-in-parens": 2,
-    "space-infix-ops": 2,
-    "space-unary-ops": [1, { "words": true, "nonwords": false }],
-    "keyword-spacing": 2,
-    "wrap-iife": [2, "any"]
```
rcebulko: Will confirm which of these are/are not handled as desired by Prettier.

.eslintrc
rsimha: I can think of several `amphtml` lint rules that we likely don't care about here. For example, the JSDoc rules that we primarily need for closure compiler type checking and optimization. Bringing in the formatting rules SGTM, though.

package.json
```diff
@@ -13,7 +13,9 @@
     "node": "^10.14.1"
   },
   "scripts": {
-    "start": "node build-system/pr-check.js"
+    "start": "node build-system/pr-check.js",
+    "lint": "eslint .",
+    "fix": "eslint . --fix"
```
rsimha: This change SGTM assuming we're good with `npm` scripts in this repo. (`amphtml` uses `gulp`, primarily because of a heavy reliance on its file streaming functionality.)

/cc @ampproject/wg-infra

estherkim: I'm ok with sticking with npm scripts.

rcebulko: Cool, resolving.

owners/.eslintrc
rcebulko: Relevant rules now live in the root `.eslintrc`; resolving.

rcebulko:
Update: Reworked to remove all JSDoc linting/plugin, added ampproject/amphtml linting rules, and made linting (temporarily) non-blocking. In a follow up PR, will fix all lint warnings/errors and make them blocking going forward.

@danielrozenberg @rsimha @estherkim A sanity check would be most appreciated :)

danielrozenberg:
Looks sane to me

PR #337, opened 2019-08-15
Progressively isolate OWNERS bot functionality and add specs

rcebulko:
Since the OWNERS bot development was spread out and from a little while ago, there's not very comprehensive tests at the moment and it's hard for me to have confidence in which parts of the system work/fail under which conditions and corner cases. This PR is a WIP gradually collecting pieces of functionality into dedicated classes and writing specs around them.

You will find this easiest to review commit-by-commit (`n` and `p` shortcuts can make this a bit quicker), as it should be possible to verify that each individual commit is a no-op. Feel free to review only a sub-range of the commits, since there are quite a few!

When reviewing added classes, start with the specs, as it will make it clear how they are/can be used.

## Before

```
|----------------|----------|----------|----------|----------|-------------------|
| File           |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
|----------------|----------|----------|----------|----------|-------------------|
| All files      |    69.54 |    42.86 |    62.07 |    69.74 |                   |
|  owners        |      100 |      100 |      100 |      100 |                   |
|   index.js     |      100 |      100 |      100 |      100 |                   |
|  owners/src    |    66.85 |    39.39 |    58.49 |    67.04 |                   |
|   git.js       |    28.13 |        0 |       10 |    29.03 |... 95,108,110,119 |
|   github.js    |    82.86 |       75 |    75.76 |    82.69 |... 73,385,386,394 |
|   owner.js     |     52.5 |       20 |    44.44 |     52.5 |... 35,137,138,140 |
|   repo-file.js |      100 |       50 |      100 |      100 |                29 |
|----------------|----------|----------|----------|----------|-------------------|
```

- Test Coverage: 70%
- Test Suites:   1 passed, 1 total
- Tests:         8 passed, 8 total
- Time:          33.502s, estimated 34s

## After

```
|-------------------|----------|----------|----------|----------|-------------------|
| File              |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
|-------------------|----------|----------|----------|----------|-------------------|
| All files         |    89.83 |    77.42 |    90.79 |    89.43 |                   |
|  owners           |      100 |       75 |      100 |      100 |                   |
|   index.js        |      100 |       75 |      100 |      100 |                34 |
|  owners/src       |    88.29 |    77.59 |    90.14 |    87.76 |                   |
|   github.js       |      100 |      100 |      100 |      100 |                   |
|   local_repo.js   |      100 |      100 |      100 |      100 |                   |
|   owner.js        |    51.02 |       20 |    41.67 |       50 |... 49,151,152,154 |
|   owners.js       |      100 |      100 |      100 |      100 |                   |
|   owners_check.js |      100 |      100 |      100 |      100 |                   |
|   repo-file.js    |      100 |       50 |      100 |      100 |                29 |
|-------------------|----------|----------|----------|----------|-------------------|
```

- Test Coverage: 100% on refactored files
- Test Suites:   5 passed, 5 total
- Tests:         66 passed, 66 total
- Time:          33.583s, estimated 34s


owners/src/owner.js
```diff
@@ -16,6 +16,7 @@
 
 const path = require('path');
 const {LocalRepository} = require('./local_repo');
+const {Owner, createOwnersMap} = require('./owner');
```
rcebulko: This was silly; removed in a later commit.

owners/src/local_repo.js
```diff
@@ -0,0 +1,124 @@
+/**
+ * Copyright 2016 Google Inc.
```
danielrozenberg: All files should have "The AMP HTML Authors" instead of "Google Inc." - you don't have to replace all of those in this PR, but do it at least for any new file you're adding

owners/test/local_repo.test.js
```diff
@@ -0,0 +1,108 @@
...
+    });
+
+    it('defaults to "origin" remote', () => {
+      expect(repo.remote).toEqual('origin');
+    });
+  });
+
+  describe('checkout', () => {
+    beforeEach(() => {
+      sandbox.stub(repo, 'runCommands');
```
danielrozenberg: Whenever possible, don't stub anything in the class that you're testing. Stub the dependencies instead

owners/test/local_repo.test.js
```diff
@@ -0,0 +1,108 @@
...
+      sandbox.assert.calledWith(
+          repo.runCommands,
+          'git fetch origin my_branch',
+          'git checkout -B my_branch origin/my_branch',
+      );
+    });
+
+    it('defaults to master', async () => {
+      await repo.checkout();
+      sandbox.assert.calledWith(
```
danielrozenberg: Nit: I prefer `calledOnceWith` in cases where you don't expect multiple calls to the same stubbed method

owners/src/local_repo.js
```diff
@@ -0,0 +1,124 @@
...
+
+  /**
+   * Read the contents of a file from the checked out repo.
+   *
+   * @param {!string} relativePath file to read.
+   * @return {string} file contents.
+   */
+  async readFile(relativePath) {
+    const filePath = this.getAbsolutePath(relativePath);
+    return await fs.readFile(filePath, {encoding: 'utf8'});
```
danielrozenberg: You can use `fs.readFileSync` and remove the async keyword here (unless there's a good use-case for it to be async of course) - if you change that, change the callers to not `await` on this (`await`ing on a non-async function is harmless though, so nbd if you miss one)

owners/src/owners.js
```diff
@@ -92,11 +92,22 @@ class OwnersParser {
...
+
+  /**
+   * Parse all OWNERS rules in the repo.
+   *
+   * @return {OwnersRule[]} a list of all rules defined in the local repo.
+   */
+  async parseAllOwners() {
+    const ownersPaths = await this.localRepo.findOwnersFiles();
+    return await Promise.all(ownersPaths.map(
+        ownersPath => this.parseOwnersFile(ownersPath)));
```
danielrozenberg: Would this cause the OWNERS.yaml files to be read for every file modified in the PR?

owners/src/owners.js
```diff
@@ -33,6 +33,15 @@ class OwnersRule {
     this.owners = owners;
   }
 
+  /**
+   * Gets the full (relative) OWNERS.yaml file path.
+   *
+   * @return {string} relative file path.
+   */
+  get filePath() {
+    return path.join(`${this.dirPath}`, 'OWNERS.yaml');
```
danielrozenberg: Is the backtick necessary here?

owners/src/owners.js
```diff
@@ -18,6 +18,129 @@ const path = require('path');
...
+  /**
+   * Add a rule to the tree.
+   *
+   * Adds to the current node if paths match; otherwise, adds it to the node
+   * keyed by the next directory name.
+   *
+   * @param {!OwnersRule} rule rule to add.
+   */
+  addRule(rule) {
+    if (this.depth > 5) {
```
danielrozenberg: Why 5? What if it's deeper than 5?

owners/src/owners.js
```diff
@@ -18,6 +18,129 @@ const path = require('path');
...
+   */
+  get allRules() {
+    const parentRules = this.isRoot ? [] : this.parent.allRules;
+    return this.rules.concat(parentRules);
+  }
+
+  /**
+   * Provides a list of ownership rules applicable to a file.
+   *
+   * The most specific rules will be first, while the root-level owners rules
```
danielrozenberg: High-level Q: are top level rules always taken into account, or are there cases where the OWNERS.yaml file can reset the ownership model?

owners/src/owners.js
```diff
@@ -18,6 +18,129 @@ const path = require('path');
...
+  /**
+   * Renders the ownership tree as a string.
+   *
+   * @return {string} visual representation of the tree.
+   */
+  toString() {
+    let lines = []
+
+        const rulePrefix = '-';
+    const childPrefix = '└---';
```
danielrozenberg: Prime real estate for emoji use ;)

danielrozenberg: (but maybe in a followup PR, let's not test the GitHub API's handling of Unicode)

owners/src/local_repo.js
rcebulko: I considered that, but the other files in the app were using `require('fs').promises` so I left it. I'll come back to this and sync-ify.

owners/src/owners.js
rcebulko: This would build an owners tree with all owners files, yes. We could read fewer files, and maybe later I'll come back in to filter this down, but reading them once per PR shouldn't be too expensive. I'll leave a TODO here to restrict the searching to paths modified in the PR, and can touch that once I start working with the actual PR files.

owners/src/owners.js
rcebulko: Prime real estate for "Box-drawing character" use*
https://en.wikipedia.org/wiki/Box-drawing_character
Could have used the horizontal one instead of the dash, but I figured that would be non-obvious and harder to read/edit.

owners/src/owners.js
rcebulko: With current rulesets, no, it's a strictly subset-ownership hierarchy. If we implement exclusion or other fancy-dancy rules, an OWNERS file may parse into multiple owners rules, and the ordering here can be used to determine which rules apply.

owners/src/owners.js
rcebulko: See this is why code reviews are helpful :sweat_smile: This was in here while I was fixing an infinite-recursion bug, and I neglected to add a `TODO`

owners/src/owners.js
rcebulko: Noop.

owners/src/local_repo.js
rcebulko: Done

owners/test/local_repo.test.js
rcebulko: I agree! And at first I used `calledOnceWith`, but removed it when I remembered that's a `pytest` built-in that's not available in Jest ;)

owners/src/local_repo.js
rcebulko: Done.

owners/test/local_repo.test.js
rcebulko: Leaving a `TODO` here for now. Since the `local_repo` module uses `util.promisify(child_process.exec)`, which is assigned and keeps a reference to the original `exec` in a scope, stubbing is a bit tricky and will either be hacky or require some reworking of the `requires` in `local_repo`

rcebulko:
Update: I wrote a script to fetch the 100 most recent open PRs, look at their check-runs, and re-run the check-run creation with the refactored code. I have good news, then bad news, then good news.

Good news (Part 1): Almost all test summaries (success/failure) and output comments (reviewer suggestions) were either an identical match. The only differences were in cases of teams, which the old bot would suggest (ie. @ampproject/runtime) but were not actually implemented; now, it just suggests actual known owners.

Bad news: The scraping abruptly crashed with an error when trying to run the updated owners check on PR [#24085](https://github.com/ampproject/amphtml/pull/24085)! Oh no!

Good news (Part 2): Upon further inspection, it appears there was in fact no owners check-run ever added to that PR; in other words, it was already silently failing on this PR (and others). Conclusion: this refactor is such a no-op that not only does it preserve functionality, but it also preserves existing bugs :grinning: 

This verification of the check-runs fills the remaining confidence gap for me that this is safe to submit. Any last comments @danielrozenberg @estherkim @rsimha ?

rsimha:
Just had a chance to review this PR. Nice refactor!

PR #379, opened 2019-09-04
[owners] Add reviewer selection algorithm 

rcebulko:
Adds reviewer selection algorithm implementation and tests. Does not yet integrate it into the owners bot, and the bot is still non-blocking (see [algorithm I2I](https://github.com/ampproject/amphtml/issues/24338)). Plan is to take this and start recommending reviewer sets, but leaving the bot as non-blocking.


owners/src/reviewer_selection.js
```diff
@@ -0,0 +1,186 @@
...
+
+Repeat until all files have an OWNER in the reviewer set.
+
+*/
+
+/**
+ * Class implementing the reviewer selection algorithm.
+ */
+class ReviewerSelection {
+  /** Part 1 **/
```
danielrozenberg: I find this extremely amusing 😝

owners/src/reviewer_selection.js
```diff
@@ -0,0 +1,186 @@
...
+    while (Object.entries(fileTreeMap).length) {
+      const bestReview = this._pickBestReview(fileTreeMap);
+      if (!bestReview) {
+        // This should be impossible unless there is no top-level OWNERS file.
+        throw Error('Could not select reviewers!');
+      }
+
+      const [bestReviewer, coveredFiles] = bestReview;
+      reviews.push([bestReviewer, coveredFiles]);
+      coveredFiles.forEach(filename => delete fileTreeMap[filename]);
```
danielrozenberg: This feels… off? I dunno why, this might be perfectly normal. Just never seen this done before this way. Maybe somebody who feels much more comfortable with JS than me can chime in

owners/test/reviewer_selection.test.js
```diff
@@ -0,0 +1,253 @@
...
+
+const sinon = require('sinon');
+const {ReviewerSelection} = require('../src/reviewer_selection');
+const {OwnersRule, OwnersTree} = require('../src/owners');
+
+describe('reviewer selection', () => {
+  const sandbox = sinon.createSandbox();
+  const ownersTree = new OwnersTree();
+
+  const rootDirRule = new OwnersRule('OWNERS.yaml', ['root']);
```
danielrozenberg: Maybe give them better fake names? It's a little confusing to think of "root" as a github username... e.g., alice, bob, carol :)

danielrozenberg: Also, since these are only used to create the tree, you can inline them instead of creating a one-time-use const. This way it's easier for me to scroll up and down when reviewing to look at what I expect the results to be :)

owners/test/reviewer_selection.test.js
```diff
@@ -0,0 +1,253 @@
...
+      sandbox.assert.calledWith(
+        ReviewerSelection._reviewersForTrees,
+        sinon.match.array.contains([childDirTree, otherChildDirTree])
+      );
+      expect(reviewers).toEqual(expect.arrayContaining(['child', 'kid']));
+    });
+  });
+
+  describe('filesOwnedByReviewer', () => {
+    it('lists files for which the reviewer is an owner', () => {
```
danielrozenberg: Use an `it.each` parameterized test here - split it on the rootFiles/childFiles/kidFiles/descendantFiles part

owners/test/reviewer_selection.test.js
```diff
@@ -0,0 +1,253 @@
...
+
+  const rootDirRule = new OwnersRule('OWNERS.yaml', ['root']);
+  const childDirRule = new OwnersRule('foo/OWNERS.yaml', ['child']);
+  const otherChildDirRule = new OwnersRule('biz/OWNERS.yaml', ['child', 'kid']);
+  const thirdChildDirRule = new OwnersRule('buzz/OWNERS.yaml', ['thirdChild']);
+  const descendantDirRule = new OwnersRule('foo/bar/baz/OWNERS.yaml', [
+    'descendant',
+  ]);
+
+  const rootDirTree = ownersTree.addRule(rootDirRule);
```
danielrozenberg: Might be worth making this an object, where the keys are the directory path:
```javascript
dirTrees = {
  '/': ownersTree.addRule(...),
  '/foo': ownersTree.addRule(...),
  '/foo/bar/baz': ownersTree.addRule(...),
  ...
};
```

This way when you make comparisons below it's much more obvious without having to parse the relationships between the owner and the tree while reading the tests:
```javascript
expect(nearestTrees).toEqual([trees['/foo']])
```

owners/test/reviewer_selection.test.js
```diff
@@ -0,0 +1,253 @@
...
+      });
+    });
+  });
+
+  describe('pickReviewers', () => {
+    let fileTreeMap;
+    let bestReviewerStub;
+
+    beforeEach(() => {
+      bestReviewerStub = sandbox.stub(ReviewerSelection, '_pickBestReview');
```
danielrozenberg: Stub this separately in each test for readability reasons

owners/test/reviewer_selection.test.js
```diff
@@ -0,0 +1,253 @@
...
+
+    it('picks reviewers until all files are covered', () => {
+      bestReviewerStub.callThrough();
+      const reviews = ReviewerSelection.pickReviewers(fileTreeMap);
+      const reviewers = reviews.map(([reviewer, files]) => reviewer);
+
+      sandbox.assert.calledThrice(ReviewerSelection._pickBestReview);
+      expect(reviewers).toEqual(['child', 'thirdChild', 'root']);
+    });
+
```
danielrozenberg: Add a test that shows this doesn't necessarily add the root user even if there are multiple potential owners

owners/src/reviewer_selection.js
rcebulko: What feels off? Removing entries from a dict?

owners/src/reviewer_selection.js
rcebulko: Lol yeah probably not strictly necessary but :man_shrugging:

owners/test/reviewer_selection.test.js
rcebulko: Renamed "root" to "rootOwner", leaving the rest since it makes it easier to read what users the tests are expecting ("child" is more meaningful than "bob" (no offense Bob))

rcebulko: Whoops, nice catch. These were used at one point, but once `OwnersTree.addRule` was updated to return the subtree, it was no longer necessary. Removed.

owners/test/reviewer_selection.test.js
rcebulko: Done.

owners/test/reviewer_selection.test.js
rcebulko: Done

owners/test/reviewer_selection.test.js
rcebulko: Done.

owners/test/reviewer_selection.test.js
rcebulko: Done.

owners/src/reviewer_selection.js
rcebulko: Wrapped the lambda in braces

PR #380, opened 2019-09-04
Pull main logic into OwnersCheck class

rcebulko:
One more (much smaller) refactor PR, pulling steps from the `Owner` class and `index.js` into the `OwnersCheck` class

### DIFF
--- a/owners/index.js
+++ b/owners/index.js
@@ -55,7 +55,7 @@ module.exports = app => {
       // become just `const latestCheckRun = ownersCheck.run();`
       await ownersCheck.init();
       const fileOwners = await Owner.getOwners(ownersCheck);
-      const latestCheckRun = ownersCheck.buildCheckRun(fileOwners);
+      latestCheckRun = ownersCheck.buildCheckRun(fileOwners);
     } catch (error) {
       // If anything goes wrong, report a failing check.
       latestCheckRun = new CheckRun(
--- a/owners/index.js
+++ b/owners/index.js
@@ -16,6 +16,7 @@
 
 const sleep = require('sleep-promise');
 const {GitHub, PullRequest} = require('./src/github');
+const {LocalRepository} = require('./src/local_repo');
 const {CheckRun, OwnersCheck} = require('./src/owners_check');
 const {Owner} = require('./src/owner');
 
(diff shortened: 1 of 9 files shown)


owners/index.js
```diff
@@ -42,15 +43,19 @@ module.exports = app => {
...
 
     try {
-      const approvers = await ownersCheck.getApprovers();
+      const localRepo = new LocalRepository(process.env.GITHUB_REPO_DIR);
+      const ownersCheck = new OwnersCheck(localRepo, github, pr);
       checkRunId = await github.getCheckRunId(pr.headSha);
-      const fileOwners = await Owner.getOwners(github, pr.number);
-      latestCheckRun = ownersCheck.buildCheckRun(fileOwners, approvers);
+
+      // TODO: Once OwnersCheck swallows the remaining logic in Owner, this can
```
danielrozenberg: Always write TODO(#issuenumber) or TODO(username) if there's no issue (issue number is preferable)

owners/test/owners.test.js
```diff
@@ -128,8 +128,9 @@ describe('owners tree', () => {
 
     it('returns the nearest tree with a rule', () => {
       tree.addRule(childDirRule);
+      tree.addRule(descendantDirRule);
 
-      expect(tree.atPath('foo/bar/baz').dirPath).toEqual('foo');
+      expect(tree.atPath('foo/bar/okay.txt').dirPath).toEqual('foo');
```
danielrozenberg: Can you explain this change?

owners/src/owners_check.js
```diff
@@ -59,53 +60,100 @@ class OwnersCheck {
...
+    const passing = this._allFilesApproved(fileOwners);
+    const text = this._buildOutputText(fileOwners);
+    const summary = `The check was a ${passing ? 'success' : 'failure'}!`;
+    return new CheckRun(summary, text);
   }
 
   /**
    * Tests if all files are approved by at least one owner.
    *
+   * TODO: Replace legacy check-run code used by old Owner class.
```
danielrozenberg: TODO(...)

owners/src/owners_check.js
```diff
@@ -59,53 +60,100 @@ class OwnersCheck {
...
+
+  /**
+   * Identifies all reviewers whose latest reviews are approvals.
+   *
+   * Also includes the author, unless the author has explicitly left a blocking
+   * review.
+   *
+   * @return {string[]} list of usernames.
+   */
+  async getApprovers() {
```
danielrozenberg: Looks like this should be a private function, and external callers should get `this.approvers` instead

owners/test/owners.test.js
rcebulko: The test was meant to check that it would return the nearest tree with a rule, as opposed to returning some deep node that has no rules. But `atPath` short-circuits once the deepest owners tree is reached, so it wasn't actually testing the right bit. This makes sure that there is a subtree that exists for `foo/bar`, but it is skipped because it has no rules.

owners/index.js
rcebulko: Done

owners/src/owners_check.js
rcebulko: Done.

owners/src/owners_check.js
rcebulko: Done.

PR #383, opened 2019-09-05
Set up owners tree/rules/check to filter rules that match the file being tested

rcebulko:
Rules themselves are strictly hierarchical, and since they'll only be applied to files that are already limited by ownership subtree, re-matching the path is redundant, repetitive, and redundant. `matchesFile` will now act as a filter, such as for globs and patterns. Basic owners rules have no filters, but upcoming rules supporting filetypes and filenames and other patterns will extend the `OwnersRule` class and may override the `matchesFile` test.

### DIFF
--- a/owners/src/owners.js
+++ b/owners/src/owners.js
@@ -99,6 +99,19 @@ class OwnersTree {
     const segments = filePath.split(path.sep);
     let subtree = this;
 
+    if (!this.isRoot) {
+      const treeSegments = this.dirPath.split(path.sep);
+      while (treeSegments.length) {
+        const nextTreeDir = treeSegments.shift();
+        const nextDir = segments.shift();
+        if (nextTreeDir !== nextDir) {
+          throw new Error(
+            `Tried to find subtree at path "${filePath}" on a subtree at path "${this.dirPath}"`
+          );
+        }
+      }
+    }
+
     while (segments.length) {
       const nextDir = segments.shift();
       if (!subtree.get(nextDir)) {
--- a/owners/src/owners.js
+++ b/owners/src/owners.js
@@ -148,7 +148,9 @@ class OwnersTree {
    * @return {boolean} true of the user is an owner of the file.
    */
   fileHasOwner(filename, username) {
-    return this.atPath(filename).hasOwner(username);
+    const allRules = this.atPath(filename).allRules;
+    const fileRules = allRules.filter(rule => rule.matchesFile(filename));
+    return fileRules.some(rule => rule.owners.includes(username));
   }
 
   /**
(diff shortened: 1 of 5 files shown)


owners/src/owners.js
```diff
@@ -115,16 +128,16 @@ class OwnersTree {
   }
 
   /**
-   * Tests if a user is in the ownership path of the tree.
+   * Tests if a user is in the ownership path of a file.
    *
+   * @param {!string} filename file to test ownership for.
    * @param {!string} username user to check ownership of.
-   * @return {boolean} true of the username is in this or an ancestor's OWNERS.
+   * @return {boolean} true of the user is an owner of the file.
```
danielrozenberg: nit: "true if"

owners/src/owners.js
```diff
@@ -115,16 +128,16 @@ class OwnersTree {
...
    */
-  hasOwner(username) {
-    const allOwners = this.allRules
-      .map(rule => rule.owners)
-      .reduce((left, right) => left.concat(right), []);
-    return allOwners.includes(username);
+  fileHasOwner(filename, username) {
+    const allRules = this.atPath(filename).allRules;
+    const fileRules = allRules.filter(rule => rule.matchesFile(filename));
+    return fileRules.some(rule => rule.owners.includes(username));
```
danielrozenberg: You can merge these two lines together:
`return allRules.some(rule => rule.matchesFile(filename) && rule.owners.includes(username));`
This will short-circuit with a `true` value on the first rule that matches

owners/src/owners.js
```diff
@@ -192,31 +205,14 @@ class OwnersRule {
...
-      return false;
-    }
-
-    for (let i = 0; i < rulePathSegments.length; ++i) {
-      if (rulePathSegments[i] !== filePathSegments[i]) {
-        return false;
-      }
-    }
-
     return true;
```
danielrozenberg: Add a TODO to track this

owners/src/owners.js
rcebulko: Done.

owners/src/owners.js
rcebulko: I'm aware this could be one-lined, I find that it's more readable to break it into steps this way (and there's a very high likelihood that Closure+V8 optimize the **** out of map/filter/reduce functions anyway)

owners/src/owners.js
rcebulko: Right now there's nothing to track, any pattern matching will be as part of the resolution to one of the existing feature requests.

rcebulko: I stuck one in pointing to an arbitrary rule feature request

owners/src/owners.js
danielrozenberg: I don't think it's unreadable to have these two together - basically it says "at least one rule both matches the file and is owned by the user - it's a pretty simple rule :)

PR #385, opened 2019-09-05
Adds a health-check server with app status, owners tree, and PR tests

rcebulko:
This PR:
- Separates owners bot functionality out of `index.js`
- Adds a status check page (`/admin/status`)
- Adds a page to view the current OWNERS hierarchy (`/admin/tree`)
    - *I cached this to prevent it being spammed and causing trouble, but I'm not sure if that's a real risk or not, and it's pretty hard to avoid with the next endpoint; opinions?
- Adds a page that generates the check-run for any PR, for testing/validation purposes.

### DIFF
--- a/owners/index.js
+++ b/owners/index.js
@@ -22,11 +22,13 @@ const {OwnersParser} = require('./src/owners');
 const {OwnersCheck} = require('./src/owners_check');
 
 const GITHUB_TOKEN = process.env.GITHUB_TOKEN;
+const GITHUB_REPO_OWNER = process.env.GITHUB_REPO_OWNER || 'ampproject';
+const GITHUB_REPO_NAME = process.env.GITHUB_REPO_NAME || 'amphtml';
 
 module.exports = app => {
   const localRepo = new LocalRepository(process.env.GITHUB_REPO_DIR);
   const ownersBot = new OwnersBot(localRepo);
-  const router = app.route('/admin');
+  const adminRouter = app.route('/admin');
 
   // Probot does not stream properly to GCE logs so we need to hook into
   // bunyan explicitly and stream it to process.stdout.
@@ -39,20 +41,24 @@ module.exports = app => {
   /** Health check server endpoints **/
   // TODO(rcebulko): Implement GitHub authentication to prevent spamming any of
   // these endpoints.
-  router.get('/status', (req, res) => {
-    res.send('The OWNERS bot is live and running!');
+  adminRouter.get('/status', (req, res) => {
+    res.send([
+      'The OWNERS bot is live and running!'
+      `Project: ${process.env.GOOGLE_CLOUD_PROJECT || 'UNKNOWN'}`
+      `Version: ${process.env.GAE_VERSION || 'UNKNOWN'}`
+    ].join('<br>'));
   });
 
-  router.get('/tree', async (req, res) => {
+  adminRouter.get('/tree', async (req, res) => {
     const parser = new OwnersParser(localRepo, req.log);
     const ownersTree = await parser.parseOwnersTree();
 
     res.send(`<pre>${ownersTree.toString()}</pre>`);
   });
 
-  router.get('/check/:prNumber', async (req, res) => {
+  adminRouter.get('/check/:prNumber', async (req, res) => {
     const octokit = new Octokit({auth: `token ${GITHUB_TOKEN}`});
-    const github = new GitHub(octokit, 'ampproject', 'amphtml', app.log);
+    const github = new GitHub(octokit, GITHUB_REPO_OWNER, GITHUB_REPO_NAME, app.log);
     const pr = await github.getPullRequest(req.params.prNumber);
     const ownersCheck = new OwnersCheck(localRepo, github, pr);
     const checkRun = await ownersCheck.run();
--- a/owners/index.js
+++ b/owners/index.js
@@ -22,6 +22,7 @@ const {OwnersParser} = require('./src/owners');
 module.exports = app => {
   const localRepo = new LocalRepository(process.env.GITHUB_REPO_DIR);
   const ownersBot = new OwnersBot(localRepo);
+  const treeCache = new OneDayCache();
   const router = app.route('/admin');
 
   // Probot does not stream properly to GCE logs so we need to hook into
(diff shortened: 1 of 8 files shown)


owners/index.js
```diff
@@ -69,3 +75,45 @@ module.exports = app => {
...
   });
 };
+
+/**
+ * Simple one-day cache for endpoints that do Git commands and file reads.
+ *
+ * TODO(rcebulko): Replace with a real caching solution once the health-check
+ * server gets real functionality.
+ */
+class OneDayCache {
```
rcebulko: This made sense when I was writing it, but I'm increasingly thinking it's not necessary.

owners/index.js
```diff
@@ -14,17 +14,19 @@
...
-const GITHUB_CHECKRUN_DELAY = 2000;
+const GITHUB_TOKEN = process.env.GITHUB_TOKEN;
 
 module.exports = app => {
-  app.on(['pull_request.opened', 'pull_request.synchronize'], onPullRequest);
-  app.on('check_run.rerequested', onCheckRunRerequest);
-  app.on('pull_request_review.submitted', onPullRequestReview);
+  const localRepo = new LocalRepository(process.env.GITHUB_REPO_DIR);
+  const ownersBot = new OwnersBot(localRepo);
+  const router = app.route('/admin');
```
danielrozenberg: Calling this `router` is kinda like creating a C++ variable with `int number = 0;` - technically correct, but there's gotta be a better name 😝
Maybe `adminRouter`?
(edit: totally aware that I probably did the same thing in one of the other apps 🙃)

owners/index.js
```diff
@@ -34,85 +36,55 @@ module.exports = app => {
...
-  async function runOwnersCheck(github, pr) {
-    const localRepo = new LocalRepository(process.env.GITHUB_REPO_DIR);
-    const ownersCheck = new OwnersCheck(localRepo, github, pr);
-    let checkRunId;
-    let latestCheckRun;
+  /** Health check server endpoints **/
+  // TODO(rcebulko): Implement GitHub authentication to prevent spamming any of
+  // these endpoints.
+  router.get('/status', (req, res) => {
+    res.send('The OWNERS bot is live and running!');
```
danielrozenberg: Does AppEngine expose version numbers in process.env? If so you can conditionally add it here

owners/index.js
```diff
@@ -14,17 +14,19 @@
...
+const {OwnersCheck} = require('./src/owners_check');
 
-const GITHUB_CHECKRUN_DELAY = 2000;
+const GITHUB_TOKEN = process.env.GITHUB_TOKEN;
 
 module.exports = app => {
-  app.on(['pull_request.opened', 'pull_request.synchronize'], onPullRequest);
-  app.on('check_run.rerequested', onCheckRunRerequest);
-  app.on('pull_request_review.submitted', onPullRequestReview);
+  const localRepo = new LocalRepository(process.env.GITHUB_REPO_DIR);
```
danielrozenberg: Does this also pull the repo? If it does, that means the repo will be synced only once, on app startup... (if this doesn't do the pulling, ignore)

owners/index.js
```diff
@@ -34,85 +36,55 @@ module.exports = app => {
...
-      await github.updateCheckRun(checkRunId, latestCheckRun);
-    } else {
-      // We need to add a delay on the PR creation and check creation since
-      // GitHub might not be ready.
-      // TODO: Verify this is still needed.
-      await sleep(GITHUB_CHECKRUN_DELAY);
-      await github.createCheckRun(pr.headSha, latestCheckRun);
-    }
-  }
+    res.send(`<pre>${ownersTree.toString()}</pre>`);
```
danielrozenberg: Suuuuuuuuuuuuuuuuuuuuuuuure...
(edit: what I mean is, this works, but looks a bit hackish - no need to do this now, but whenever you feel like you have an hour to run in place, make `.hbs` templates and use those instead)

owners/index.js
```diff
@@ -34,85 +36,55 @@ module.exports = app => {
...
-   * @param {!GitHub} github GitHub API interface.
-   * @param {!number} prNumber pull request number.
-   */
-  async function runOwnersCheckOnPrNumber(github, prNumber) {
-    const pr = await github.getPullRequest(prNumber);
-    await runOwnersCheck(github, pr);
-  }
+  router.get('/check/:prNumber', async (req, res) => {
+    const octokit = new Octokit({auth: `token ${GITHUB_TOKEN}`});
+    const github = new GitHub(octokit, 'ampproject', 'amphtml', app.log);
```
danielrozenberg: Flags instead of hard coded repo slug? This way you can test it on a fork when you want to make serious changes

owners/src/owners_bot.js
```diff
@@ -0,0 +1,73 @@
...
+    const ownersCheck = new OwnersCheck(this.repo, github, pr);
+    const checkRunId = await github.getCheckRunId(pr.headSha);
+    const latestCheckRun = await ownersCheck.run();
+
+    if (checkRunId) {
+      await github.updateCheckRun(checkRunId, latestCheckRun);
+    } else {
+      // We need to add a delay on the PR creation and check creation since
+      // GitHub might not be ready.
+      // TODO: Verify this is still needed.
```
danielrozenberg: TODO(rcebulko)

owners/index.js
rcebulko: Ignoring :upside_down_face:

rcebulko: There's a separate `LocalRepository.checkout` method. I thought doing file system ops in a constructor might be a bad idea :laughing:

owners/index.js
rcebulko: LOL yep I know, it is a bit hackish but since the toString method is already there this'll be a nice way to keep a finger on the pulse of what's being parsed. Eventually it would be nice to have a more interactive view of the file tree and ownership, but as a debug tool this will do the trick for now

owners/index.js
rcebulko: Done.

rcebulko: Done.

owners/index.js
rcebulko: Done.

owners/index.js
rcebulko: Done.

owners/src/owners_bot.js
rcebulko: Done.

Issue #278, opened 2019-07-23
[owners] implement file level only owner


erwinmombay:
scoped at directory level traversing down the subtree for ownership inheritance

erwinmombay:
migrated from: https://github.com/google/github-owners-bot/issues/4

rcebulko:
## Syntax proposal
### Patterns

Support _extremely_ simple pattern-based paths
  - `*` represents any string of characters (glob, not regex)
  - by default, applies recursively (ie. `*.css` applies to all CSS files in folder and subdirectories)
  - with `./` prefix, applies only to current directory (ie. `./*.css` applies to `./main.css` but not `./foo/bar.css`)
  - if no glob pattern is present, it's an exact match (ie. `./package.json` matches only `package.json` in that directory, while `package.json` matches all `package.json` files in all subdirectories as well)
  - multiple comma-separated patterns may be specified (ex. `*.js, *.css`)

Pattern ownership is additive
  - if `user1` is listed in the OWNERS file as owning the directory, then `user1` also owns any pattern-based rules

For the purposes of the reviewer selection algorithm, pattern-based rules have the same depth as the directory ownership rules.

### Rules

Pattern-based rules will be defined as a dict in the OWNERS.yaml file, adjacent in the list to the file owners. They key will be the pattern, while the value will be the list of owners of files affected by the pattern. Example:

```yaml
- alice
- bob
- ./package.json:
  - charlie
```

 In the case of an OWNERS file with pattern rules but no directory owners, the dict should still be within a list. Example:

```yaml
- *.css:
  - alice
  - bob
```

rcebulko:
Note: Pattern-based rules are **_NOT_** planned to support subdirectories/full paths at this time. Unless a need arises, the appropriate subdirectory can have its own OWNERS file.

rcebulko:
@rsimha @erwinmombay Any opinions?

erwinmombay:
could we follow glob matching rules were **/* is needed for recursive matching? just to align with glob rules properly

rcebulko:
@erwinmombay So instead of `*.js` to match recursively and `./*.js` to match in same-directory only, use `*.js` for same-directory and `**/*.js` for recursive?

rcebulko:
[...]

PR #386, opened 2019-09-06
Add file matching for pattern-based rules

rcebulko:
First step towards #278 

This PR adds the rule class that handles matching files based on a glob pattern. It also splits the definition of rule classes into a separate `rules` module.

Note that the parser does not yet create these pattern-based rules.

@danielrozenberg Again, this looks like a lot if you look at the files alone, but 90% of this PR is tests and moving code out into a new module.

### DIFF
--- a/owners/src/owners.js
+++ b/owners/src/owners.js
@@ -226,6 +226,55 @@ class OwnersRule {
 }
 
 /**
+ * A pattern-based ownership rule applying to all matching files in the
+ * directory and all subdirectories.
+ *
+ * Treats `*` as a wildcard glob pattern; all other characters are treated as
+ * literals.
+ */
+class PatternOwnersRule extends OwnersRule {
+  /**
+   * Constructor.
+   *
+   * @param {!string} ownersPath path to OWNERS file.
+   * @param {string[]} owners list of GitHub usernames of owners.
+   * @param {!string} pattern filename/type pattern.
+   */
+  constructor(ownersPath, owners, pattern) {
+    super(ownersPath, owners);
+    this.pattern = pattern;
+    this.regex = new RegExp(
+      pattern
+        .split('*')
+        .map(PatternOwnersRule.escapeRegexChars)
+        .join('.*?')
+        .split(/\s*,\s*/)
+        .join('|')
+    );
+  }
+
+  /**
+   * Escapes all regex characters in a string.
+   *
+   * @param {!string} str string to escape.
+   * @return {string} copy of the string that can be used in a regex.
+   */
+  static escapeRegexChars(str) {
+    return str.replace(/[[\]{}()*+?.\\^$|]/g, '\\$&');
+  }
+
+  /**
+   * Test if a file is matched by the pattern rule.
+   *
+   * @param {!string} filePath relative path in repo to the file being checked.
+   * @return {boolean} true if the rule applies to the file.
+   */
+  matchesFile(filePath) {
+    return this.regex.test(path.basename(filePath));
+  }
+}
+
+/**
  * Parser for OWNERS.yaml files.
  */
 class OwnersParser {
@@ -291,4 +340,4 @@ class OwnersParser {
   }
 }
 
-module.exports = {OwnersParser, OwnersRule, OwnersTree};
+module.exports = {OwnersParser, OwnersRule, PatternOwnersRule, OwnersTree};
--- a/owners/src/owners.js
+++ b/owners/src/owners.js
@@ -16,6 +16,7 @@
 
 const path = require('path');
 const yaml = require('yamljs');
+const {OwnersRule} = require('./rules.js');
 
 /**
  * A tree of ownership keyed by directory.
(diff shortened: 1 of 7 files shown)


owners/test/index.test.js
```diff
@@ -526,55 +526,4 @@ describe('owners bot', () => {
       await probot.receive({name: 'pull_request_review', payload: review35});
     });
   });
-
-  describe('when there is a failure to run the check', () => {
```
rcebulko: This was a relic that should have been removed in the PR that moved error checking into OwnersCheck.

owners/src/rules.js
```diff
@@ -84,8 +84,6 @@ class PatternOwnersRule extends OwnersRule {
         .split('*')
         .map(PatternOwnersRule.escapeRegexChars)
         .join('.*?')
-        .split(/\s*,\s*/)
```
rcebulko: Realized this should be a part of the parser, not a part of the rule itself.

owners/src/owners.js
```diff
@@ -219,6 +219,55 @@ class OwnersRule {
...
+  }
+
+  /**
+   * Escapes all regex characters in a string.
+   *
+   * @param {!string} str string to escape.
+   * @return {string} copy of the string that can be used in a regex.
+   */
+  static escapeRegexChars(str) {
+    return str.replace(/[[\]{}()*+?.\\^$|]/g, '\\$&');
```
danielrozenberg: YO DAWG I HEARD YOU LIKE REGEX

owners/src/owners.js
```diff
@@ -219,6 +219,55 @@ class OwnersRule {
...
+  constructor(ownersPath, owners, pattern) {
+    super(ownersPath, owners);
+    this.pattern = pattern;
+    this.regex = new RegExp(
+      pattern
+        .split('*')
+        .map(PatternOwnersRule.escapeRegexChars)
+        .join('.*?')
+        .split(/\s*,\s*/)
+        .join('|')
```
danielrozenberg: This is kind of cryptic... I wonder if using something like [minimatch](https://www.npmjs.com/package/minimatch) or [glob](https://www.npmjs.com/package/glob) instead of this whole shebang would make sense

owners/test/owners.test.js
```diff
@@ -254,9 +258,77 @@ describe('owners rules', () => {
...
+
+    describe('regexEscape', () => {
+      it.each([
+        ['file.txt', 'file\\.txt'],
+        ['this+that.txt', 'this\\+that\\.txt'],
+        ['*.css, *.js', '\\*\\.css, \\*\\.js'],
+        ['with space.txt', 'with space\\.txt'],
+        ['with-hyphen.txt', 'with-hyphen\\.txt'],
+      ])('escapes "%p" as "%p"', (text, expected) => {
+        expect(PatternOwnersRule.escapeRegexChars(text)).toEqual(expected);
```
danielrozenberg: This seems like it should only be an internal method. Can you test the higher-level function that calls this instead, like you do above?

owners/test/owners.test.js
```diff
@@ -254,9 +258,77 @@ describe('owners rules', () => {
...
+      });
+
+      it('matches comma-separated patterns', () => {
+        const rule = new PatternOwnersRule('OWNERS.yaml', [], '*.js, *.css');
+
+        expect(rule).toMatchFile('main.js');
+        expect(rule).toMatchFile('style.css');
+      });
+
+      it('matches files in subdirectories', () => {
```
danielrozenberg: it.each

owners/src/owners.js
```diff
@@ -217,6 +218,16 @@ class OwnersRule {
...
     return true;
   }
+
+  /**
+   * Describe the rule.
+   *
+   * @return {string} description of the rule.
+   */
+  toString() {
+    const ownersList = this.owners.length ? this.owners.join(', ') : '-';
```
danielrozenberg: In what cases would this be empty? Only when the OWNERS.yaml file is empty? That's an error, right?

owners/test/owners.test.js
```diff
@@ -226,14 +226,14 @@ describe('owners tree', () => {
...
           '└───foo',
-          ' * child',
+          ' * All: child',
           '    └───bar',
           '        └───baz',
-          '         * descendant',
+          '         * All: descendant',
           '└───biz',
-          ' * child',
+          ' * All: child',
```
danielrozenberg: Add file specific rules to this test

owners/src/owners.js
rcebulko: lawlz

owners/src/owners.js
rcebulko: The bottom two lines here disappear in a later, the comma bit belongs in the parser, which just leaves `*` as the only thing being replaced.

owners/test/owners.test.js
rcebulko: Sure, I'll move some of these tests to `matchesFile`

owners/src/owners.js
rcebulko: You're right, I was thinking in cases where there are pattern-based rules but no regular rules, but in that case the rule would never be created without owners. Removing this logic.

owners/src/owners.js
rcebulko: Raised discussion in #278

owners/test/owners.test.js
rcebulko: Done.

owners/src/owners.js
rcebulko: Done.

owners/test/owners.test.js
rcebulko: Done.

PR #394, opened 2019-09-09
Improve status check and check-run output

rcebulko:


owners/src/owners_check.js
```diff
@@ -203,14 +203,14 @@ class OwnersCheck {
         );
 
         if (fileApprovers.length) {
-          return `- ${filename} (${fileApprovers.join(', ')})`;
+          return `- ${filename} _(${fileApprovers.join(', ')})_`;
```
danielrozenberg: So.... this _could_ break if two one have a username that starts with a `_`. This is such an edge case that I don't know if you care about fixing it, but you could map the fileApprovers to prefix with a slash if the first character of the username is `_` :)

rcebulko: That's a great catch! I didn't even think about Markdown escaping to be honest. I'll do it for all usernames and files as well, since the filenames could easily have underscores that start adding and removing italics all over the place.

rcebulko: Update: I just tried to create such a username, and got the following error from GitHub: **Username may only contain alphanumeric characters or single hyphens, and cannot begin or end with a hyphen.** So it looks like underscores aren't even allowed, and definitely not on either end, probably for this exact reason. Regardless, it seems like an unlikely or impossible edge case that's not worth fixing :)

danielrozenberg: Ha, fair. I thought I've seen usernames with underscore but it's probably either my imagination or legacy usernames :P

PR #398, opened 2019-09-09
Collect and display OWNERS parser errors

rcebulko:


owners/src/types.js
```diff
@@ -42,4 +42,14 @@ let FileTreeMap;
...
 
+/**
+ * The result of parsing OWNERS rules, along with any errors encountered.
+ *
+ * @typedef {{
+ *   rules: null|OwnersRule[],
+ *   errors: OwnersParserError[],
+ * }}
+ */
+let OwnersParserResult;
```
rsimha: This line might need a trailing ` // eslint-disable-line no-unused-vars` if the linter complains.

owners/src/parser.js
```diff
@@ -37,23 +46,29 @@ class OwnersParser {
    * Parse an OWNERS.yaml file.
    *
    * @param {!string} ownersPath OWNERS.yaml file path.
-   * @return {OwnersRule} parsed OWNERS file rule.
+   * @return {OwnersParserResult} parsed OWNERS file rule.
```
erwinmombay: add `!`

owners/src/parser.js
```diff
@@ -66,9 +81,19 @@ class OwnersParser {
...
+    const allErrors = [];
+
+    ownersPaths.forEach(ownersPath => {
+      const {rules, errors} = this.parseOwnersFile(ownersPath);
+      allRules.push(...rules);
+      allErrors.push(...errors);
+    });
+
+    return {
+      rules: allRules,
```
erwinmombay: this makes it not nullable then right? in the definition it states it can be nullable

owners/src/parser.js
rcebulko: `!` is not used in the `@return` tag, only in `@param` tags. `@return` requires an explict `null|Type` if the return value is optional/nullable.

owners/src/types.js
rcebulko: It was missing from the `module.exports`; fixed.

owners/src/parser.js
```diff
@@ -37,23 +46,29 @@ class OwnersParser {
...
     const lines = yaml.parse(contents);
+    const errors = [];
+    const rules = [];
 
-    if (!(lines instanceof Array)) {
-      this.logger.warn(
-        `Failed to parse file '${ownersPath}'; must be a YAML list`
+    if (lines instanceof Array) {
+      const stringLines = lines.filter(line => typeof line === 'string');
+      const ownersList = stringLines.filter(line => line.indexOf('/') === -1);
```
erwinmombay: couldn't you do this in one filter? doesn't look like you need the intermediate stringLines var

owners/src/parser.js
rcebulko: Removed `null|` from typedef; resolving.

owners/src/parser.js
rcebulko: This sets up for later PRs which will be splitting rules into dicts and strings, since dicts will be used for pattern-based rules.

owners/src/parser.js
rcebulko: Update: Not sure where I read this, but I distinctly recall using `!` until I was corrected/got some lint error/typescript warning to remove them. Now I don't know what to believe. At some point I'll be updating eslint's config around JSDoc (specifically `Array<T>` vs `T[]`; I'll look into this then.

PR #403, opened 2019-09-10
Add support for glob and pattern parsing + improve testing and error handling

### LINKED ISSUE #278: [owners] implement file level only owner

### LINKED ISSUE #282: [owners] Provide a comment with instructions at the top of every `OWNERS.yaml` file
When someone new to the codebase wants to edit an OWNERS.yaml file, it's hard to find the format for declaring owners for individual files / directories, and for declaring a working group as an owner.

A single line comment that points to the OWNERS.yaml spec would be useful.

I'm happy to write up a PR if a link exists.

rcebulko:
See the added `OWNERS.example.yaml` for supported syntax.
Closes #278 
Closes #282 

As a summary/guide to understanding and reviewing the parser, here is the Jest test summary for the module:

```
owners parser error
  toString
    ✓ displays the file containing the error
    ✓ displays the error message

owners parser
  parseOwnersFile
    ✓ reads the file from the local repository
    ✓ assigns the OWNERS directory path
    ✓ parses a YAML list
    ✓ parses a YAML list with blank lines and comments
    team rule declarations
      ✓ returns no rule
      ✓ records an error
    owner with a leading @
      ✓ parses ignoring the @ sign
      ✓ records an error
    rule dictionary
      ✓ parses a single owner into a pattern rule
      ✓ parses a list of owners into a pattern rule
      ✓ reports errors for non-string owners
      ✓ starting with **/ parses into a recursive pattern rule
      ✓ parses no rule if no valid owners are listed
      ✓ reports an error for patterns containing illegal '/'
      ✓ parses comma-separate patterns as separate rules
    files containing top-level dictionaries
      ✓ returns no rules
      ✓ returns a parser error
  parseAllOwnersRules
    ✓ reads all owners files in the repo
    ✓ does not include invalid rules
    ✓ collects errors from all parsed files
  parseOwnersTree
    ✓ adds each rule to the tree
    ✓ returns parser errors
```

### DIFF
--- a/owners/src/parser.js
+++ b/owners/src/parser.js
@@ -23,12 +23,23 @@ const {OwnersTree} = require('./owners_tree');
  */
 class OwnersParserError extends Error {
   /**
+   * Constructor
+   *
+   * @param {!string} ownersPath OWNERS.yaml file path (for error reporting).
+   * @param {!string} message error message;
+   */
+  constructor(ownersPath, message) {
+    super(message);
+    this.ownersPath = ownersPath;
+  }
+
+  /**
    * Displays the error message.
    *
    * @return {string} error message.
    */
   toString() {
-    return `OwnersParserError: ${this.message}`;
+    return `OwnersParserError [${this.ownersPath}]: ${this.message}`;
   }
 }
 
@@ -48,9 +59,48 @@ class OwnersParser {
   }
 
   /**
+   * Parse a single owner declaration.
+   *
+   * TODO(rcebulko): Add support for teams.
+   *
+   * @private
+   * @param {!string} ownersPath OWNERS.yaml file path (for error reporting).
+   * @param {!string} owner owner username.
+   * @return {OwnersParserResult<string[]>} list of owners' usernames.
+   */
+  _parseOwnersLine(ownersPath, owner) {
+    const owners = [];
+    const errors = [];
+
+    if (owner[0] === '@') {
+      const lineResult = this._parseOwnersLine(ownersPath, owner.slice(1));
+
+      owners.push(...lineResult.result);
+      errors.push(
+        new OwnersParserError(
+          ownersPath,
+          `Ignoring unnecessary '@' in '${owner}'`
+        ),
+        ...lineResult.errors
+      );
+    } else if (owner.indexOf('/') !== -1) {
+      errors.push(
+        new OwnersParserError(
+          ownersPath,
+          `Failed to parse owner '${owner}'; team ownership not yet supported`
+        )
+      );
+    } else {
+      owners.push(owner);
+    }
+
+    return {result: owners, errors};
+  }
+
+  /**
    * Parse an OWNERS.yaml file.
    *
-   * @param {!string} ownersPath OWNERS.yaml file path.
+   * @param {!string} ownersPath OWNERS.yaml file path (for error reporting).
    * @return {OwnersParserResult<OwnersRule[]>} parsed OWNERS file rule.
    */
   parseOwnersFile(ownersPath) {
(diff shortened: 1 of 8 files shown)


owners/src/parser.js
```diff
@@ -47,10 +62,140 @@ class OwnersParser {
...
+   * @private
+   * @param {!string} ownersPath OWNERS.yaml file path (for error reporting).
+   * @param {!string} owner owner username.
+   * @return {OwnersParserResult<string[]>} list of owners' usernames.
+   */
+  _parseOwnersLine(ownersPath, owner) {
+    const owners = [];
+    const errors = [];
+
+    if (owner[0] === '@') {
```
danielrozenberg: Is `owner` guaranteed to not be empty here?

owners/src/parser.js
```diff
@@ -47,10 +62,140 @@ class OwnersParser {
...
+   */
+  _parseOwnersDict(ownersPath, ownersDict) {
+    const [[pattern, ownersList]] = Object.entries(ownersDict);
+    const rules = [];
+    const errors = [];
+
+    // Treat each comma-separated subpattern as its own rule definition.
+    pattern.split(/\s*,\s*/).forEach(subpattern => {
+      const owners = [];
+      const isRecursive = subpattern.indexOf('**/') === 0;
```
danielrozenberg: `subpattern.startsWith('**/')`

owners/src/parser.js
```diff
@@ -47,10 +62,140 @@ class OwnersParser {
...
+      if (isRecursive) {
+        subpattern = subpattern.slice(3);
```
danielrozenberg: Supernit: you can also do `subpattern.slice('**/'.length)` to make this more clear, or even replace `'**/'` with a top-level const for that purpose

owners/src/parser.js
```diff
@@ -47,10 +62,140 @@ class OwnersParser {
...
+      }
+
+      if (subpattern.indexOf('/') !== -1) {
+        errors.push(
+          new OwnersParserError(
+            ownersPath,
+            `Failed to parse rule for pattern '${subpattern}'; ` +
+              "directory patterns other than '**/' not supported"
```
danielrozenberg: I'm surprised the different quote styles don't break Lint

owners/src/parser.js
```diff
@@ -61,19 +206,29 @@ class OwnersParser {
 
     if (lines instanceof Array) {
       const stringLines = lines.filter(line => typeof line === 'string');
```
danielrozenberg: Did you consider doing the iterations here in the higher level function and using a switch decide which parser method to call? If you do something like that you can remove the `.forEach`'s in the sub-methods and make them only parse a single "line" at a time

```javascript
for (const line of lines) {
  switch (typeof line) {
    case 'string':
      …
      break;

    case 'object':
      …
      break;
    }
  }
}
```

owners/src/parser.js
rcebulko: It's guaranteed by the method signature to be a string. It can (by signature) be an empty string (though it never will be because of how it's called), but that won't break this (`''[0] === undefined`)

owners/src/parser.js
rcebulko: I forgot that was in JS! Nice!

owners/src/parser.js
rcebulko: Lint explicitly allows different quotes when it allows you to avoid escaping quotes inside the string.

owners/src/parser.js
rcebulko: I like this idea, done. Also ended up removing the need for the double-quotes you commented on above, since it became a format string

owners/src/parser.js
rcebulko: That doesn't really work out the way I'd like, since all string lines should be in the owners list for a single rule (not N separate single-owner rules). Meanwhile, each dict line needs to be parsed separately. Added to that, this approach lets me re-use `parseOwnersList` between top-level and dict-level owners lists. I'm not sure what `forEach`'s you're talking about removing; but I did try a handful of ways of shifting things around, and this came out to be among the most sane

PR #408, opened 2019-09-11
Pull initialization from owners check into owners bot

rcebulko:
YARPR: Yet Another Refactor Pull Request

`OwnersCheck` has an `init` method which does asynchronous tasks like checking out the repo, reading/parsing from the filesystem, and fetching PR info from GitHub. The rest of the check runs synchronously. Since next steps include adding one or two more checks which will make use of some of the same things (files changed, ownership tree, reviewers/approvers), it's worth pulling this initialization into the `OwnersBot`, which will then pass it down to the `OwnersCheck` and eventually also the `DependencyOwnersCheck` and `OwnersSyntaxCheck`. This PR moves that initialization up, removes any `await`'s before `ownersCheck.run` (since it's no long async), and shifts all those initialization tests to the proper test file.

All files still have 100% test coverage.

### DIFF
--- a/owners/test/owners_bot.test.js
+++ b/owners/test/owners_bot.test.js
@@ -52,12 +52,12 @@ describe('owners bot', () => {
 
   describe('initPr', () => {
     beforeEach(() => {
-      sandbox.stub(GitHub.prototype, 'getReviews').returns(
-        [approval, otherApproval]
-      );
-      sandbox.stub(GitHub.prototype, 'listFiles').returns(
-        ['changed_file1.js', 'foo/changed_file2.js']
-      );
+      sandbox
+        .stub(GitHub.prototype, 'getReviews')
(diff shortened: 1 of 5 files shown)


owners/test/owners_bot.test.js
```diff
@@ -26,20 +28,71 @@ const {
 
 describe('owners bot', () => {
   let sandbox;
-  const fakeGithub = new GitHub({}, 'ampproject', 'amphtml', console);
-  const pr = new PullRequest(1337, 'author_user', 'test_sha');
+  const github = new GitHub({}, 'ampproject', 'amphtml', console);
```
rcebulko: The handful of changes in these test files are here because tests from two different files used similar constants (like the PR), so now that tests share one instance they are getting updated to reflect that. No functional change.

PR #418, opened 2019-09-13
[owners] Add support for parsing teams

rcebulko:


owners/src/owners_bot.js
```diff
@@ -32,8 +33,28 @@ class OwnersBot {
...
+  /**
+   * Initialize the bot's list of teams.
+   *
+   * Also initializes each team's member list, spaced out to avoid hitting rate
+   * limits. This is so that the member lists of many teams do not need to be
+   * requested all at once when parsing the owners tree.
+   *
+   * @param {!GitHub} github GitHub API interface.
+   */
+  async initTeams(github) {
```
danielrozenberg: This should be cached and invalidated every X minutes (say, 10? 60?) because at some point we won't be reinitializing the owners bot often, but the teams could change...

rcebulko: There's already a `TODO(rcebulko)` for this in `index.js` where this method is called. The whole bot relies on a shared local repository that is currently being checked out for every PR. Once these last few core features are done, I'm going to be going in and caching/isolating a lot of stuff like this.

rcebulko: There's already a `TODO(rcebulko)` for this in `index.js` where this method is called. The whole bot relies on a shared local repository that is currently being checked out for every PR. Once these last few core features are done, I'm going to be going in and caching/isolating a lot of stuff like this.

rcebulko: There's already a `TODO(rcebulko)` for this in `index.js` where this method is called. The whole bot relies on a shared local repository that is currently being checked out for every PR. Once these last few core features are done, I'm going to be going in and caching/isolating a lot of stuff like this.

Issue #284, opened 2019-07-23
[owners] Automatically assign incoming renovate PRs to the current build cop


erwinmombay:
/per @rsimha 
Incoming [renovate PRs](https://github.com/ampproject/amphtml/pulls/renovate-bot) tend to get ignored by most build cops. The owners bot is in a good position to assign each incoming PR to the current build cop for prompt attention.

We can do two things when a renovate PR is initially created, or a new commit is pushed:

- Assign (or reassign) the PR to the current build cop
- Add the current build cop as reviewer

/cc @danielrozenberg who is working on an API to return the github ID of the current build cop
/cc @mrjoro

danielrozenberg:
I think having a self-contained app that only handles this is a better choice. There's no reason to tack on a feature into the owners bot, and with Probot and the lessons learned from previous apps that we wrote it should be super simple to create this app

rcebulko:
Agreed with @danielrozenberg , see https://github.com/ampproject/amp-github-apps/issues/283#issuecomment-521016027

rsimha:
@rcebulko @danielrozenberg I suggested handling this via the owners bot for a couple reasons:

- To avoid the risk of conflicts / notification spam from multiple apps that assign / re-assign reviewers for PRs
- There is a legitimate ownership aspect to renovate PRs, in that they shouldn't be merged until they are reviewed by whoever owns the dependency being upgraded

What I had in mind was for the owners bot to special-case renovate PRs and examine the dependency being upgraded. After one or more owners are identified, the bot can go on to use the same logic as it would for any other PR, and watch for review events to determine if the owners check is satisfied.

/cc @erwinmombay who had input on this

rcebulko:
I recently found the [depcheck](https://www.npmjs.com/package/depcheck) package, which could be extremely useful here. It identifies which files depend on each package listed as a dependence or devDependency is package.json (including transitive dependencies). I've included the output of running it on the `ampproject/amphtml` root directory below. I'd propose that the renovate-bot assignment logic just identify the dependency being upgraded and assign reviewers using the same algorithm, as if the PR were modifying all of those files. That would keep it simple, consistent, and easy to debug the owners set (as opposed to looking at line numbers, commit histories, etc.).

```json
{
  "dependencies":[
    "moment",
    "preact",
    "preact-compat",
    "prop-types",
    "web-activities"
  ],
  "devDependencies":[
    "@babel/plugin-proposal-class-properties",
    "@babel/preset-env",
[...]

Issue #281, opened 2019-07-23
[owners] print out warning in checks tab for empty OWNERS.yaml


erwinmombay:
when we find an owners file that is empty or invalid formatting, print a warning on the "checks" tab for it to be fixed.

rcebulko:
We could probably make this a blocking check rather than just a warning; I don't see much reason we'd want to allow a broken/empty OWNERS file to be merged

rcebulko:
As I've been working with the OWNERS files and parsing, there's a lot of cases where incorrect YAML could totally break parsing without being obvious. I think either A) a test suite should be added to explicitly test that all OWNERS files can be parsed without error, and block PRs which try to submit breaking changes to OWNERS files or B) a standalone check (probably also run by the OWNERS bot) checks any modified OWNERS files and blocks the PR if it introduces breaking changes.

I think B is probably the best option, especially since it could live in the existing owners bot app and just add a separate check-run to PRs altering OWNERS files. This would become more valuable as we expand the supported syntax of OWNERS files to include wildcards, teams, file-level rules, etc. If we do it this way, once we fix any existing issues in OWNERS files, we wouldn't need the warnings as we would have a guarantee that all OWNERS files comply with the spec, and the owners bot itself wouldn't have to handle as many edge/error cases.

rsimha:
Agree that a check of this sort is best run by the owners bot, eventually in blocking mode. +1 to B in the comment above.

danielrozenberg:
I'm of the other opinion - I think that we should add a blocking check on the `amphtml` repository's Travis runs that verifies nobody breaks the ownership model by verifying that all `OWNERS.yaml` files are correct. Just like we don't allow merging .js or .json files!

That way the owners bot is guaranteed* to work with correct data. Trying to recover what are the correct rules from a set of incorrect files can open us up to troubles

*to the extent that anything in software is guaranteed

estherkim:
I'm with @danielrozenberg that the syntax check should be a Travis check.

I understand that having the same parser in 2 repos is unwieldy, but we already are dealing with 2 repos with having the `OWNERS.yaml` files living in `amphtml`. It makes sense that `amphtml` checks the syntax of its own files, and we have `determineBuildTargets()` that can cleanly detect an `OWNERS` build target.

A problem with using the bot to check syntax, is that someone could go around the owners bot and add themselves as an owner. The owners bot should only read from master, not PR changes. Using Travis as a gate check prevents this.

[...]

PR #425, opened 2019-09-16
Update getCheckRunId to support multiple different checks

rcebulko:
Since the OwnersBot will soon also run other checks (notably an owners-syntax-check and eventually a dependency-owners-check), this refactors the `getCheckRunId` method to return multiple check-run IDs instead of just the owners check-run ID.

### DIFF
--- a/owners/src/github.js
+++ b/owners/src/github.js
@@ -21,7 +21,7 @@ class PullRequest {
   /**
    * Constructor.
    *
-   * @param {!number} number pull request number.
+   * @param {number} number pull request number.
    * @param {!string} author username of the pull request author.
    * @param {!string} headSha SHA hash of the PR's HEAD commit.
    */
@@ -67,7 +67,7 @@ class Team {
   /**
    * Constructor.
    *
-   * @param {!number} id team ID.
+   * @param {number} id team ID.
    * @param {!string} org GitHub organization the team belongs to.
    * @param {!string} slug team name slug.
    */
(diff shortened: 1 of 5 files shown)


rcebulko:
@estherkim The owners **bot** is planned to eventually own 3 checks:
- the _owners check_ will verify that the PR has approval of at least one owner for each changed file
- the _owners syntax check_ will only pop up on PRs modifying OWNERS files, and will prevent submitting changes to `OWNERS.yaml` that contain invalid syntax (see #281 )
- the _dependency owners check_ will be  a special-case run for PRs (especially from renovate-bot) which modify `package.json`, and will assign reviewers based on the files that depend on the modified package (see #284 )

When a PR comes it, it will always trigger the _owners check_; it will run the _owners syntax check_ iff any `OWNERS.yaml` file is included; and it will run the dependency owners check iff `package.json` is included

estherkim:
Thanks for clarifying! I think _owners check_ and _dependency owners check_ should be merged into 1 check. 

To any contributor, having more than 1 owners check or even different owners checks on their PRs can be confusing.

owners/src/github.js
```diff
@@ -273,11 +271,11 @@ class GitHub {
    * will return `null`.
    *
    * @param {!string} sha SHA hash for head commit to lookup check-runs on.
-   * @return {number|null} check-run ID if one exists, otherwise null.
+   * @return {Object<!string, !number>} map from check names to check-run IDs.
```
danielrozenberg: I believe that `!number` is not a thing

rcebulko: Indeed it is not! Removed.

Issue #395, opened 2019-09-09
[owners] Support no-notify and always-notify owners


rcebulko:
Some contributors, such as top-level owners, may not want to be automatically notified or added as a reviewer to PRs. In other cases, some contributors may have horizontal ownership over a wide range of directories, such as over *.protoascii files. For these cases, we should support special syntax. Suggestion:

- `?someuser`: The user `someuser` is an owner, but should not be auto-notified or added as a reviewer by the bot; they can be suggested by the bot, and added manually when the PR is ready for final approval
- `!someuser`: The user `someuser` is an owner, and even when their review is not required, they should always be notified/tagged/added as a reviewer so they can monitor changes
- `!!someuser`: The user `someuser` is a "super-owner" of the directory/file-pattern, and changes to the covered files may not be submitted without their approval

/cc @rsimha

danielrozenberg:
I like the `?` syntax - this was a problem we had with the bundle-size bot where we wanted some people to be able to review PRs, but not have them auto-added as reviewers.

The `!` syntax is also useful, but I'm not sure whether the `!!` syntax is a little over-engineery. I can't imagine a reasonable scenario where a `!` user is notified, says nothing, and then regrets it after the PR has been merged by other OWNERS

rcebulko:
@danielrozenberg Suppose `languageperson` should have veto power over any `*.protoascii` file changes, and I'm an owner of `extensions/amp-foo`. I could create `extensions/amp-foo/stuff.protoascii`, get a review from anyone, and submit (since I own it, I can approve it). In that case, it could be submitted before `languageperson` ever gets the chance to see it or approve/reject.

I agree it's a corner-case, and probably worth holding off on implementing `!!` until we see a need arise.

honeybadgerdontcare:
For `*.protoascii` files I believe `!someuser` would be sufficient. Being aware of an incoming change to validation rules is helpful but we don't always need to be gatekeepers of it as some changes are fairly straight-forward.

rcebulko:
@honeybadgerdontcare I'm a little unclear now; your comment on [this PR](https://github.com/ampproject/amphtml/pull/24518) suggests you *would not* need a "require-approval" gatekeeping syntax; whereas here you suggest you'd look forward to/need the ability to require approval. Would you mind clarifying how important each is to you/your subteam?

honeybadgerdontcare:
Sorry for the confusion.

[...]

PR #430, opened 2019-09-17
Adds always- and never-notify modifiers to Owner class

rcebulko:
Also includes methods in the Owners Bot to use these modifiers to determine the set of reviewers to request and owners to notify/tag in the PR

Part of #395

### DIFF
--- a/owners/src/types.js
+++ b/owners/src/types.js
@@ -68,7 +68,7 @@ let FileRef;
  *
  * @typedef {{
  *   checkRun: CheckRun,
- *   reviewers: ?string[],
+ *   reviewers: string[],
  * }}
  */
 let OwnersCheckResult;
--- a/owners/src/types.js
+++ b/owners/src/types.js
@@ -68,7 +68,7 @@ let FileRef;
  *
  * @typedef {{
  *   checkRun: CheckRun,
- *   reviewers: ?ReviewerFiles[],
+ *   reviewers: ?string[],
  * }}
  */
 let OwnersCheckResult;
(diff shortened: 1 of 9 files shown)


owners/src/types.js
```diff
@@ -63,10 +63,21 @@ let OwnersParserResult;
  */
 let FileRef;
 
+/**
+ * The result of an owners check.
+ *
+ * @typedef {{
+ *   checkRun: CheckRun,
+ *   reviewers: ?string[],
```
danielrozenberg: Why did you choose to have this optional, as opposed to always returning a list, even when it's empty (=> less null checks)

owners/test/owner.test.js
```diff
@@ -79,14 +111,32 @@ describe('owner teams', () => {
...
   describe('toString', () => {
     it("returns the team members' usernames as a comma-separated list", () => {
-      expect(owner.toString()).toEqual('auser, anothermember');
+      expect(owner.toString()).toEqual('[auser, anothermember]');
+    });
+
+    it('includes any modifiers', () => {
+      const owner = new TeamOwner(myTeam, OWNER_MODIFIER.NOTIFY);
+      expect(owner.toString()).toContain(' (always notify)');
+      expect(owner.toString().endsWith(' (always notify)')).toBe(true);
```
danielrozenberg: You can replace these two with
```javascript
expect(owner.toString()).toMatch(/ \(always notify\)$/);
```

owners/src/types.js
rcebulko: Sure, done.

owners/test/owner.test.js
rcebulko: Whoops, didn't mean to have both in here (I think it was the result of a merge/rebase/stash or something). Regardless, I'll swap out for this matcher. Done.

PR #443, opened 2019-09-18
[owners] Create a comment tagging owners with the always-notify modifier

rcebulko:
Owners specified as always-notify, ie. `!rcebulko`, should be tagged on a PR even if they are not a reviewer. This PR uses existing API interfaces to verify that such a comment does not yet exist, determine which users/teams should be notified for which files, and creates a PR comment tagging those users/teams.

### DIFF
--- a/owners/src/owners_bot.js
+++ b/owners/src/owners_bot.js
@@ -157,15 +157,16 @@ class OwnersBot {
 
     const notifies = this._getNotifies(fileTreeMap);
 
-    const tag = name => `@${name}`;
-    const header = 'Some owners are notified of changes to files in this PR:';
     const fileNotifyComments = Object.entries(notifies).map(
-      ([filename, names]) => `- ${filename}: ${names.map(tag).join(', ')}`
+      ([name, filenames]) => {
+        const header = `Hey @${name}, these files were changed:`;
+        const files = filenames.map(filename => `- ${filename}`);
+        return [header, ...files].join('\n');
+      }
     );
 
     if (fileNotifyComments.length) {
-      const comment = [header, ...fileNotifyComments].join('\n');
-      await github.createBotComment(prNumber, comment);
+      await github.createBotComment(prNumber, fileNotifyComments.join('\n\n'));
     }
   }
 
@@ -223,17 +224,20 @@ class OwnersBot {
    * Determine the set of owners to notify/tag for each file.
    *
    * @param {!FileTreeMap} fileTreeMap map from filenames to ownership subtrees.
-   * @return {Object<string, string[]>} map from filenames to user/team names.
+   * @return {Object<string, string[]>} map from user/team names to filenames.
    */
   _getNotifies(fileTreeMap) {
     const notifies = {};
     Object.entries(fileTreeMap).forEach(([filename, subtree]) => {
-      const fileNotifies = subtree
+      subtree
         .getModifiedOwners(OWNER_MODIFIER.NOTIFY)
-        .map(owner => owner.name);
-      if (fileNotifies.length) {
-        notifies[filename] = fileNotifies;
-      }
+        .map(owner => owner.name)
+        .forEach(name => {
+          if (!notifies[name]) {
+            notifies[name] = [];
+          }
+          notifies[name].push(filename);
+        });
     });
 
     return notifies;
--- a/owners/src/owners_bot.js
+++ b/owners/src/owners_bot.js
@@ -99,7 +99,6 @@ class OwnersBot {
     const ownersCheck = new OwnersCheck(tree, changedFiles, reviewers);
 
     const checkRunIdMap = await github.getCheckRunIds(pr.headSha);
-    // TODO(rcebulko): Make this into a loop through multiple check/name pairs.
     const checkRunId = checkRunIdMap[OWNERS_CHECKRUN_NAME];
     const ownersCheckResult = ownersCheck.run();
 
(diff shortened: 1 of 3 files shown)


owners/test/owners_bot.test.js
```diff
@@ -215,11 +223,21 @@ describe('owners bot', () => {
...
       });
     });
+
+    it('creates a notification comment', async () => {
+      expect.assertions(1);
+      getCheckRunIdsStub.returns({});
+      sandbox.stub(OwnersBot.prototype, 'createNotifications');
+      await ownersBot.runOwnersCheck(github, pr);
+
+      sandbox.assert.calledOnce(ownersBot.createNotifications);
```
estherkim: What if we used `expect(ownersBot.createNotifications).toHaveBeenCalled(1)`? Could we remove `expect.assertions(1)` and `expect(true).toBe(true)`?

For reference, I use this when testing async code in the pr-deploy tests.

rcebulko: Throughout the project, both when Erwin owned it and after I took over, we've been using `sinon` sandboxes for mocking and stubbing. I'm not inclined to mix in Jest mock functions in this PR, and this is not the first instance of the `expect(true).toBe(true)` pattern. We could go through and rewrite all the tests with `jest` mocks and remove `sinon` (assuming Jest provides similar sandboxing features, which I'm not certain of), which would let us use 'expect...toHaveBeenCalled`, but the documentation on that generally seems to indicate it's for use with explicitly mocked callback functions whose invocations need to be validated.

estherkim: A rewrite sounds time consuming, but do you agree that these patterns can be improved?
1. Using 2 types of assertions (jest and sinon) in a test suite
2. Using `expect(true).toBe(true)` to determine that `sinon.assert.calledOnce(spy)` ran

rcebulko: Sinon is designed to be used in another test suite, be it mocha or jest or something else, so I don't see an issue with that. I don't like the expects there, but sinon doesn't have a similar expect.assertions method. I'm not sure what the best solution would be.

estherkim: One solution would be to use jest mocks and jest assertions. What do you think?

rcebulko: #447

rcebulko: Tests added in this PR have been updated. The other tests in this project will be updated in a later PR

rcebulko: #454

owners/src/owners_bot.js
```diff
@@ -125,6 +129,34 @@ class OwnersBot {
...
+  /**
+   * Adds a comment tagging always-notify owners of changed files.
+   *
+   * @param {!GitHub} github GitHub API interface.
+   * @param {!number} prNumber pull request number.
+   * @param {!FileTreeMap} fileTreeMap map from filenames to ownership subtrees.
+   */
+  async createNotifications(github, prNumber, fileTreeMap) {
+    const botComments = await github.getBotComments(prNumber);
+    if (botComments.length) {
```
estherkim: What if a file was added to a PR? Wouldn't you have to modify the existing bot comment?

owners/src/owners_bot.js
```diff
@@ -125,6 +129,34 @@ class OwnersBot {
...
 
+  /**
+   * Adds a comment tagging always-notify owners of changed files.
+   *
+   * @param {!GitHub} github GitHub API interface.
+   * @param {!number} prNumber pull request number.
+   * @param {!FileTreeMap} fileTreeMap map from filenames to ownership subtrees.
+   */
+  async createNotifications(github, prNumber, fileTreeMap) {
+    const botComments = await github.getBotComments(prNumber);
```
estherkim: Do you want the owners bot to leave 1 comment overall or multiple?

owners/test/owners_bot.test.js
estherkim: Thanks for following up!

owners/src/owners_bot.js
```diff
@@ -125,6 +129,34 @@ class OwnersBot {
...
+      // Avoid adding duplicate notification comment.
+      return;
+    }
+
+    const notifies = this._getNotifies(fileTreeMap);
+
+    const tag = name => `@${name}`;
+    const header = 'Some owners are notified of changes to files in this PR:';
+    const fileNotifyComments = Object.entries(notifies).map(
+      ([filename, names]) => `- ${filename}: ${names.map(tag).join(', ')}`
```
estherkim: Can we update the comment body to be more user friendly, and only mention an owner once? Something like 
```
Hey owner1, these files were changed
- file a
- file b

Hey owner2, these files were changed:
- file c

owners/src/owners_bot.js
rcebulko: For now one, but leaving open the option of there being multiple

owners/src/owners_bot.js
rcebulko: Adding a TODO on that for now, would like to address that case (along with cases where files are removed or the OWNERS file changes) separately

owners/src/owners_bot.js
rcebulko: Soitenly! Done.

owners/test/owners_bot.test.js
rcebulko: Thanks for pointing it out, would never have even found that this feature exists

owners/src/owners_bot.js
rsimha: This comment by @estherkim and the one above it raise good points. In both cases, I think updating the same comment will be cleaner, more accurate, and avoid comment spam. (For example, large PRs that gradually end up touching several files.)

rcebulko:
@rsimha Updating the existing comment seems like the right way to go. I'll follow up with PRs to delete the old comment and add the new one.

rsimha:
> For posterity's sake and to avoid spamming PRs, let's not delete the comment, but update the existing one.

Agree.

Pretty sure this is Github's behavior (worth double-checking):

- Editing a comment in-place will only notify newly @mentioned folks
- Deleting and re-adding the same comment + changes will re-notify _everyone_ in the new comment.

PR #448, opened 2019-09-19
[owners] Add backwards compatibility for comma-separated rules without braces

rcebulko:
Pull #437 added support for brace-set expansion, but that broke backwards-compatibility with comma-delimited rules since they could no longer be split by commas. Since the comma-separated form is already in use in the repo, this PR reverts to the original comma-separated functionality for rules which do not contain a brace `{}`, allowing existing comma-delimited rules to remain. Once this is submitted and deployed, I will update existing rules to use the brace-set syntax and remove this backwards compatibility (it is not documented in the [example file](https://github.com/ampproject/amp-github-apps/blob/master/owners/OWNERS.example.yaml))

### DIFF
--- a/owners/src/parser.js
+++ b/owners/src/parser.js
@@ -108,7 +108,7 @@ class OwnersParser {
       owner = owner.slice(1);
     }
 
-    if (owner.indexOf('/') !== -1) {
+    if (owner.includes('/')) {
       const team = this.teamMap[owner];
 
       if (team) {
@@ -187,7 +187,7 @@ class OwnersParser {
         pattern = pattern.slice(GLOB_PATTERN.length);
       }
 
-      if (pattern.indexOf('/') !== -1) {
+      if (pattern.includes('/')) {
         errors.push(
           new OwnersParserError(
             ownersPath,
--- a/owners/src/parser.js
+++ b/owners/src/parser.js
@@ -167,53 +167,64 @@ class OwnersParser {
    *     rule.
    */
   _parseOwnersDict(ownersPath, ownersDict) {
-    let [[pattern, ownersList]] = Object.entries(ownersDict);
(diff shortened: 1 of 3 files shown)


owners/src/parser.js
```diff
@@ -164,53 +164,64 @@ class OwnersParser {
    *     rule.
    */
   _parseOwnersDict(ownersPath, ownersDict) {
-    let [[pattern, ownersList]] = Object.entries(ownersDict);
+    const [[fullPattern, ownersList]] = Object.entries(ownersDict);
     const rules = [];
     const errors = [];
+    let patternList = [];
```
rcebulko: Don't let the diff fool you; all of the parsing logic is just indented into a `forEach` loop for each subpattern, the same way it was before #437 .

rsimha: Pro-tip: Adding `?w=1` to the diff URL hides all whitespace-only changes.

https://github.com/ampproject/amp-github-apps/pull/448/files?w=1

owners/src/parser.js
```diff
@@ -167,53 +167,64 @@ class OwnersParser {
...
-            `directory patterns other than '${GLOB_PATTERN}' not supported`
-        )
-      );
-    } else if (typeof ownersList === 'string') {
-      const lineResult = this._parseOwnersLine(ownersPath, ownersList);
-      owners.push(...lineResult.result);
-      errors.push(...lineResult.errors);
+    // TODO(rcebulko): Remove backwards-compatibility once all owners files have
+    // been updated to use brace syntax.
+    if (fullPattern.indexOf('{') === -1) {
```
rsimha: I believe this supports ES6 syntax. This can be rewritten as `!fullPattern.includes('{')`.

PR #451, opened 2019-09-19
[owners] Add a "require review" modifier

rcebulko:
In some instances (such as validator files), certain types of files or directories should require reviews even when they have an owner's approval. In the case of validator files, even though they may live within individual extensions and could be approved by low-level owners, they should require approval from a member of wg-caching before submission. This PR adds that "gatekeeper" modifier.

### DIFF
--- a/owners/src/parser.js
+++ b/owners/src/parser.js
@@ -93,6 +93,9 @@ class OwnersParser {
     } else if (owner.startsWith('#')) {
       modifier = OWNER_MODIFIER.NOTIFY;
       owner = owner.slice(1);
+    } else if (owner.startsWith('!')) {
+      modifier = OWNER_MODIFIER.REQUIRE;
+      owner = owner.slice(1);
     }
 
     if (owner.startsWith('@')) {
--- a/owners/src/parser.js
+++ b/owners/src/parser.js
@@ -90,7 +90,7 @@ class OwnersParser {
     if (owner.startsWith('?')) {
       modifier = OWNER_MODIFIER.SILENT;
       owner = owner.slice(1);
-    } else if (owner.startsWith('!')) {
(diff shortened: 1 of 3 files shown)


owners/src/parser.js
```diff
@@ -90,9 +90,12 @@ class OwnersParser {
     if (owner.startsWith('?')) {
       modifier = OWNER_MODIFIER.SILENT;
       owner = owner.slice(1);
-    } else if (owner.startsWith('!')) {
+    } else if (owner.startsWith('#')) {
       modifier = OWNER_MODIFIER.NOTIFY;
       owner = owner.slice(1);
+    } else if (owner.startsWith('!')) {
+      modifier = OWNER_MODIFIER.REQUIRE;
```
rsimha: I assume the code that processes this modifier is in a different PR?

owners/src/parser.js
```diff
@@ -90,9 +90,12 @@ class OwnersParser {
     if (owner.startsWith('?')) {
       modifier = OWNER_MODIFIER.SILENT;
       owner = owner.slice(1);
-    } else if (owner.startsWith('!')) {
+    } else if (owner.startsWith('#')) {
```
rsimha: Are there any existing uses of `!` that need to be changed to `#`?

owners/src/parser.js
rcebulko: Yes; this is the parser-only side. I was working on them together in the same branch, but decided to split out this part since it's self-contained.

owners/src/parser.js
rcebulko: Nope, it's not yet documented, nor is it used, nor is the functionality of always-notify itself submitted yet (#443 ) so this change is safe to make.

PR #452, opened 2019-09-19
[owners] Add helper methods to get rules and owners for a file

rcebulko:


owners/src/owners_tree.js
```diff
@@ -87,6 +87,31 @@ class OwnersTree {
...
+  /**
+   * Provides a list of owners rules for a file.
+   *
+   * @param {!string} filename file to get owners for.
+   * @return {Owner[]} list of all owners for the file.
+   */
+  fileOwners(filename) {
+    return this.fileRules(filename)
+      .map(rule => rule.owners)
+      .reduce((left, right) => left.concat(right), []);
```
estherkim: Do you need this? `Array.prototype.map()` already returns an array.

owners/src/owners_tree.js
```diff
@@ -129,21 +154,20 @@ class OwnersTree {
...
-          if (!modifiedOwners[owner.name]) {
-            modifiedOwners[owner.name] = owner;
-          }
-        });
-    });
+    this.fileOwners(filename)
+      .filter(owner => owner.modifier === modifier)
+      .forEach(owner => {
+        if (!modifiedOwners[owner.name]) {
+          modifiedOwners[owner.name] = owner;
```
estherkim: `modifiedOwners` would work nicely as a map here.

owners/src/owners_tree.js
```diff
@@ -156,12 +180,9 @@ class OwnersTree {
...
    */
   fileHasOwner(filename, username) {
-    const allRules = this.atPath(filename).allRules;
-    const fileRules = allRules.filter(rule => rule.matchesFile(filename));
-
-    return fileRules.some(rule =>
-      rule.owners.some(owner => owner.includes(username))
-    );
+    return this.atPath(filename)
+      .fileRules(filename)
```
estherkim: I'm guessing not, but can you use `this.fileOwners(filename)` instead?

owners/src/owners_tree.js
rcebulko: `rule.owners` is an array, this flattens the array of arrays.

owners/src/owners_tree.js
rcebulko: You're right, I can! Done.

owners/src/owners_tree.js
rcebulko: The issue with that, or even with adding to a set, is that there may be multiple distinct `Owner` instances which refer to the same user or team, and there is no way of overloading the equality operator used for set comparison in ES6. So this creates a dict keyed by user/team name, which ensures no duplicates. For example, something like

```
return Array.from(
      new Set(
        this.fileOwners(filename)
        .filter(owner => owner.modifier === modifier)
        .forEach(modifiedOwners.add, modifiedOwners)
      )
    );
```

Would fail `owners tree > getModifiedFileOwners > does not return duplicate owners`

owners/src/owners_tree.js
estherkim: Ah okay. No need to check for dupes?

owners/src/owners_tree.js
estherkim: A map is a dict though? I'm thinking something like
```suggestion
const modifiedOwners = new Map();
this.fileOwners(filename).forEach(owner => {
    if (owner.modifier === modifier) {
        modifiedOwners.set(owners.name, owner);    
    }
});
```

estherkim: Whoa, first time using a code suggestion. Interesting

owners/src/owners_tree.js
rcebulko: Not here, dupes aren't relevant because even if identical rules exist they're at different precedence, so order/depth is important

owners/src/owners_tree.js
rcebulko: I can make this change. What's the benefit of Map over Objects?

estherkim: Could be helpful if you need to iterate by `owner.name` later on. More comparisons here -
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#Objects_and_maps_compared

rcebulko: (I thought you meant the may function)

estherkim: lol doh

rcebulko: Done!

PR #453, opened 2019-09-19
[owners] Request reviewers if PR description contains #addowners

rcebulko:
In order to start testing out reviewer assignment without fully enabling it everywhere, this PR makes the Owners Bot use the PR description as a gatekeeper. If and only if the description contains `#addowners`, it will assign reviewers to the PR.

### DIFF
--- a/owners/test/owners_bot.test.js
+++ b/owners/test/owners_bot.test.js
@@ -220,7 +220,7 @@ describe('owners bot', () => {
       });
     });
 
-    it('does not create review requests', async (done) => {
+    it('does not create review requests', async done => {
       await ownersBot.runOwnersCheck(github, pr);
 
       sandbox.assert.notCalled(github.createReviewRequests);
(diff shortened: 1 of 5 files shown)


owners/test/owners_bot.test.js
```diff
@@ -220,6 +219,32 @@ describe('owners bot', () => {
...
+    it('does not create review requests', async done => {
+      await ownersBot.runOwnersCheck(github, pr);
+
+      sandbox.assert.notCalled(github.createReviewRequests);
+      done();
+    });
+
+    describe('when the PR description contains #addowners', () => {
+      beforeEach(() => {
+        pr.description = 'Assign reviewers please #addowners';
```
rsimha: Might be worth adding a test where the description contains embedded HTML and / or code tags.

rcebulko: Discussed offline: The worst case is not a particularly bad case (reviewers get assigned when we don't particularly expect them to), and even that is rare.

PR #457, opened 2019-09-19
[owners] Disable stdout logging stream during test

rcebulko:
Bunyan logging evidently needs special treatment in order to properly log to GCE, but the code implementing this was still running during tests and made test output excessively verbose. This PR gates the `app.log.target.addStream` call, skipping if the node environment is `'test'`

### DIFF
--- a/owners/test/github.test.js
+++ b/owners/test/github.test.js
@@ -180,7 +180,10 @@ describe('GitHub API', () => {
 
   describe('customRequest', () => {
     beforeEach(() => {
-      sandbox.stub(process, 'env').value({GITHUB_ACCESS_TOKEN: '_TOKEN_'});
+      sandbox.stub(process, 'env').value({
+        ...process.env,
+        GITHUB_ACCESS_TOKEN: '_TOKEN_',
+      });
     });
 
     it('returns the response', async () => {
--- a/owners/test/github.test.js
+++ b/owners/test/github.test.js
@@ -181,8 +181,8 @@ describe('GitHub API', () => {
   describe('customRequest', () => {
     beforeEach(() => {
       sandbox.stub(process, 'env').value({
-        ...process.env,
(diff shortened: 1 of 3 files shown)


owners/test/github.test.js
```diff
@@ -180,7 +180,10 @@ describe('GitHub API', () => {
 
   describe('customRequest', () => {
     beforeEach(() => {
-      sandbox.stub(process, 'env').value({GITHUB_ACCESS_TOKEN: '_TOKEN_'});
+      sandbox.stub(process, 'env').value({
+        ...process.env,
```
rsimha: Would it not work to set `NODE_ENV` to `test` here instead of using the spread operator?

owners/test/index.test.js
```diff
@@ -62,7 +62,7 @@ describe('owners bot', () => {
 
   beforeEach(() => {
     sandbox = sinon.createSandbox();
-    sandbox.stub(process, 'env').value({GITHUB_REPO_DIR});
+    sandbox.stub(process, 'env').value({...process.env, GITHUB_REPO_DIR});
```
rsimha: Here too.

owners/test/github.test.js
rcebulko: It would work equally well. It seemed reasonable to inherit env vars (like LOG_LEVEL), and then the individual stubs wouldn't have to know about stubbing that particular env var (it's handled by probot, and here we're just overwriting process.env). So, this is taking the existing process.env and stubbing the particular property we want, rather than the whole environment. I could go either way on which is preferable.

rsimha: My preference is to stub just the `NODE_ENV` variable in this PR. This way, it will be an explicit decision to stub another variable in future.

rcebulko: Done.

PR #460, opened 2019-09-20
[owners] Prioritize more specific rules during reviewer selection.

rcebulko:
In cases where pattern rules apply to files, owners of those patterns should be suggested before owners of the whole directory. This is especially relevant in cases of top-level pattern rules, which are currently being treated as equal with top level owners. This PR updates reviewer selection so that when multiple types of rules are present at the same depth, the most specific rules are the ones used to choose the next reviewer.

### DIFF
--- a/owners/src/rules.js
+++ b/owners/src/rules.js
@@ -34,6 +34,7 @@ class OwnersRule {
     this.filePath = ownersPath;
     this.dirPath = path.dirname(ownersPath);
     this.owners = owners;
+    this.priority = 1;
   }
 
   /**
@@ -97,6 +98,7 @@ class PatternOwnersRule extends OwnersRule {
       nocomment: true,
       nonegate: true,
     });
+    this.priority = 2;
   }
 
   /**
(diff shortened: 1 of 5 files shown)


owners/src/rules.js
```diff
@@ -34,6 +34,7 @@ class OwnersRule {
     this.filePath = ownersPath;
     this.dirPath = path.dirname(ownersPath);
     this.owners = owners;
+    this.priority = 1;
```
rsimha: For improved clarity, I think an enum should be used for the priorities, with names like `TOP_LEVEL_OWNER`, `WILDCARD_OWNER`, `PATTERN_OWNER`, etc. This will also make the tests easier to read and maintain.

rcebulko: Done.

owners/src/rules.js
```diff
@@ -17,6 +17,12 @@
 const path = require('path');
 const minimatch = require('minimatch');
 
+const RULE_PRIORITY = {
```
rsimha: Add a comment above this for documentation.

PR #463, opened 2019-09-23
[owners] Add pending reviewers to the current reviewer set

rcebulko:
Ensures that the bot does not suggest new reviewers when the pending reviewers could provide owners coverage. The logic for this had been written already, it was just never wired together.

### DIFF
--- a/owners/test/owners_bot.test.js
+++ b/owners/test/owners_bot.test.js
@@ -100,6 +100,9 @@ describe('owners bot', () => {
           {filename: 'changed_file1.js', sha: '_sha1_'},
           {filename: 'foo/changed_file2.js', sha: '_sha2_'},
         ]);
+      sandbox.stub(GitHub.prototype, 'getReviewRequests').returns([
+        'requested'
+      ]);
     });
 
     it('parses the owners tree', async () => {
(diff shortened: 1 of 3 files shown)


owners/test/owners_bot.test.js
```diff
@@ -160,6 +170,7 @@ describe('owners bot', () => {
       sandbox.stub(GitHub.prototype, 'createCheckRun');
       sandbox.stub(GitHub.prototype, 'getReviews').returns([]);
       sandbox.stub(GitHub.prototype, 'listFiles').returns([]);
+      sandbox.stub(GitHub.prototype, 'getReviewRequests').returns([]);
```
estherkim: Can we add a use case where `pendingReviewers` exist?

rcebulko: There already is one, which directly stubs `_getReviewRequests`. This just blocks all the other tests from making the network request

rcebulko: That is tested within OwnersCheck (which is what receives the map of reviewer approvals)

rcebulko: Also relevant, I'm planning on splitting the actual running of the owners check out from creating the notifications and adding reviewers, so the notifs/review requests can be triggered independently by #addowners comments. When that happens, it will come along with more tests around how existing reviewers are used. But right now, since the Owners Check is the only consumer of the reviewer statuses, any tests that could be added here would be duplicating tests already in owners check.

PR #477, opened 2019-09-25
[owners] Ignore irrelevant review states from GitHub

### LINKED ISSUE #478: [owners] Post-approval comments seen as rejecting reviews
I noticed that https://github.com/ampproject/amphtml/pull/24730 state "needs approval" for a file despite the requester already having approved the PR. Might be because new commits were added?

```
Current Coverage
- [NEEDS APPROVAL] src/service/extensions-impl.js (requested: jridgewell)
- test/unit/test-extensions.js (choumx)
```

https://github.com/ampproject/amphtml/pull/24730/checks?check_run_id=236887665

rcebulko:
In looking at recent PRs in `ampproject/amphtml`,  I noticed that https://github.com/ampproject/amphtml/pull/24686 listed `@rsimha` as "requested" but not approved, even though he'd given an approving review already. Upon digging into this, I found that the GitHub Reviews API, in addition to returning reviews with expect states (ie. approved & changes requested), it was also including review comments with a "commented" state. This was not documented in the [REST API docs](https://developer.github.com/v3/pulls/reviews/), but is mentioned in the [GraphQL API docs](https://developer.github.com/v4/enum/pullrequestreviewstate/).

As a result, the logic which looks at the most recent review was mistakenly identifying comments as rejecting reviews. Upon reviewing the remaining possible states (pending & dismissed), I determined that neither should be considered at actual review; a dismissed review should not count either way, and a pending review is the same.

This PR updates the GitHub `getReviews` method to return only actual reviews with relevant review states.

Closes #478

### DIFF
--- a/owners/src/github.js
+++ b/owners/src/github.js
@@ -244,17 +244,17 @@ class GitHub {
     // possible review states. The only ones we care about are "APPROVED" and
     // "CHANGES_REQUESTED", since the rest do not indicate a definite approval
     // or rejection.
-    const allowedStates = ['approved', 'changes_requested']
+    const allowedStates = ['approved', 'changes_requested'];
     return response.data
-        .filter(({state}) => allowedStates.includes(state.toLowerCase()))
-        .map(
-          json =>
-            new Review(
-              json.user.login.toLowerCase(),
-              json.state,
-              new Date(json.submitted_at)
-            )
-        );
+      .filter(({state}) => allowedStates.includes(state.toLowerCase()))
+      .map(
+        json =>
+          new Review(
+            json.user.login.toLowerCase(),
+            json.state,
+            new Date(json.submitted_at)
+          )
+      );
   }
 
   /**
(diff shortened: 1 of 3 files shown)


owners/src/github.js
```diff
@@ -240,14 +240,21 @@ class GitHub {
...
-          new Date(json.submitted_at)
-        )
-    );
+    // See https://developer.github.com/v4/enum/pullrequestreviewstate/ for
+    // possible review states. The only ones we care about are "APPROVED" and
+    // "CHANGES_REQUESTED", since the rest do not indicate a definite approval
+    // or rejection.
+    const allowedStates = ['approved', 'changes_requested'];
+    return response.data
+      .filter(({state}) => allowedStates.includes(state.toLowerCase()))
```
erwinmombay: just a qq, do states actually change in case or are we just being defensive here cause well, github is never consistent LOL

rcebulko: The are ''always'' capitalized lol. Being defensive here.

PR #489, opened 2019-10-01
[owners] Expose status/info pages on separate port

rcebulko:
Exposes three endpoints: the status page; the ownership tree display; and the list of teams known by the app

Since none of these endpoints utilize the GitHub API, there is minimal risk to exposing them. By placing them on a dedicated server/port, the GCE firewall can be configured to allow traffic directly to that port.

### DIFF
--- a/owners/package.json
+++ b/owners/package.json
@@ -19,7 +19,7 @@
     "lodash": "4.17.15",
     "minimatch": "3.0.4",
     "probot": "7.5.0",
-    "server": "^1.0.19",
+    "server": "1.0.19",
     "sleep-promise": "8.0.1",
     "yamljs": "0.3.0"
   },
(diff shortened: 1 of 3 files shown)


owners/package.json
```diff
@@ -19,6 +19,7 @@
     "lodash": "4.17.15",
     "minimatch": "3.0.4",
     "probot": "7.5.0",
+    "server": "^1.0.19",
```
rsimha: Use exact versions (without the `^`).

For reference, see this [check](https://github.com/ampproject/amphtml/blob/master/build-system/tasks/check-exact-versions.js) in `amphtml`.

danielrozenberg: As a side-note, renovate-bot auto-fixes that in a followup PR, so it's not a disaster if we miss this :)

rcebulko: Well I removed it so TAKE THAT RENOVATE-BOT

PR #490, opened 2019-10-02
[owners] Only request reviewers on new pull requests

### LINKED ISSUE #488: [owners] Owners bot re-adds reviewers after removal
Since the owners bot is webhooked to run after new commits or review status changes, it triggers when someone removes a reviewer, causing the bot to immediately reassign the reviewer. This could also be an issue when commits are added later on, re-adding previously removed reviewers. The desired behavior is instead that reviewer assignment should only occur upon PR creation. This means that you couldn't trigger assignment after the fact, but that's a reasonable trade-off since the owners check will still show suggested reviewers if more are necessary.

Example: https://github.com/ampproject/amphtml/pull/24388

rcebulko:
Since the owners bot is webhooked to run after new commits or review status changes, it triggers when someone removes a reviewer, causing the bot to immediately reassign the reviewer. This could also be an issue when commits are added later on, re-adding previously removed reviewers. The desired behavior is instead that reviewer assignment should only occur upon PR creation. This means that you couldn't trigger assignment after the fact, but that's a reasonable trade-off since the owners check will still show suggested reviewers if more are necessary.

Closes #488

### DIFF
--- a/owners/index.js
+++ b/owners/index.js
@@ -116,7 +116,7 @@ module.exports = app => {
     await ownersBot.runOwnersCheck(
       GitHub.fromContext(context),
       PullRequest.fromGitHubResponse(context.payload.pull_request),
-      /* requestOwners=*/ true
+      /* requestOwners */ true
     );
   });
 
--- a/owners/index.js
+++ b/owners/index.js
@@ -112,7 +112,15 @@ module.exports = app => {
   });
 
   /** Probot request handlers **/
-  app.on(['pull_request.opened', 'pull_request.synchronize'], async context => {
+  app.on(['pull_request.opened'], async context => {
+    await ownersBot.runOwnersCheck(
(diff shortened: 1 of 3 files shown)


owners/index.js
```diff
@@ -112,7 +112,15 @@ module.exports = app => {
   });
 
   /** Probot request handlers **/
-  app.on(['pull_request.opened', 'pull_request.synchronize'], async context => {
+  app.on(['pull_request.opened'], async context => {
+    await ownersBot.runOwnersCheck(
+      GitHub.fromContext(context),
+      PullRequest.fromGitHubResponse(context.payload.pull_request),
+      /* requestOwners=*/ true
```
danielrozenberg: nit: I think the common style here is to write `/* requestOwners */` without the equals sign. Not 100% sure if we have a style-guide rule for this

PR #494, opened 2019-10-02
[owners] Cache tree parse on public-facing status server endpoints

rcebulko:
Since the goal is to make the status server public-facing, it should not include any calls to rate-limited APIs or an filesystem I/O, as this presents a potential vulnerability to DOS/abuse. Currently, the only endpoint which has this issue is the `/tree` page, which displays the current ownership tree by parsing the locally checked-out repository on the GCE instance filesystem.

This PR instead creates a locally cached tree parse, which is updated on a regular (10 minute) interval, so that the public-facing endpoint does no filesystem I/O and just renders based on the in-memory tree parse.


owners/index.js
```diff
@@ -42,7 +44,7 @@ module.exports = app => {
     app.log
   );
   // TODO(rcebulko): Add a mechanism to periodically refresh teams.
```
rsimha: Can this TODO be removed?

rcebulko: Not yet, no

PR #498, opened 2019-10-03
[bundle-size] Stop handling gzip sizes

danielrozenberg:


bundle-size/app.js
```diff
@@ -562,51 +562,39 @@ module.exports = app => {
...
+    const bundleSizeText = `${brotliBundleSize}KB`;
+    const bundleSizeFile = `${headSha}.br`;
 
+    try {
+      await getBuildArtifactsFile(userBasedGithub, bundleSizeFile);
+      app.log(
+        `The file bundle-size/${bundleSizeFile} already exists in the ` +
+          'build artifacts repository on GitHub. Skipping...'
+      );
+    } catch (unusedException) {
```
rcebulko: It looks like this try block was moved into the `catch`; is that intentional? The comment that was removed seemed useful.

danielrozenberg: Yes, it's intentional - before that there was a `continue;` line above for when this needed to be skipped, but now since there's no for loop here, I had to move this block inside. I'll restore the comment...

danielrozenberg:
> IIRC there was a bit of churn/bugfixes when the brotli bit was added. Is there any way we can test that this will run as expected/won't break builds before we deploy it fully?

_This_ is the breaking change that I introduced yesterday 😅
Hopefully the changes in tests show that this fix is good?

Issue #442, opened 2019-09-18
[owners] Transition from YAML as owners file format


rcebulko:
YAML is tricky to write and inconsistent to parse, which can cause problems as contributors try to write owners rules. Some things need to be quoted, some don't, lists and dicts can be defined in many ways, and it's generally less familiar to people than JSON.

Plan:
- [x] Add support for parsing OWNERS.____; prefer the YAML form when present
- [x] Create script to take the parsed form of an OWNERS.yaml file and output the matching OWNERS.____
- [x] Generate & submit OWNERS.____ for all OWNERS.yaml files in `amphtml`
- [x] Remove all OWNERS.yaml files
- [ ] Remove support for YAML

/cc @kristoferbaxter

danielrozenberg:
JSON is hard for humans to write and read... How about protoascii? :D

rcebulko:
That seems like a good option, since contributors will likely be familiar with it anyway

mrjoro:
protoascii sounds good to me!

JSON also has the issue that it doesn't allow comments, which are sometimes useful in the files.

rsimha:
After chatting off line with @rcebulko, I too am in favor of protoascii.

rcebulko:
Example: current root OWNERS.yaml

```yaml
# For an explanation of the OWNERS.yaml rules and syntax, see:
# https://github.com/ampproject/amp-github-apps/blob/master/owners/OWNERS.example.yaml

- "?cramforce"
- "?dvoytenko"
- "?jridgewell"
- "**/validator-*.{protoascii,html,out}": "#ampproject/wg-caching"
- "*.md": ampproject/wg-outreach
- "{.*,{{babel,bundles}.config,gulpfile}.js,renovate.json}": ampproject/wg-infra
- "{package.json,yarn.lock}":
  - ampproject/wg-infra
  - ampproject/wg-runtime
  - ampproject/wg-performance

# TODO(#24685): Remove this once most owners files have been update and/or the
# Travis check is implemented.
- "**/OWNERS.yaml": "#rcebulko"
```

```protobuf
// For an explanation of the OWNERS.yaml rules and syntax, see:
// https://github.com/ampproject/amp-github-apps/blob/master/owners/OWNERS.example.yaml

rules: {
  owners: {
    name: "cramforce"
    notify: NEVER
  }
  owners: {
    name: "dvoytenko"
    notify: NEVER
  }
  owners: {
    name: "jridgewell"
    notify: NEVER
[...]

PR #505, opened 2019-10-04
[owners] Update parser to support JSON5 owners files

rcebulko:
Per #442 , this PR adds parser support for parsing owners files in [JSON5](https://json5.org) syntax. JSON5 is a superset of JSON which was designed to be less fragile and more human-readable and -writable. Importantly, it supports comments; unquoted keys; single-quotes; and trailing commas. All of these are useful to allow and reduce the fragility of owners files.

Most of the files touched by this PR are just removing the references to YAML syntax. The only real changes live in `parser.js` and `parser.test.js`, and as always it's got 100% test coverage, so I recommend using the tests as your guide of what's new and what it does.

All tests under the `JSON5 format` section below read directly from the example owners file. This is to ensure that the parser and reference file always stay in sync.

For a summary view of what the modified parser does, here is the Jest test output:
```
owners parser error
  toString
    ✓ displays the file containing the error
    ✓ displays the error message
owners parser
  parseOwnerDefinition
    when given something not an owner definition
      ✓ reports an error
      ✓ returns no result
    for a non-string name
      ✓ reports an error
      ✓ returns no result
    for a name with a leading "@"
      ✓ reports an error
      ✓ ignores the "@"
    modifiers
      ✓ defaults to no modifier
      ✓ selects always-notify modifier when "notify" is true
      ✓ selects never-notify modifier when "requestReviews" is false
      when multiple options are specified
        ✓ reports an error
        ✓ ignores the options
    team owner declarations
      ✓ returns a team owner
      when the team is unknown
        ✓ reports an error
        ✓ returns no result
    wildcard owner declarations
      ✓ parses a wildcard owner
      ✓ ignores modifiers
  parseRuleDefinition
    ✓ creates an owners rule
    when given something not a rule definition
      ✓ reports an error
      ✓ returns no result
    for a non-list owners property
      ✓ reports an error
      ✓ returns no result
    for a rule with no valid owners
      ✓ reports an error
      ✓ returns no result
    when a pattern is specified
      ✓ creates a same-directory rule
      ✓ creates a recursive rule for a pattern starting with **/
      for a non-string pattern
        ✓ reports an error
        ✓ returns no result
      for a directory glob pattern
        ✓ reports an error
        ✓ returns no result
  parseOwnersFile
    YAML format (**this section is old, ignore it and move onto JSON5 format***)
      ✓ reads the file from the local repository
      ✓ assigns the OWNERS directory path
      ✓ parses a YAML list
      ✓ parses a YAML list with blank lines and comments
      ✓ parses a wildcard owner
      ✓ handles and reports YAML syntax errors
      team rule declarations
        ✓ returns a rule with all team members as owners
        ✓ records an error for unknown teams
      owner with a leading @
        ✓ parses ignoring the @ sign
        ✓ records an error
      rule dictionary
        ✓ parses a single owner into a pattern rule
        ✓ parses a list of owners into a pattern rule
        ✓ reports errors for non-string owners
        ✓ starting with **/ parses into a recursive pattern rule
        ✓ parses no rule if no valid owners are listed
        ✓ reports an error for patterns containing illegal '/'
        ✓ parses brace-set glob patterns
        ✓ parses comma-separate patterns as separate rules
      files containing top-level dictionaries
        ✓ returns no rules
        ✓ returns a parser error
      owner modifiers
        user owner
          ✓ parses an always-notify `#` modifier
          ✓ parses a never-notify `?` modifier
          ✓ parses a require-review `!` modifier
        team owner
          ✓ parses an always-notify `#` team modifier
          ✓ parses a never-notify `?` team modifier
          ✓ parses a require-review `!` team modifier
        wildcard owner
          ✓ reports an error for an always-notify `#` modifier
          ✓ reports an error for a never-notify `?` modifier
          ✓ reports an error for a require-review `!` modifier
    JSON5 format (**these tests all use the example owners file**)
      ✓ reads the file from the local repository
      ✓ assigns the OWNERS directory path
      ✓ handles and reports JSON syntax errors
      ✓ handles and reports JSON files without a `rules` property
      for valid owners files
        ✓ parses basic owner rules
        ✓ parses filename rules
        ✓ parses pattern rules
        ✓ parses recursive pattern rules
        ✓ reports an error for directory glob patterns
        ✓ parses brace-set rules
        owner definitions
          ✓ parses user owners
          ✓ parses team owners
          ✓ parses owner modifiers
  parseAllOwnersRules
    ✓ reads all owners files in the repo
    ✓ does not include invalid rules
    ✓ collects errors from all parsed files
  parseOwnersTree
    ✓ adds each rule to the tree
    ✓ returns parser errors
```


owners/OWNERS.example
```diff
@@ -0,0 +1,124 @@
...
+ */
+
+// AMPHTML uses a custom owners bot to enforce ownership rules across its
+// codebase to ensure that all incoming pull requests are reviewed by the right
+// people. This file is an example of the supported format.
+
+// OWNERS files support JSON5 syntax (see json5.org), a more human-readable and
+// writable variant of JSON which supports comments, unquoted obeject keys,
+// single quotes, and trailing commas. In this file, "this directory" refers to
+// the directory containing the OWNERS.yaml file.
```
rsimha: I suggest changing this to a `@fileoverview` section with [block style](https://jsdoc.app/about-block-inline-tags.html#overview) comments.

owners/OWNERS.example
```diff
@@ -0,0 +1,124 @@
...
+
+// The result of parsing this file would be a tree with these rules:
+//
+// **/*: someuser, dontdothis, ampproject/wg-cool-team [members...],
+//       dvoytenko (never notify), rcebulko (always notify), rando
+// ./{package.json}: packager
+// ./{*.js}: scripter
+// **/{*.protoascii}: ampproject/wg-caching [members...]
+// ./{*.{js,css}}: frontend
+// ./{package.json,yarn.lock}: ampproject/wg-infra [members...]
```
rsimha: In addition to this and the comprehensive list of examples above it, I think it would be useful to include an example of a canonical "hello world" owners file with just one or two entries at the top of the file. WDYT?

owners/src/parser.js
```diff
@@ -138,16 +139,16 @@ class OwnersParser {
    * Parse a list of owners.
    *
    * @private
-   * @param {!string} ownersPath OWNERS.yaml file path (for error reporting).
+   * @param {!string} ownersPath OWNERS file path (for error reporting).
    * @param {string[]} ownersList list of owners.
    * @return {OwnersParserResult<Owner[]>} list of owners' usernames.
```
rsimha: Maybe add an [`@deprecated`](https://jsdoc.app/tags-deprecated.html) annotation to this and all other yaml parsing functions?

owners/OWNERS.example
rcebulko: Added a simple example in the fileoverview

owners/OWNERS.example
rcebulko: Done

PR #513, opened 2019-10-07
Set renovate-bot to update all packages in one PR, once a week

danielrozenberg:
I think this is preferable for this repository, given that we have a lot of dependencies that get a `x.x.1` version update every few days...

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,5 +1,10 @@
 {
   "extends": [
     "config:base"
-  ]
+  ],
+  "packageRules": [{
+    "packagePatterns": [".*"],
+    "groupName": "packages",
+    "schedule": ["after 9pm on sunday"]
+  }]
 }


renovate.json
```diff
@@ -1,5 +1,10 @@
 {
   "extends": [
     "config:base"
-  ]
+  ],
+  "packageRules": [{
+    "packagePatterns": [".*"],
+    "groupName": "packages",
+    "schedule": ["after 9pm on sunday"]
+  }]
```
rsimha: I think the only line you really need is the one with `schedule`. The file could be simplified to:
```
{
  "extends": [
    "config:base"
  ],
  "schedule": "after 9pm on sunday"
}
```

danielrozenberg: `packagePatterns` + `groupName` is what causes grouping of package updates. Without this, we'll get dozens of update PRs at the same time

rsimha: Aha, missed that part.

PR #514, opened 2019-10-07
[bundle-size] Read .json instead of .br file

danielrozenberg:


bundle-size/test/app.test.js
```diff
@@ -33,15 +34,13 @@ jest.setTimeout(30000);
...
- * via `require` directly will cause all future calls to `require(sameFile)` to
- * return the previously modified object.
- *
  * @param {!string} name name of the JSON fixture file (without .json).
  * @return {!object} the named JSON test fixture file.
  */
 function getFixture(name) {
-  return deepcopy(require(`./fixtures/${name}`));
+  return JSON.parse(
+    fs.readFileSync(path.join(__dirname, `fixtures/${name}.json`))
```
rcebulko: Nit: Since this is part of a test, if this `getFixture` gets called for each test case, it's a bunch of excessive file reads. While that's probably pretty minimal, it might make sense to instead read once at the top of the file into a constant and just repeatedly parse the constant string here.

danielrozenberg: Re-reading the same file ensures we get a fresh copy every time. That's why I had `deepcopy` before. I realized later that using `fs` instead of `require` makes more sense for test fixtures, and this lets us get rid of a dependency, at the cost you just mentioned of increased file reads. I think it's safe to let it slide in this case exactly because it's a test... :)

PR #515, opened 2019-10-07
[bundle-size] Create new API endpoint to receive bundleSizes as a JSON payload

danielrozenberg:
More changes included in this PR:
* Removes the text portion of the commit message now that the complete JSON is going to be included
* Reorganizes the tests into sub-sections using `describe`s. It might be easier to review these changes using a better diff program, such as Meld :) otherwise, trust that most of the test file changes is just shuffles (modulo required changes to the tests to cater to the aforementioned text removal), except for the addition of the `/commit/:headSha/store.json` test sub-section

### DIFF
--- a/bundle-size/app.js
+++ b/bundle-size/app.js
@@ -264,13 +264,13 @@ module.exports = app => {
     } catch (error) {
       const partialHeadSha = check.head_sha.substr(0, 7);
       const partialBaseSha = baseSha.substr(0, 7);
-      app.log(
+      app.log.error(
         'ERROR: Failed to retrieve the bundle size of ' +
           `${partialHeadSha} (PR #${check.pull_request_id}) with branch ` +
           `point ${partialBaseSha} from GitHub: ${error}`
       );
       if (lastAttempt) {
-        app.log('No more retries left. Reporting failure');
+        app.log.warn('No more retries left. Reporting failure');
         Object.assign(updatedCheckOptions, {
           conclusion: 'action_required',
           output: {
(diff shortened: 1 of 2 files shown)


bundle-size/app.js
```diff
@@ -649,4 +649,61 @@ module.exports = app => {
 
     response.end();
   });
+
+  // TODO(danielrozenberg): replace the /store with this one, once the amphtml
+  // repo is in sync with this new change.
+  v0.post('/commit/:headSha/store.json', async (request, response) => {
```
rcebulko: I don't know if I've ever seen a POST URL with a file extension... :thinking:

bundle-size/app.js
```diff
@@ -649,4 +649,61 @@ module.exports = app => {
...
+    }
+    if (
+      !bundleSizes ||
+      Object.values(bundleSizes).some(value => typeof value !== 'number')
+    ) {
+      return response
+        .status(400)
+        .end(
+          'POST request to /store must have a key/value object ' +
+            '(string->numeric) field "bundleSizes"'
```
rcebulko: Nit: `Map<string, number>` or similar JSDoc format might be clearer

bundle-size/app.js
```diff
@@ -649,4 +649,61 @@ module.exports = app => {
...
+      app.log(
+        `Stored the new bundle size file bundle-size/${jsonBundleSizeFile} ` +
+          'the artifacts repository on GitHub'
+      );
+    } catch (error) {
+      const errorMessage =
+        `ERROR: Failed to create the bundle-size/${jsonBundleSizeFile} file ` +
+        'in the build artifacts repository on GitHub!\n' +
+        `Error message was: ${error}`;
+      app.log(errorMessage);
```
rcebulko: Nit: Use `app.log.error` or `app.log.warn` in this file where appropriate

bundle-size/app.js
danielrozenberg: There's always a first time :P

bundle-size/app.js
danielrozenberg: Neat

bundle-size/app.js
danielrozenberg: Makes sense. Done.

Issue #516, opened 2019-10-08
[owners] Support required reviewers


rcebulko:
Much like how we have always-notify and no-notify owners, we should support requiring a review. This could be useful for allowing certain groups or individuals to be "gatekeepers" of especially sensitive files or directories. It would also double as a solution to #499 , since the `ampproject/reviewers` team could be added as required at the top level.

The owners check should fail even if all files have normal owners coverage, but at least one file does not have approval from a required reviewer (or member of a required team). Ideally reviewer selection will prioritize required reviewers when making suggestions, but if not, after normal reviewer selection if there are still required reviewers missing it can add on randomly from the required sets.

Issue #499, opened 2019-10-03
[owners] Only request reviews from members of `ampproject/reviewers-amphtml`


rcebulko:
Since some working groups contain owners without the ability to give approving reviews, it's possible people could be assigned as reviewers who can't actually provide the review/another reviewer would be needed. When suggesting reviewers, we should only suggest those in the `ampproject/reviewers-amphtml` group.

One concern to mind when implementing this is that this is a relatively `amphtml`-centric need, but the solution should be relatively repository-agnostic so that other repos utilizing the owners bot aren't adversely affected.

This shouldn't have any effect on ownership coverage; it should only filter out which users can be considered during the reviewer selection algorithm. Proposed approach:

- [ ]  Provide a `isAllowedReviewer(username) --> boolean` test function to the reviewer selection algorithm and filter when selecting reviewers
- [ ] Let `isAllowedReviewer` default to `username => true`
- [ ] Add a `ALLOWED_REVIEWERS_TEAM` env variable where a team can be supplied
- [ ] When `process.env.ALLOWED_REVIEWERS_TEAM` is present in the team list, provide `username => team.allUsernames.includes(username)` as the `isAllowedReviewer` test function (or some set-based variant)

@rsimha thoughts?

rsimha:
> When suggesting reviewers, we should only suggest those in the `ampproject/reviewers-amphtml` group.

Agree, this should absolutely be the behavior.

The proposed approach sounds good in general. Re: `ALLOWED_REVIEWERS_TEAM`, are env variables the only means of passing data to a bot instance? Or is there an alternative way to configure an instance via, say, GCE settings?

rcebulko:
Well...about that...

[...]

PR #520, opened 2019-10-08
[owners] Adds support for required reviewers

rcebulko:
This PR adds support for checking required reviewers. It does not include parser changes to allow specifying required reviewers.

For context, see #516 

~Notably, this will also address #499 , since we can require approval from `ampproject/reviewers-amphtml` on all files.~
^^^ This won't work since that would make all reviewers owners of all files =/

### DIFF
--- a/owners/test/owners_check.test.js
+++ b/owners/test/owners_check.test.js
@@ -257,22 +257,22 @@ describe('owners check', () => {
     });
   });
 
-  describe('hasOwnersApproval', () => {
+  describe('hasFullOwnersCoverage', () => {
     it('returns true if any approver is an owner of the file', () => {
       expect(
-        ownersCheck._hasOwnersApproval(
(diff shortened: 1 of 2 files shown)


owners/test/owners_check.test.js
```diff
@@ -257,22 +265,168 @@ describe('owners check', () => {
...
+            'foo/required/info.html',
+            ownersCheck.tree.atPath('foo/required/info.html')
+          )
+        ).toBe(false);
+      });
+
+      it('fails if required user owners are pending approval', () => {
+        ownersTree.addRule(
+          new OwnersRule('foo/required/OWNERS', [
+            new UserOwner('extra_reviewerr', OWNER_MODIFIER.REQUIRE),
```
danielrozenberg: nit: double r

Issue #522, opened 2019-10-08
[owners] Support top-level owners file `reviewerTeam` property


rcebulko:
Per #499 , there is a need to have the owners check fail if a PR has approval from the OWNERS of the file, but not from someone in `ampproject/reviewers-amphtml` (or a comparable reviewer whitelist for other repositories). By integrating this into the owners check, we enable the repository to allow write-access to anyone. 

The proposed solution is to allow a `reviewerTeam` property in the repository root OWNERS file, which can specify a GitHub team name. The owners check will report a failure for any PR unless at least one member of this team has reviewed it. 

In any non-root owners file, this key will report an error and be ignored. If no such key is present in the repository root owners file, the condition will not be included.

rcebulko:
:raised_hands:

PR #525, opened 2019-10-08
[owners] Add support for an optional allowed reviewer set

rcebulko:
Addresses #522 

This PR adds support for an optional reviewer set. It modifies the owners check to fail unless at least one approving review comes from a member of the "reviewer set". The reviewer set defaults to a wildcard owner, so by default it will do nothing. If a reviewer set rule is specified, it must be satisfied, or the check-run will have an `action_required` status with a message indicating that a reviewer is required.

This PR does not include any changes to reviewer suggestion logic, nor does it implement the actual parsing to produce a reviewer set rule. This will come in a future PR.

### DIFF
--- a/owners/src/owners_check.js
+++ b/owners/src/owners_check.js
@@ -103,14 +103,28 @@ class OwnersCheck {
           delete fileTreeMap[filename];
         }
       });
-      const passing = !Object.keys(fileTreeMap).length;
+      const prHasFullOwnersCoverage = !Object.keys(fileTreeMap).length;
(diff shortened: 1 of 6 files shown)


rcebulko:
@rsimha I plan to make that a part of the PR which actually adds it to the parser. In practice, it will just be a top-level `reviewerTeam: 'ampproject/reviewers-amphtml'`, but I'd rather discuss syntax and implementation in separate threads if that makes sense, kinda like frontend UI and backend

PR #526, opened 2019-10-08
[owners] Add support for parsing the reviewer team option

rcebulko:
Addresses #522 ; adds parser support for reviewer set added in #525 . Also includes update to example owners file.

### DIFF
--- a/owners/OWNERS.example
+++ b/owners/OWNERS.example
@@ -128,11 +128,19 @@
       pattern: "*.{js,css}",
       owners: [{ name: "frontend" }],
     },
-  ]
+  ],
+
+  // The root-level owners file--and ONLY the root-level owners file--may
+  // specify an optional `reviewerTeam` property, providing the name of a GitHub
+  // team. If present, the owners check will require that at least one member of
+  // this team has given approval. It is possible that one reviewer is both a
+  // member of this reviewer team and provides owners coverage.
+  reviewerTeam: 'ampproject/reviewers-amphtml'
 }
 
 // The result of parsing this file would be a tree with these rules:
 //
+// Reviewers: ampproject/reviewers-amphtml [members...]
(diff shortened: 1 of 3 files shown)


owners/OWNERS.example
```diff
@@ -128,11 +128,19 @@
...
+  // specify an optional `reviewerTeam` property, providing the name of a GitHub
+  // team. If present, the owners check will require that at least one member of
+  // this team has given approval. It is possible that one reviewer is both a
+  // member of this reviewer team and provides owners coverage.
+  reviewerTeam: 'ampproject/reviewers-amphtml'
 }
 
 // The result of parsing this file would be a tree with these rules:
 //
+// Reviewers: ampproject/reviewers-amphtml [members...]
```
rsimha: Sanity check: The list of members of `ampproject/reviewers-amphtml` is not be publicly visible. I assume this PR doesn't make it visible (via logs, etc.)?

rcebulko: Incorrect, it does make it visible (it shows the reviewer list in checkrun output if the pr is missing a reviewer). Should it be private?

rsimha: Yeah, I don't think the owners bot should leak info that is otherwise private. Would it be possible to have the bot use the list but not expose it?

rsimha: @mrjoro can comment on whether this is a valid concern or not.

mrjoro: The GitHub teams not being public is a bug that we're aiming to fix (primarily around the WGs but I think the same applies to the reviewers & collaborators list).  GitHub doesn't allow them to be public, but we're looking at exporting them to be visible on amp.dev instead.

https://github.com/ampproject/amp.dev/issues/1699

rcebulko: I was planning on also making them visible within the owner tree rendering that shows who is an owner of each directory

Issue #271, opened 2019-07-22
Keep track of bundle size for extensions via a non-blocking check


rsimha:
See https://github.com/ampproject/amphtml/issues/20843

rsimha:
@danielrozenberg could you move the work plan to a separate comment authored by you? Right now, I'm getting notified each time the comment is edited.

danielrozenberg:
lol sure, I'll remove it from the description and add it here:

Workplan:
- [x] submit and deploy #428
- [x] after #428 is deployed, create a PR that migrates all `.br` files to `.json` (with the proper format of course) → done at https://github.com/ampproject/amphtml-build-artifacts/pull/18
- [x] unrollback #412
- [x] create a new PR that reads bundle-sizes from the `.json` file instead of the `.br` file, submit and deploy it → done at #514
- [x] create a PR to receive a JSON payload with multiple bundle sizes when storing → done at #515
- [x] once #514 and #515 are deployed, update the amphtml repo to send to the JSON payload on master builds → done at https://github.com/ampproject/amphtml/pull/24955
- [x] do a switcharoo from /store.json to /store (both on bundle-size bot and in amphtml, make sure to keep in sync) → done at #519 and https://github.com/ampproject/amphtml/pull/24958 and #523
- [ ] start storing extensions and other runtime files for master push builds  → done at https://github.com/ampproject/amphtml/pull/25042
- [x] create a PR to receive a JSON payload with multiple bundle sizes when reporting → done at #527
- [ ] once deployed, update the amphtml repo to report the JSON payload on pull requests → done at https://github.com/ampproject/amphtml/pull/25042
- [ ] do a switcharoo from /report.json to /report (wait at least 2 weeks before doing this, to reduce changes of errors in straggling PRs)
- [ ] delete all `.br` and extensionless files from https://github.com/ampproject/amphtml-build-artifacts/tree/master/bundle-size
- [ ] decide how to report and block PRs that increase bundle size of any files other than `dist/v0.js`

danielrozenberg:
Closing since the "keep track" part is done. Work continues in #617

PR #527, opened 2019-10-09
Add a /report.json endpoint to receive a complete bundleSizes JSON object

danielrozenberg:
Related to #271

### DIFF
--- a/bundle-size/app.js
+++ b/bundle-size/app.js
@@ -564,6 +564,60 @@ module.exports = app => {
     }
   });
 
+  v0.post('/commit/:headSha/report.json', async (request, response) => {
+    const {headSha} = request.params;
+    const {baseSha, bundleSizes} = request.body;
+
+    if (typeof baseSha !== 'string' || !/^[0-9a-f]{40}$/.test(baseSha)) {
+      return response
+        .status(400)
+        .end('POST request to /report must have commit SHA field "baseSha"');
+    }
+    if (
+      !bundleSizes ||
+      Object.values(bundleSizes).some(value => typeof value !== 'number') ||
+      !bundleSizes['dist/v0.js']
+    ) {
+      return response
+        .status(400)
+        .end(
+          'POST request to /report.json must have a key/value object ' +
+            'Map<string, number> field "bundleSizes", with at least one key ' +
+            'set to "dist/v0.js"'
+        );
+    }
+
+    // TODO(#271, danielrozenberg): this is a transitionary solution. This API
+    // endpoint accepts a JSON with multiple bundle sizes, but for now it only
+    // processes the size of dist/v0.js. Once we determine how we want to report
+    // and block PRs based on other bundle sizes we can start processing them.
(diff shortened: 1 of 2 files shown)


bundle-size/app.js
```diff
@@ -564,6 +564,60 @@ module.exports = app => {
     }
   });
 
+  v0.post('/commit/:headSha/report.json', async (request, response) => {
+    const {headSha} = request.params;
+    const {baseSha, bundleSizes} = request.body;
+
+    if (typeof baseSha !== 'string' || !/^[0-9a-f]{40}$/.test(baseSha)) {
```
rsimha: Would it make sense to extract this into a `isValidSha` function?

bundle-size/app.js
```diff
@@ -564,6 +564,60 @@ module.exports = app => {
...
+
+    // TODO(#271, danielrozenberg): this is a transitionary solution. This API
+    // endpoint accepts a JSON with multiple bundle sizes, but for now it only
+    // processes the size of dist/v0.js. Once we determine how we want to report
+    // and block PRs based on other bundle sizes we can start processing them.
+    const bundleSize = bundleSizes['dist/v0.js'];
+
+    const check = await getCheckFromDatabase(headSha);
+    if (!check) {
+      return response.status(404).end(`${headSha} not in database`);
```
rcebulko: Nit: A bit more detail in the error message (what database) might be helpful

rsimha: If this message appears in the status check or  Travis logs, perhaps expand on what database is being referred to?

bundle-size/test/app.test.js
```diff
@@ -849,6 +852,491 @@ describe('bundle-size', () => {
...
 
+    describe('/commit/:headSha/report.json', () => {
+      let jsonPayload;
+
+      beforeEach(() => {
+        jsonPayload = {
+          baseSha: '5f27002526a808c5c1ad5d0f1ab1cec471af0a33',
+          bundleSizes: {
+            'dist/v0.js': 12.34,
+            'dist/amp4ads-v0.js': 11.22,
```
rsimha: Yay! (I know it’s just a test, but still.)

/cc @lannka

bundle-size/app.js
rcebulko: I've thought the same thing a few times, I feel like I've seen this in a handful of places in this app (though idr how many were the same N occurrences). Might be worth `isValidSha` or `assertValidSha` that throws

bundle-size/app.js
danielrozenberg: Done

bundle-size/app.js
danielrozenberg: Not really, this is only used once, and once again for a soon-to-be-deleted endpoint. I'd rather not extract this into a method in that case :)

bundle-size/app.js
danielrozenberg: Done

PR #529, opened 2019-10-09
[DO NOT SUBMIT] Add useless concat

rcebulko:


owners/src/github.js
```diff
@@ -105,6 +105,8 @@ class Team {
    */
   constructor(id, org, slug) {
     Object.assign(this, {id, org, slug, members: []});
+    const message = 'something something' + 'something';
```
danielrozenberg: The example I meant here is for the case where a newline _is_ required (to avoid line-too-long issues) but you forget to add a space inside the concat

rcebulko: (this had a newline but gulp fix removed it)

Issue #540, opened 2019-10-17
[owners] Migrate from GCE to GAE


rcebulko:
The original implementation of the Owners bot was based around a Google Compute Engine (GCE) deployment. This decision was made because GCE provides local filesystem I/O, which is useful for maintaining a copy of the repository, using command-line-fu to identify owners files, and reading the contents of those owners files. No other files are read, and none are written.

Using GCE introduces a handful of issues, most notably:
- deployment is fragile, takes down the app, and relies on a startup script that is not checked in
- there is no reliable way to manage environment variables (especially sensitive keys/tokens); they are just embedded in the startup script, and can't be checked into the repository used to deploy
- the team is less familiar with GCE than GAE, making maintenance and debugging harder
- firewall restrictions around GCE make SSH and debugging difficult
- the amount of bootstrapping and configuration embedded only in GCE instance metadata makes it hard/impossible for other projects to make use of the owners bot

As a resolution to these issues, the plan is to migrate to Google App Engine (GAE). This will be a lot of work up-front, but should make automated and continuous deployment and development of the app much easier and less error-prone going forward.

Steps needed to migrate to a GAE deployment process:
- [x] configure app to read from `.env` file; write the `.env` file within the startup script during migration
- [x] hoist a base `Repository` class out of the `LocalRepository` class, and implement a `RemoteRepository` subclass
  - [x] using the [GitHub code search API](https://developer.github.com/v3/search/#search-code), provide the `listOwnersFiles` method
  - [x] using the [GitHub file contents API](https://developer.github.com/v3/repos/contents/#get-contents), provide the `readFile` method
- [x] maintain a map from owners file paths to the SHA of their latest commit
- [x] implement a caching layer to prevent overloading GitHub API on file reads
- [x] lift the collection and parsing of owners files/the ownership tree out of the owners check and into a single server-shared tree
- [x] isolate bootstrapping code (ie. initializing teams and ownership trees)
- [x] add webhooks to catch updates to owners files and update the corresponding rules in the cache
- [x] create `app.yaml` and deploy to GAE
- [x] migrate team and tree updating code to use Cron tasks
- [x] spin down GCE version of app

/cc @rsimha @erwinmombay @danielrozenberg

rcebulko:
One concern/point to be aware of:

[...]

PR #544, opened 2019-10-17
[owners] Add GitHub interface method to search for files by name

rcebulko:
Part of #540


owners/src/github.js
```diff
@@ -436,7 +435,45 @@ class GitHub {
...
+    return response.data.map(({filename}) => filename);
+  }
+
+  /**
+   * Searches for files in a repo with a given name.
+   *
+   * See https://developer.github.com/v3/search/#search-code
+   *
+   * @param {!string} filename filename to search for.
+   * @return {FileRef[]} list of returned results.
```
danielrozenberg: You seem to be shifting a lot between `string` and `FileRef`. When do you actually need the SHA field?

rcebulko: I only need the SHA field when I'm requesting a file's contents from the GitHub API. I'd added FileRef when I thought I'd need to fetch files in a PR, and so I started returning the SHA in listFiles. Then, it was determined we didn't need that, so it became clunky to work with {filename, sha} instead of just strings. Now, I actually want file refs so I can fetch the contents. So, other methods use string or Array<string>, and FileRef is just here for search/fetching file contents

PR #549, opened 2019-10-18
[owners] JSDoc updates

rcebulko:
This PR:
- migrates uses of `{type[]}` to `{!Array<!type>}`
- adds `!` to `@return` declarations
- fixes some optional paramater docs

### DIFF
--- a/owners/src/github.js
+++ b/owners/src/github.js
@@ -36,7 +36,7 @@ class PullRequest {
   /**
    * Whether or not the pull request is open.
    *
-   * @return {!boolean} true if the pull request is open.
+   * @return {boolean} true if the pull request is open.
    */
   get isOpen() {
     return this.state.toLowerCase() === 'open';
--- a/owners/src/github.js
+++ b/owners/src/github.js
@@ -36,7 +36,7 @@ class PullRequest {
   /**
    * Whether or not the pull request is open.
    *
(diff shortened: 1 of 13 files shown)


owners/src/github.js
```diff
@@ -36,7 +36,7 @@ class PullRequest {
   /**
    * Whether or not the pull request is open.
    *
-   * @return {boolean} true if the pull request is open.
+   * @return {!boolean} true if the pull request is open.
```
danielrozenberg: I think we're not supposed to use `!` for scalar types, or for return types at all? /cc @rsimha

rcebulko: See I thought that too re: return types but then on another PR somewhere I got called out for never using it and I couldn't find any docs saying not to so I thought I hallucinated it :thinking:

rcebulko: As for scalars, you're right, will remove.

estherkim: Here's a closure compiler example with `@return {!Element}`
https://github.com/google/closure-compiler/wiki/Types-in-the-Closure-Type-System#typedefs

rsimha: Just scalar types. See [this](https://github.com/ampproject/amphtml/blob/master/build-system/eslint-rules/closure-type-primitives.js#L147-L167) custom `amphtml` rule.

PR #550, opened 2019-10-18
[owners] Define virtual repository using GitHub API to list/read files

rcebulko:
Part of #540 . This VirtualRepository class could be swapped in after this PR, but would still induce a heavy API load and is not natively rate limited. The plan is to rate-limit on startup, then let it fetch on-demand whenever files change. Regardless, this is just the repository class, which is still unused anywher.


owners/src/repo.js
```diff
@@ -61,6 +66,83 @@ class Repository {
...
+  }
+
+  /**
+   * Read the contents of a file from the repo.
+   *
+   * @param {string} relativePath file to read.
+   * @return {string} file contents.
+   */
+  async readFile(relativePath) {
+    const file = this._knownFiles.get(relativePath);
```
rcebulko: Note: Once #552 is reviewed and merged, a follow-up PR will update this to use the cache layers/cloud storage backing.

PR #557, opened 2019-10-22
[owners] Await app initialization before handling webhooks

rcebulko:
This is a small change to make room for the in-memory file cache to warm up from Cloud Storage on application startup.

Part of #540

### DIFF
--- a/owners/index.js
+++ b/owners/index.js
@@ -53,7 +53,7 @@ module.exports = app => {
    */
   function listen(events, cb) {
     app.on(events, async context => {
-      await appInitialized();
+      await appInitialized;
 
       let github;
       try {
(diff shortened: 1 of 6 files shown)


owners/index.js
```diff
@@ -31,15 +31,18 @@ const [GITHUB_REPO_OWNER, GITHUB_REPO_NAME] = GITHUB_REPO.split('/');
...
     GITHUB_REPO_NAME,
     app.log
   );
-  ownersBot.initTeams(sharedGithub);
+
+  const repo = new LocalRepository(process.env.GITHUB_REPO_DIR);
+  const ownersBot = new OwnersBot(repo);
+
+  const teamsInitialized = ownersBot.initTeams(sharedGithub);
+  const appInitialized = teamsInitialized.then(() => ownersBot.refreshTree());
```
rcebulko: `refreshTree` is async and therefore returns a promise, so `appInitialized` will wait for the team map to be initialized, then refresh the tree, the resolve and unblock any pending webhooks.

PR #558, opened 2019-10-22
[owners] Use cache in virtual repo and warm up cache

rcebulko:
This PR:
- adds the caching layers to the virtual repository
- adds a script to warm up the Cloud Storage cache with rate-limited requests to the GitHub API, to prevent overloading the API on app startup

Part of #540

### DIFF
--- a/owners/index.js
+++ b/owners/index.js
@@ -42,7 +42,9 @@ module.exports = app => {
   const ownersBot = new OwnersBot(repo);
 
   const teamsInitialized = ownersBot.initTeams(sharedGithub);
-  const appInitialized = teamsInitialized.then(() => ownersBot.refreshTree());
+  const appInitialized = teamsInitialized.then(
+    () => ownersBot.refreshTree(app.log)
+  );
 
   /**
    * Listen for webhooks and provide handlers with a GitHub interface and the
@@ -114,7 +116,7 @@ module.exports = app => {
     );
 
     if (changedOwners.length) {
-      await ownersBot.refreshTree(changedOwners, github.logger);
+      await ownersBot.refreshTree(github.logger);
     }
   });
 
--- a/owners/index.js
+++ b/owners/index.js
@@ -42,8 +42,8 @@ module.exports = app => {
   const ownersBot = new OwnersBot(repo);
 
   const teamsInitialized = ownersBot.initTeams(sharedGithub);
-  const appInitialized = teamsInitialized.then(
-    () => ownersBot.refreshTree(app.log)
+  const appInitialized = teamsInitialized.then(() =>
+    ownersBot.refreshTree(app.log)
   );
 
   /**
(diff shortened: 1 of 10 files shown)


owners/index.js
```diff
@@ -34,7 +34,7 @@ module.exports = app => {
   const localRepo = new LocalRepository(process.env.GITHUB_REPO_DIR);
   const ownersBot = new OwnersBot(localRepo);
   const sharedGithub = new GitHub(
-    new Octokit({auth: `token ${GITHUB_ACCESS_TOKEN}`}),
+    new Octokit({auth: GITHUB_ACCESS_TOKEN}),
```
danielrozenberg: 😬

owners/scripts/warm_cache.js
```diff
@@ -0,0 +1,73 @@
...
+
+const {GitHub} = require('../src/github');
+const {VirtualRepository} = require('../src/repo');
+const {CompoundCache} = require('../src/file_cache');
+const {OwnersParser} = require('../src/parser');
+
+const CLOUD_STORAGE_BUCKET = process.env.CLOUD_STORAGE_BUCKET;
+const GITHUB_ACCESS_TOKEN = process.env.GITHUB_ACCESS_TOKEN;
+const GITHUB_REPO = process.env.GITHUB_REPO || 'ampproject/amphtml';
+const [GITHUB_REPO_OWNER, GITHUB_REPO_NAME] = GITHUB_REPO.split('/');
```
danielrozenberg: Validate your flags with this one neat trick that crashes the server on `require()` if the flags don't pass validation 😂

danielrozenberg: (which is to say... this is totally fine because we don't want to overengineer everything)

owners/src/file_cache.js
```diff
@@ -16,17 +16,41 @@
...
+class FileCache {
+  /**
+   * Fetch the contents of a file.
+   *
+   * @param {string} filename file to get contents of.
+   * @param {string} getContents function to get contents if file not in cache.
+   * @return {string} file contents.
+   */
+  async readFile(filename, getContents) {
+    return await getContents();
```
danielrozenberg: Where is `getContents()` defined in this context?

owners/scripts/warm_cache.js
rcebulko: Neat trick!

owners/src/file_cache.js
rcebulko: If I'm understanding your question correctly, `getContents` is provided by whatever client code is trying to read a file through the cache. The idea being, `getContents` is the "real" way to read the file (ie. filesystem IO or API call), but it may not be required if the file is in the cache.

owners/index.js
rcebulko: Octokit docs are terrible and inconsistent :unamused:

rcebulko:
Travis is borked at the moment, but all tests in `owners` pass as does Prettier, so I'm submitting even though Travis is red.

PR #565, opened 2019-10-23
[owners] Remove dependency on Probot when testing GitHub API

rcebulko:
When I originally wrote these tests, I wasn't familiar with Jest and I didn't know how to create a GitHub client instance without faking a Probot event and getting it from the context. Spinning up all those apps for each test made it pretty slow, especially as the class grew to contain many API methods (~6.5s to run all tests).

This PR swaps in an Octokit instance and removes the dependency on Probot, bringing the test runtime down to ~1.5s

### DIFF
--- a/owners/src/github.js
+++ b/owners/src/github.js
@@ -139,10 +139,10 @@ class GitHub {
    *     https://probot.github.io/api/latest/interfaces/githubapi.html).
    * @param {string} owner GitHub repository owner.
    * @param {string} repository GitHub repository name.
-   * @param {!Logger} logger logging interface.
+   * @param {Logger=} [logger=console] logging interface.
    */
   constructor(client, owner, repository, logger) {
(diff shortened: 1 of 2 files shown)


owners/src/github.js
```diff
@@ -139,10 +139,10 @@ class GitHub {
    *     https://probot.github.io/api/latest/interfaces/githubapi.html).
    * @param {string} owner GitHub repository owner.
    * @param {string} repository GitHub repository name.
-   * @param {!Logger} logger logging interface.
+   * @param {Logger=} [logger=console] logging interface.
```
rsimha: Same comment as the other PR. It's not common in `amphtml` to use JSDoc to specify default args. Should the function signature be as follows?
```
constructor(client, owner, repository, logger=console)
```

rcebulko: See #568

PR #566, opened 2019-10-23
[owners] Split repos into separate modules

rcebulko:
In an effort to organize some of the modules in the project, I'm moving some of them into submodules. This is being broken up into a few small PRs to ~prevent~ reduce rebase-hell


owners/src/repo/abstract_repo.js
```diff
@@ -0,0 +1,46 @@
...
+ * distributed under the License is distributed on an "AS-IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Interface for reading from a GitHub repository.
+ */
+module.exports = class AbstractRepository {
```
rsimha: Some suggestions:
- Rename the base class just `Repository`
- Add `@interface` to the JSDoc to indicate that this is an abstract class
- Implement a placeholder constructor that throws if someone tries to instantiate the base class

rcebulko: Re: first 2, done.
Re: your third point, that won't work out. ES6 constructors have to call their super constructor, so the constructor can't throw. Adding the interface JSDoc though.

rsimha: Might something like this work?
```js
constructor() {
  if (this.constructor === Repository) {
    throw new TypeError('Cannot instantiate base class'); 
  }
}
```

rcebulko: See #568

PR #570, opened 2019-10-24
[owners] Don't read .env file during tests

rcebulko:
There have been a few cases where tests passed locally because they read from a `.env` file, but since those env vars weren't stubbed the tests failed on Travis. This PR gates the `require('dotenv').config()` call so it doesn't initialize the environment when `NODE_ENV` is `test`

Also cleans up bootstrap function.

### DIFF
--- a/owners/bootstrap.js
+++ b/owners/bootstrap.js
@@ -62,9 +62,11 @@ function bootstrap(logger) {
         logger.error(err);
         process.exit(1);
       });
+
+    components = {ownersBot, github, initialized};
   }
 
-  return {ownersBot, github, initialized};
+  return components;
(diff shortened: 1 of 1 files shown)


owners/bootstrap.js
```diff
@@ -15,7 +15,7 @@
  */
 
 // Keep the components in scope so that, if multiple files include this, the
-// bootstrapping is only done once and all files share the components.
+// bootstrapping is only done once and all files share the
```
danielrozenberg: ?

rcebulko: Lol find and replace bug. Fixed.

PR #573, opened 2019-10-24
[owners] Add cron.yaml and app.yaml for GAE deployments

rcebulko:


owners/cron.yaml
```diff
@@ -0,0 +1,25 @@
...
+# limitations under the License.
+
+cron:
+# Note that both of these happen automatically in response to webhooks for new
+# or edited owners file and teams; this cron task ensures that the app stays in
+# sync even if webhooks are dropped or fail.
+
+- description: "Sync with the latest owners file and re-parse the owners tree"
+  url: /_cron/refreshTree
+  schedule: every hour from 7:00 to 23:00
```
danielrozenberg: We're technically an international project, might as well make these 24/7 :)

rcebulko: good call!

owners/app.yaml
```diff
@@ -1,6 +1,15 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
 
-# Temporary setting to keep gcloud from uploading node_modules
-skip_files:
- - ^node_modules$
```
rsimha: Curious: Was `yarn` being run on the old GCE instance?

rcebulko: No, `npm install` was used. GCE doesn't use `app.yaml`, so I think this was just a relic of a past attempt at a GAE migration

owners/cron.yaml
rcebulko: Done

PR #577, opened 2019-10-24
[bundle-size] Overwrite /report with /report.json endpoint

danielrozenberg:
amphtml repo is in sync, so we can remove the legacy code


bundle-size/app.js
```diff
@@ -524,57 +524,9 @@ module.exports = app => {
...
-          baseSha,
-          bundleSize,
-          /* lastAttempt */ retriesLeft == 0
-        );
-      } while (retriesLeft > 0 && !reportSuccess);
-    }
-  });
-
-  v0.post('/commit/:headSha/report.json', async (request, response) => {
+  // TODO(danielrozenberg): remove the .json optional endpoint once the amphtml
```
rcebulko: Is this TODO still relevant?

danielrozenberg: Yes, I want to remove the optional `.json` extension from this endpoint so eventually there will be only one copy of it

PR #580, opened 2019-10-28
[owners] Switch info server to template rendering and add OWNERS example file

rcebulko:
This PR:
- replaces info server endpoints that concatenated HTML with `lodash` template rendering
  > Note that `lodash` is already a dependency of Probot
- Uses the OWNERS.example file as a syntax-highlighted documentation page
- Fixes a bug introduced by prefixing path names with the repository name
- Splits apart the repo owner and name in environment variables
- Removes the old OWNERS.yaml example

### DIFF
--- a/owners/info_server.js
+++ b/owners/info_server.js
@@ -14,7 +14,12 @@
  * limitations under the License.
  */
 
+const fs = require('fs');
 const express = require('express');
+const _ = require('lodash');
+const hl = require('highlight').Highlight;
+
+const EXAMPLE_OWNERS_PATH = './OWNERS.example';
 
 /**
  * Generic server wrapping express routing.
@@ -85,6 +90,11 @@ class Server {
       });
     });
   }
+
+  render(view, ctx = {}) {
+    const template = fs.readFileSync(`./templates/${view}.template.html`);
+    return _.template(template)(ctx);
+  }
 }
 
 /**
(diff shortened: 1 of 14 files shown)


owners/info_server.js
```diff
@@ -14,7 +14,14 @@
  * limitations under the License.
  */
 
+const fs = require('fs');
 const express = require('express');
+const _ = require('lodash');
+const hl = require('highlight').Highlight;
```
danielrozenberg: Probot uses [handlebars](https://handlebarsjs.com/) templates built-in, so you can consider using those instead of highlight and reduce the number of dependencies used here (although if you do plan on switching to handlebars, you should still explicitly add it as a dependency)

owners/templates/example.template.html
```diff
@@ -0,0 +1,141 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <title>Owners File Syntax</title>
+
+  <style type="text/css">
+    /*
+      github.com style (c) Vasily Polovnyov <vast@whiteants.net>
```
danielrozenberg: What's the license for this file?

owners/info_server.js
rcebulko: I'm not sure I understand what you're saying here. Highlight provides syntax-highlighting of the OWNERS file (rendering into HTML with classes). Lodash is a part of Probot and it contains templates built-in, so this is doing the same thing re: using an existing dependency.

owners/templates/example.template.html
rcebulko: It's taken directly from `lodash` example files, so it's the same as that repo aka MIT

danielrozenberg: I think that means we can't use it then? I'm not sure how MIT and Apache interact

rcebulko: NodeJS is MIT...

rcebulko: And MIT license is *more* permissive than Apache

danielrozenberg: Dependencies ≠ inclusion in the codebase...

rcebulko: @rsimha are you aware of what the recommendation is for this?

danielrozenberg: Kinda want to resurrect this 🙃

PR #586, opened 2019-10-29
[owners] Final fixes to GAE deployment

### LINKED ISSUE #540: [owners] Migrate from GCE to GAE

rcebulko:
Closes #540


owners/package.json
```diff
@@ -24,14 +24,14 @@
     "lodash": "4.17.15",
     "minimatch": "3.0.4",
     "probot": "7.5.3",
-    "sleep-promise": "8.0.1"
+    "sleep-promise": "8.0.1",
+    "smee-client": "1.1.0"
```
danielrozenberg: This should only be used as a dev dependency. GAE exposes externally accessible URLs for the webhooks and those should be used by the github app

rcebulko: Oh, my understanding was the opposite. I'll revert that then

rcebulko: Do you know what the default webhook route would be?

PR #590, opened 2019-10-29
[owners] Remove (most) references to real usernames in tests

### LINKED ISSUE #567: Replace all GitHub identifiers in tests and fixtures with placeholders
/cc @rsimha

rcebulko:
Addresses #567


estherkim:
Can you remove `id`, `node_id` and `avatar_url` as well?

rcebulko:
> Can you remove `id`, `node_id` and `avatar_url` as well?

That's gonna be a decent bit more work, since it's gonna involve manually finding all those fields and updating them one by one (unlike usernames per the issue, since that could be done via find-and-replace). I can see the benefit of not including real usernames in tests themselves, but residual references in unused properties of fixture files seems relatively harmless (especially since it's all public data, in the repo, not PII, etc. For the time being I'd prefer to submit this cleanup, possibly file a tracking issue for identifying and removing all possible references to any users, and then discuss with the rest of @ampproject/wg-infra on that issue. WDYT?

estherkim:
I would argue that `id`, `node_id`, and `avatar_url` aren't "residual references" but that they are as real as `username`. For example, by keeping `avatar_url`, you now have real people's avatars in a single JSON file. Yes they're public, but they're not easily harvested like this. Just as you are removing some real data, you should remove ALL real data.

It's definitely unfortunate that removing it is not as easy as a find&replace, but we should not be using any real data as test data.

danielrozenberg:
All my tests contain real usernames of people, and that's all publicly accessible data for anyone who can generate a GitHub token (aka, anyone) - why does it matter that much? :)

rcebulko:
> I would argue that `id`, `node_id`, and `avatar_url` aren't "residual references" but that they are as real as `username`. For example, by keeping `avatar_url`, you now have real people's avatars in a single JSON file. Yes they're public, but they're not easily harvested like this. Just as you are removing some real data, you should remove ALL real data.
> 
> It's definitely unfortunate that removing it is not as easy as a find&replace, but we should not be using any real data as test data.

They are extremely easily "harvested" as they are provided by any number of unauthenticated APIs ([like this one](https://api.github.com/orgs/ampproject/public_members)). My understanding is that the purpose of #567 was more generally best practices in testing re: using fake names and identifiers, not necessarily sanitation for any particular privacy concerns. The time-cost/benefit of removing every URL and ID seems a lot lower than that of clearing out usernames from test code.

estherkim:
> All my tests contain real usernames of people, and that's all publicly accessible data for anyone who can generate a GitHub token (aka, anyone) - why does it matter that much? :)

By uploading JSON data, you're circumventing the authentication step that GitHub has in place. It may be minimal but it's there for a reason.

> unauthenticated APIs ([like this one](https://api.github.com/orgs/ampproject/public_members)).

Not everyone chooses to make their org membership public, yet their info is found in tests. If you got their info from an authenticated API call, it should not be uploaded as JSON.

> My understanding is that the purpose of #567 was more generally best practices in testing re: using fake names and identifiers

Totally agree with the intent of the issue, and sorry if I wasn't clear in my first comment. `id`, `node_id`, and `avatar_url` count as identifiers. As a best practice (not even going down the privacy route), it is real data and should be scrubbed.

rcebulko:
Agree with both of you. As I said before, I'd prefer to file a separate tracking issue for scrubbing all fixture data and get the actual code changes in, and determine how to prioritize scrubbing identifiers like that in the appropriate thread. I don't think additional identifiers being present needs to block this changeset which removes quite a few, and which will progressively bit-rot with time.

rsimha:
I agree that we should look to remove all real data from tests, whether they are user IDs, avatars, profiles URLs, etc. Getting there with a follow up PR is fine if it doesn't break tests in the interim. (Looks like that's true here.)

One final comment: Before merging this, I'd encourage you to remove all @mentions of usernames from commit messages in this PR. Otherwise, the people mentioned will get notified every time this PR is rebased, and every time someone references the commits in their forks once the PR is merged.

rcebulko:
> I agree that we should look to remove all real data from tests, whether they are user IDs, avatars, profiles URLs, etc. Getting there with a follow up PR is fine if it doesn't break tests in the interim. (Looks like that's true here.)
> 
> One final comment: Before merging this, I'd encourage you to remove all @mentions of usernames from commit messages in this PR. Otherwise, the people mentioned will get notified every time this PR is rebased, and every time someone references the commits in their forks once the PR is merged.

Wow, I didn't realize GitHub still tagged things that were part of commit messages! Will remove

rcebulko:
Updated title of #567 and removed all @ mentions from commit messages @estherkim @rsimha

PR #591, opened 2019-10-29
Fix `set-value` dependency vulnerability

### LINKED ISSUE #270: Several security vulnerabilities seen in `amp-github-apps` repo
We should ~enable renovate for~ upgrade the packages in this repo to fix these and future security vulnerabilities.

image

image

rcebulko:
Fix performed via `npm audit fix`
Closes #270

### DIFF
--- a/test-status/package-lock.json
+++ b/test-status/package-lock.json
@@ -4094,9 +4094,9 @@
       },
       "dependencies": {
         "handlebars": {
-          "version": "4.1.2",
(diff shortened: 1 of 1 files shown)


rsimha:
For posterity, it'd be good to document which of our dependencies brought in the offending package, and whether there was a renovate PR that would have fixed the vulnerability.

rcebulko:
It's a pretty deep dependency chain...
set-value > union-value > cache-base > base > snapdragon > braces > chokidar > nodemon
there may be other chains, but there was no pending renovate PR for any of these that I saw.

PR #597, opened 2019-10-31
[owners] Fix GitHub Search and App Engine issues

rcebulko:
The GitHub Search API is the only available way to find OWNERS files in the repo through the GitHub API. It is, however, non-deterministic, and often responds missing a few results. This PR:
- separates syncing the virtual repo's OWNERS file list from returning a list of owners files, ensuring files can't "go missing"
- switches the search pagination to use Octokit's built-in pagination
- syncs the virtual repository's OWNERS file list during initialization
- disables App Engine automatic scaling

### DIFF
--- a/owners/app.yaml
+++ b/owners/app.yaml
@@ -13,3 +13,7 @@
 
 runtime: nodejs
 env: flex
+
+automatic_scaling:
+  min_num_instances: 1
+  max_num_instances: 1
--- a/owners/src/api/github.js
+++ b/owners/src/api/github.js
@@ -445,31 +445,18 @@ class GitHub {
   async searchFilename(filename) {
     this.logger.info(`Searching repo for files named "${filename}"`);
 
-    const files = [];
-    let page = 1;
-    let isLastPage = false;
-    while (!isLastPage) {
-      this.logger.info(`Fetching page ${page}`);
-      const response = await this.client.search.code({
-        q: `filename:${filename} repo:${this.owner}/${this.repository}`,
-        per_page: MAX_PER_PAGE,
-        page,
-      });
+    const files = await this._paginate(
+      this.client.search.code,
+      {q: `filename:${filename} repo:${this.owner}/${this.repository}`},
+    );
 
-      const {items} = response.data;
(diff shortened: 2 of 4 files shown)


owners/app.yaml
```diff
@@ -13,3 +13,6 @@
 
 runtime: nodejs
 env: flex
+
+automatic_scaling:
+  max_num_instances: 1
```
danielrozenberg: Also `min_num_instances: 1`? 🙃

owners/src/repo/virtual_repo.js
```diff
@@ -58,6 +77,14 @@ module.exports = class VirtualRepository extends Repository {
    * @param {?function} cacheMissCallback called when there is a cache miss.
    */
   async warmCache(cacheMissCallback) {
+    // The GitHub Search API is non-deterministic, and will frequently return
+    // duplicates or leave out files. It syncs periodically so if a given search
+    // drops one or two files, it usually won't cause any problems. When
+    // starting up the virtual repo and warming the cache, however, we want to
+    // make sure we know all the files we need to know about, so we try a couple
+    // times.
+    await Promise.all([this.sync(), this.sync(), this.sync()]);
```
danielrozenberg: I... I don't know how to feel about this 😐

rcebulko: I don't like it at all. It's what's currently there, I don't really want to check it in but I put it here cause it's what currently did the job. Very open to ideas

owners/app.yaml
rcebulko: I don't think that's needed based on examples I saw in docs, but added anyway to be sure :+1:

owners/src/repo/virtual_repo.js
rcebulko: For now, removed this hack and restored the sync-on-read approach. It will do the job for now, until I have the chance to implement more resilient caching/initialization

rcebulko:
@danielrozenberg Pulled all the virtual repo changes out to simplify this PR, will address that issue in a follow-up

PR #602, opened 2019-11-01
[owners] Fetch and save the set of known file refs through the cache

rcebulko:
The GitHub Search API is the only available API method by which the virtual repo can determine the list of OWNERS files. Since this API is non-deterministic and often leaves out results, it's useful to keep a copy of the file ref list available, rather than fetching it from the API each time the tree is parsed. This PR updates the virtual repo to fetch the list of file refs through the cache, and updates the cache whenever it sees that a file ref sha has changed.

This means that even if the first request during warm-up doesn't catch all the owners files, each subsequent attempt will improve coverage, and the combination will be available going forward (even on system start-up, as it's cached in Cloud Storage).

### DIFF
--- a/owners/src/repo/virtual_repo.js
+++ b/owners/src/repo/virtual_repo.js
@@ -69,9 +69,8 @@ module.exports = class VirtualRepository extends Repository {
     if (fileChanged) {
       const fileRefsPath = this.repoPath('__fileRefs__');
       await this.cache.invalidate(fileRefsPath);
-      await this.cache.readFile(
-        fileRefsPath,
-        () => JSON.stringify(Array.from(this._fileRefs))
+      await this.cache.readFile(fileRefsPath, () =>
+        JSON.stringify(Array.from(this._fileRefs))
       );
     }
   }
--- a/owners/src/repo/virtual_repo.js
+++ b/owners/src/repo/virtual_repo.js
@@ -30,7 +30,7 @@ module.exports = class VirtualRepository extends Repository {
     super();
     this.github = github;
     this.logger = github.logger;
-    /** @type {!Map<string, string>} */
+    /** @type {?Map<string, string>} */
     this._fileRefs = new Map();
     this.cache = cache;
   }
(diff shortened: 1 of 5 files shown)


owners/src/repo/virtual_repo.js
```diff
@@ -30,7 +30,7 @@ module.exports = class VirtualRepository extends Repository {
     super();
     this.github = github;
     this.logger = github.logger;
-    /** @type {!Map<string, string>} */
+    /** @type {?Map<string, string>} */
```
danielrozenberg: Doesn't look optional to me ¯\_(ツ)_/¯

owners/src/repo/virtual_repo.js
```diff
@@ -39,7 +39,40 @@ module.exports = class VirtualRepository extends Repository {
    * Fetch the latest versions of all OWNERS files.
    */
   async sync() {
-    await this.findOwnersFiles();
+    let fileChanged = false;
+    const ownersFiles = await this.github.searchFilename('OWNERS');
```
danielrozenberg: Here's an insane idea: when syncing/warming up, download the entire repository as a tar file (https://api.github.com/repos/ampproject/amphtml/tarball) in memory and read the files from there 🙃
(I checked to see if there's a way to list all the files in a repo using the GitHub API but there really doesn't seem to be) // cc @estherkim @rsimha for sanity check on this insane idea

owners/src/repo/virtual_repo.js
rcebulko: I swear I've reverted this change from `?` to `!` like four times idk how it keeps popping back into place :rage:

owners/src/repo/virtual_repo.js
rcebulko: That is quite the impressive combination of insane and fully plausible

rsimha: The tar file for the entire repo is ~130 MB in size, so it's not unreasonable to download it during warm-up if you think it's faster than an API call. (You'd have to work around GAE file system restrictions, right?)

danielrozenberg: Doing the tar expansion in-memory and discarding it, once for warm-up, is _probably_ possible in GAE

PR #606, opened 2019-11-05
Set up build targets for amp-github-apps

### LINKED ISSUE #605: Set up build targets/conditional test running for this repo
As the number of apps in this repo grows, Travis testing is taking longer and longer because every test for every app is run each time. Instead, we should have a system like `ampproject/amphtml`'s  build targets, to only run the test command for apps with files changed in the PR.

The plan is to pretty much replicate the existing `build-targets.js` and `checks.js`, albeit with much simpler rules for now--only looking at whether some file in the app directory is changed, without implemented the more detailed unit/integration path detection and other tests used in `amphtml`. Edits can be made as needed later on.

/cc @ampproject/wg-infra

rcebulko:
- Transfers a lot of the helper functions for Git, Travis, and  `exec` directly from `ampproject/amphtml`
- Mimics the core logic/structure of build targets from `ampproject/amphtml`
- Defines build targets for each app

Closes #605

### DIFF
--- a/build-system/build-targets.js
+++ b/build-system/build-targets.js
@@ -31,10 +31,6 @@ const ALL_TARGETS = ['bundle-size', 'owners', 'pr-deploy', 'test-status'];
  */
 const targetMatchers = [
   {
-    targets: ALL_TARGETS,
-    func: file => file.startsWith('build-system') || path.dirname(file) === '.',
-  },
-  {
     targets: ['bundle-size'],
     func: file => file.startsWith('bundle-size/'),
   },
@@ -68,6 +64,11 @@ function determineBuildTargets(fileName = 'build-targets.js') {
     .map(matcher => matcher.targets)
     .reduce((left, right) => left.concat(right), [])
     .forEach(buildTargets.add, buildTargets);
+  
+  if (buildTargets.size === 0) {
+    // Default to running all test targets if no other targets are matched.
+    ALL_TARGETS.forEach(buildTargets.add, buildTargets)
+  }
 
   const targetList = Array.from(buildTargets)
     .sort()
--- a/build-system/build-targets.js
+++ b/build-system/build-targets.js
@@ -23,29 +23,24 @@
 const {bold, cyan, yellow} = require('ansi-colors');
 const {gitDiffNameOnlyMaster} = require('./git');
 
-const ALL_TARGETS = ['bundle-size', 'owners', 'pr-deploy', 'test-status'];
+const APP_TARGETS = {
+  'BUNDLE_SIZE': 'bundle-size',
+  'OWNERS': 'owners',
+  'PR_DEPLOY': 'pr-deploy',
+  'TEST_STATUS': 'test-status',
+};
+const ALL_TARGETS = [...Object.keys(APP_TARGETS)];
 
 /**
  * A mapping of functions that match a given file to one or more build targets.
  */
-const targetMatchers = [
-  {
-    targets: ['bundle-size'],
-    func: file => file.startsWith('bundle-size/'),
-  },
-  {
-    targets: ['owners'],
-    func: file => file.startsWith('owners/'),
-  },
-  {
-    targets: ['pr-deploy'],
-    func: file => file.startsWith('pr-deploy/'),
-  },
-  {
-    targets: ['test-status'],
-    func: file => file.startsWith('test-status/'),
-  },
-];
+const appTargetMatchers = Object.entries(APP_TARGETS).map((target, appName) => {
+  return {
+    targets: [target],
+    func: file => file.startsWith(appName),
+  };
+});
+const targetMatchers = [...appTargetMatchers];
 
 /**
  * Populates buildTargets with a set of build targets contained in a PR after
@@ -82,5 +77,7 @@ function determineBuildTargets(fileName = 'build-targets.js') {
 }
 
 module.exports = {
+  ALL_TARGETS,
+  APP_TARGETS,
   determineBuildTargets,
 };
--- a/build-system/build-targets.js
+++ b/build-system/build-targets.js
@@ -21,7 +21,6 @@
  * determine which tasks are required to run for pull request builds.
  */
 const {bold, cyan, yellow} = require('ansi-colors');
-const path = require('path');
 const {gitDiffNameOnlyMaster} = require('./git');
 
 const ALL_TARGETS = ['bundle-size', 'owners', 'pr-deploy', 'test-status'];
@@ -64,10 +63,10 @@ function determineBuildTargets(fileName = 'build-targets.js') {
     .map(matcher => matcher.targets)
     .reduce((left, right) => left.concat(right), [])
     .forEach(buildTargets.add, buildTargets);
-  
+
   if (buildTargets.size === 0) {
     // Default to running all test targets if no other targets are matched.
-    ALL_TARGETS.forEach(buildTargets.add, buildTargets)
+    ALL_TARGETS.forEach(buildTargets.add, buildTargets);
   }
 
   const targetList = Array.from(buildTargets)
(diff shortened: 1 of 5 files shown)


build-system/build-targets.js
```diff
@@ -0,0 +1,104 @@
...
+/**
+ * @fileoverview
+ * This script sets the build targets for our PR check, where the build targets
+ * determine which tasks are required to run for pull request builds.
+ */
+const {bold, cyan, yellow} = require('ansi-colors');
+const path = require('path');
+const {gitDiffNameOnlyMaster} = require('./git');
+
+const ALL_TARGETS = ['BUNDLE_SIZE', 'OWNERS', 'PR_DEPLOY', 'TEST_STATUS'];
```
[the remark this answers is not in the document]
rcebulko: Yeah I guess to some degree, I'll fiddle with it

build-system/build-targets.js
```diff
@@ -0,0 +1,104 @@
...
+/**
+ * Determines if a file is a filetype containing code.
+ *
+ * @param {string} filePath
+ * @return {boolean}
+ */
+function isCode(filePath) {
+  const fileName = path.basename(filePath);
+  return !(
+    fileName.startsWith('.') ||
```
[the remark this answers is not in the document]
estherkim: That may introduce additional work down the road because in the case that we change or add any dot file like `.env` that affects tests, someone would have to update this.

We should lean towards running tests as opposed to skipping them.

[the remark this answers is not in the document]
rcebulko: Fair enough, I'll drop this bit and just run it for any changes to the root directory

build-system/git.js
```diff
@@ -0,0 +1,77 @@
+/**
+ * Copyright 2018 The AMP HTML Authors. All Rights Reserved.
```
rsimha: nit: 2019.

build-system/build-targets.js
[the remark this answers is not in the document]
rsimha: > We should lean towards running tests as opposed to skipping them.

This is the kicker.

+1 to getting rid of `isCode` and defaulting to running all tests for changes that don't obviously belong to one particular app (like dotfiles in the root dir).

There's one worthwhile optimization though: Don't run tests against PRs that only update documentation. (`amphtml` runs `gulp check-links`. May be worth adding here in a separate PR. There's also `gulp lint` and `gulp prettify` that check code / documentation style against local changes no matter what the PR does.)

build-system/build-targets.js
```diff
@@ -0,0 +1,104 @@
...
+ * A mapping of functions that match a given file to one or more build targets.
+ */
+const targetMatchers = [
+  {
+    targets: ALL_TARGETS,
+    func: file =>
+      (isCode(file) && file.startsWith('build-system') ||
+      file === 'package.json' ||
+      file === 'package-lock.json'),
+  },
```
rsimha: Remove this function. Instead, make `determineBuildTargets()` default to `ALL_TARGETS` for any file that doesn't match one of the other targets.

build-system/travis.js
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Returns true if this is a Travis Push build.
+ * @return {boolean}
+ */
+function isTravisPushBuild() {
+  return isTravisBuild() && process.env.TRAVIS_EVENT_TYPE === 'push';
+}
+
+module.exports = {isTravisBuild, isTravisPullRequestBuild, isTravisPushBuild};
```
rsimha: Style nit: Split this list across multiple lines so that adding a new one later will generate a one-line diff.

build-system/git.js
```diff
@@ -0,0 +1,77 @@
...
+ * @return {!Array<string>}
+ */
+function gitDiffNameOnlyMaster() {
+  const masterBaseline = gitMasterBaseline();
+  return getStdout(`git diff --name-only ${masterBaseline}`)
+    .trim()
+    .split('\n');
+}
+
+module.exports = {gitDiffNameOnlyMaster};
```
rsimha: Same style nit: Put each export on a new line, even if there's just one.

build-system/pr-check.js
```diff
@@ -29,18 +33,51 @@ function timedExecOrDie(cmd) {
   log.info('Done running', colors.cyan(cmd), '...');
 }
 
+/**
+ * Set up and execute tests for an app.
+ *
+ * TODO(rcebulko): Replace this with gulp tasks.
```
rsimha: Agree with this TODO. While you're at it, we should also apply the same logging standards that `amphtml` uses.

build-system/build-targets.js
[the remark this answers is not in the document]
rcebulko: Just seeing this, but already removed `isCode` per recs from @danielrozenberg and @estherkim . We don't have gulp set up here, but `pr-check` runs `eslint .` before any build target checking. As for excluding documentation, I think I'll put that in as a separate PR.

build-system/build-targets.js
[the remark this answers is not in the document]
rcebulko: Done.

build-system/build-targets.js
rcebulko: Done.

build-system/pr-check.js
rcebulko: Adding a TODO for logging

build-system/git.js
rcebulko: Done

build-system/travis.js
rcebulko: Done

build-system/git.js
rcebulko: Done.

rcebulko:
@rsimha @danielrozenberg I tried out something in the middle of the suggestions from each of you, reverting to separate build target names but adding a map from build targets to app names and generating some of the build targets based on that. Curious what your thoughts are re: does this just add more complexity or simplify some things.

build-system/build-targets.js
```diff
@@ -23,29 +23,24 @@
...
 const {gitDiffNameOnlyMaster} = require('./git');
 
-const ALL_TARGETS = ['bundle-size', 'owners', 'pr-deploy', 'test-status'];
+const APP_TARGETS = {
+  'BUNDLE_SIZE': 'bundle-size',
+  'OWNERS': 'owners',
+  'PR_DEPLOY': 'pr-deploy',
+  'TEST_STATUS': 'test-status',
+};
+const ALL_TARGETS = [...Object.keys(APP_TARGETS)];
```
rcebulko: This is just app targets right now, but leaves plenty of room for other targets (ex. `DOCS`) to be added separately

build-system/pr-check.js
```diff
@@ -55,18 +59,21 @@ function runAppTests(appName) {
...
+    buildTargets = determineBuildTargets(FILENAME);
     log.info(`Detected build targets: ${Array.from(buildTargets).join(', ')}`);
   }
 
+  for (const [target, appName] of Object.entries(APP_TARGETS)) {
+    if (buildTargets.has(target)) {
+      runAppTests(appName);
+    }
+  }
+
```
rcebulko: This is where targets besides the app targets could define what should be run.

estherkim:
> @rsimha @danielrozenberg I tried out something in the middle of the suggestions from each of you, reverting to separate build target names but adding a map from build targets to app names and generating some of the build targets based on that. Curious what your thoughts are re: does this just add more complexity or simplify some things. Take a peek at [this commit](https://github.com/ampproject/amp-github-apps/pull/606/commits/3fed09bdf46a75401dd51207174d4c66ec2752fa) and lmk WDYT?
> 
> /cc @estherkim any opinions?

Hmmm I liked the original way (all caps despite the dupes) because it's the easiest to read. The map seems too complex and I had to do a double take because `ALL_TARGETS` and `APP_TARGETS` look so similar at first.

rcebulko:
Reverted back to the original implementation. @estherkim @danielrozenberg

PR #609, opened 2019-11-05
Remove GitHub identifiers from most test fixtures

rcebulko:
This doesn't take care of all of them, but hits a few of the major offending directories

Addresses #567


rcebulko:
@estherkim There are a few more, I'm trying to get this done piecemeal when I have bursts of downtime between other projects

rcebulko:
(The remaining ones are reviews and teams, which are very user ID-heavy, so I'll need to address those in a separate chunk of time)

PR #611, opened 2019-11-06
[bundle-size-chart] Initial version

danielrozenberg:
Remember that this is mostly a hack and not a serious piece of code :)

// cc: @kevinkimball


bundle-size-chart/app.yaml
```diff
@@ -0,0 +1,21 @@
+# Copyright 2019, the AMP HTML authors
```
estherkim: Should this file be uploaded to GitHub?

bundle-size-chart/src/app.ts
```diff
@@ -0,0 +1,142 @@
...
+const CRON_TIMEOUT_MS = 600000; // 10 minutes, the GAE cron timeout limit.
+const MAX_COMMIT_DAYS = 365;
+const GCLOUD_STORAGE_BUCKET = 'amp-bundle-size-chart';
+
+dotenv.config();
+
+const app = express();
+app.use(express.static('static'));
+
+app.get('/_cron', async (request, response) => {
```
estherkim: Would like to see methods split out here with corresponding unit tests. Not needed right away.

danielrozenberg: I very much doubt I'll ever write tests for this 😅

bundle-size-chart/app.yaml
danielrozenberg: Yeah, there's nothing secret about this file. I dunno why I excluded it from the other apps...

PR #621, opened 2019-11-20
[bundle-size] [test-status] Refactor test to remove the `waitUntilNocksScopeIsDone` method

danielrozenberg:


bundle-size/test/api.test.js
```diff
@@ -427,36 +427,42 @@ describe('bundle-size api', () => {
...
+            })
+            .reply(200, () => {
+              setTimeout(() => {
+                resolve(nocks);
+              }, 0);
             });
-            return true;
-          })
-          .reply(200);
+        }).then(nocks => nocks.done());
```
jridgewell: Wouldn't this be clearer with `nocks.done()` as the last statement in the test?

danielrozenberg: after `await lastNetworkRequest;`? Yeah, I can do that

danielrozenberg: Oh wait, since nocks is returned with the promise resolve, I have to do it *with* that await

danielrozenberg: Done, PTAL

I'll fix the broken test tomorrow

bundle-size/test/api.test.js
```diff
@@ -454,15 +454,15 @@ describe('bundle-size api', () => {
...
+        });
 
         await request(probot.server)
           .post('/v0/commit/26ddec3fbbd3c7bd94e05a701c8b8c3ea8826faa/report')
           .send(jsonPayload)
           .set('Content-Type', 'application/json')
           .set('Accept', 'application/json')
           .expect(202);
-        await lastNetworkRequest;
+        await lastNetworkRequest.then(nocks => nocks.done());
```
jridgewell: I don't understand why this needs to be in a then callback. Is `nocks.done()` async? If so, we should always be awaiting its return value.

rcebulko: It is not; `nocks.done()` will cause the test to fail if all nocked requests have not been received and responded to.

jridgewell: Then there is no difference between `await lastNetworkRequest.then(() => nocks.done())` and `await lastNetworkRequest; nocks.done()`. The latter is cleaner.

rcebulko: Agreed there. The diff I'm seeing didn't include `nocks.done()` after the await

danielrozenberg: Only issue is that `nocks` is defined inside the `new Promise` scope, so I can't refer to is. That's why I'm returning it with `resolve(nocks)` and using `.then` on it.

Another option I can do is this:
```javascript
let nocks;
const lastNetworkRequest = new Promise(resolve => {
  nocks = ...
});

...

await lastNetworkRequest;
nocks.done();
```

Is this preferable?

🚲🏚

jridgewell: Ahh, I didn't see that you were doing `resolve(nocks)`, thought this was all in scope.

```js
const nocks = await lastNetworkRequest;
nocks.resolve();
```

But why do you need a Promise to begin with?

rcebulko: The `nocks.done()` call must happen *after* the reply callback, not within.

jridgewell: That's how the code I wrote behaves.

danielrozenberg: I switched to the syntax that you proposed in https://github.com/ampproject/amp-github-apps/pull/621#discussion_r349354245. The reason we need this delayed resolution using a Promise is because the `request` returns immediately (with a `202 CREATED`) but the server continue to perform background work. If we didn't have `await lastNetworkRequest`, the `nocks.done()` assertion would be executed before, well, the last network request was satisfied...

PR #627, opened 2019-11-25
[owners] Disable reviewer assignment on "WIP"/draft/"DO NOT SUBMIT" PRs

### LINKED ISSUE #587: [owners] Disable reviewer assignment on draft, WIP, and DNS PRs
Any PR created as a draft, or with "WIP" or "DO NOT SUBMIT" in the title should not have reviewers automatically added.

rcebulko:
Also cleaned up some `nock` calls with chaining. I recommend reviewing commit-by-commit, as viewing all changes at once may make it harder than it needs to be.

Closes #587


owners/test/index.test.js
```diff
@@ -57,7 +58,18 @@ const ownersRules = [
...
 ];
 
-describe('owners bot', () => {
+/**
+ * Require a fresh copy of a module, clearing the require cache.
+ *
+ * @param {string} path module path.
+ * @return {*} required module.
+ */
+function _rerequire(path) {
```
danielrozenberg: Look at how I implemented `getFixture` in my projects.

I considered doing this `delete.require`/`require` thing too, but realized that this abuses the `require` function, which is supposed to be for code, not for fixtures. You should change all the fixture-retrieving code to load it with the `fs` module and not with `require`

https://github.com/ampproject/amp-github-apps/blob/master/bundle-size/test/_test_helper.js#L28-L32

owners/index.js
```diff
@@ -20,6 +20,8 @@ const InfoServer = require('./info_server');
 const {GitHub, PullRequest, Team} = require('./src/api/github');
 const {OwnersCheck} = require('./src/owners_check');
 
+const NO_REQUEST_OWNERS_REGEX = /\b(WIP|work in progress|DO NOT SUBMIT)\b/i;
```
danielrozenberg: Also `DO NOT MERGE`

owners/index.js
```diff
@@ -46,11 +48,28 @@ module.exports = app => {
...
   }
 
+  /**
+   * Identify if a PR is in a state that reviewers should be assigned.
+   *
+   * @param {!object} pr GitHub PR payload.
+   * @return {boolean} false if the PR is a draft or the title contains WIP/DNS.
+   */
+  function shouldAssignReviewers(pr) {
+    return !(pr.draft || NO_REQUEST_OWNERS_REGEX.test(pr.title));
```
danielrozenberg: So we're not looking at the description? A lot of people use the description to add WIP

rcebulko: There's a much higher chance of a PR description containing something like "ensure we do not submit broken code in case of ..." or other accidental collisions. This isn't supposed to catch everything, just the cases where it's obvious enough that it annoys people.

owners/index.js
rcebulko: Done

owners/test/index.test.js
rcebulko: Done.

PR #632, opened 2019-11-26
[bundle-size] Split GitHub util functions into a separate file

danielrozenberg:


danielrozenberg:
You are correct, this is just moves and renames

Issue #617, opened 2019-11-14
[bundle-size] Blocking check for bundle size changes for files other than `v0.js`


danielrozenberg:
Continuing the work that started in #271. @ampproject/wg-infra discussed this plan offline and we decided on the following:

We'll add an `APPROVERS.json` file to bundle-size directory of http://github.com/ampproject/amphtml that specifies which team(s) members can approved which *compiled* files (e.g., `dist/v0.js`, `dist/v0/amp-story-1.0.js`) and what's the allowed growth threshold (`0.1` == 0.1kb) before requiring approval. Unspecified files will not require approvals at all.

```json5
{
  "dist/v0.js": {
    "approvers": ["ampproject/wg-performance", "ampproject/wg-runtime"],
    "threshold": 0.1
  },
  "dist/amp4ads-v0.js": {
    "approvers": ["ampproject/wg-ads"],
    "threshold": 0.1
  },
  "dist/v0/amp-story-0.1.js": {
    "approvers": ["ampproject/wg-stories"],
    "threshold": 0.5
  },
  "dist/v0/amp-story-1.0.js": {
    "approvers": ["ampproject/wg-stories"],
    "threshold": 0.5
  },
  // …
}
```

To simplify this model, only teams will be allowed in the `approvers` list, not individual. Also, PRs that modify files that require approval from more than one *set* of teams will fall back to require approval from @ampproject/wg-runtime or @ampproject/wg-performance. Members of @ampproject/wg-infra will still be able to approve any bundle-size change.

rcebulko:
Could the `dist` prefix be excluded, if it only applies to files in `dist`?

danielrozenberg:
Yeah but I already pass `dist/` explicitly everywhere in this app so might as well. Space is cheap 🙃

PR #634, opened 2019-11-27
[bundle-size] Select a reviewer based on APPROVERS.json

danielrozenberg:
#617

Doesn't yet assign this reviewer, but the selections will appear in logs and will let me verify the behaviour in real scenarios

Tag-along changes:
* Shorten the error message for base bundle-size not found, since it's a common occurance
* Update JSDoc types with more accurate types
* Remove incorrect comment
* Add an optional `GitHubUtils` argument to `installApiRouter` to facilitate cache flushing in tests


bundle-size/github-utils.js
```diff
@@ -137,6 +141,80 @@ class GitHubUtils {
...
+  /**
+   * Choose a random reviewer from list of potential approver teams.
+   *
+   * @param {!Array<string>} potentialApproverTeams potential teams that can
+   *   approve this pull request.
+   * @param {number} pullRequestId the pull request id.
+   * @return {string} the chosen reviewer username.
+   */
+  async getRandomReviewer(potentialApproverTeams, pullRequestId) {
+    // First, randomly choose a team.
```
rcebulko: I don't think this two-tiered approach is desirable. Imagine a case where we have `wg-viewers` (2 members) and `wg-caching` (12 members). Members of `wg-viewers` will get way more reviews than members of `wg-caching`. Since team member lists are cached, it shouldn't be particularly expensive to just get the full set of all members of all 2-3 teams and choose randomly from there

bundle-size/test/fixtures/teams.listMembers.3065818.json
```diff
@@ -0,0 +1,42 @@
+[
+  {
+    "login": "jridgewell",
```
rcebulko: Preemptively remove real usernames and IDs pls

bundle-size/github-utils.js
danielrozenberg: Makes sense. PTAL at latest commit

danielrozenberg: Discussed offline. Merging.

PR #645, opened 2019-12-04
Show all JS files in default view of bundle-size chart

rsimha:
This PR removes the filter so that all JS files are shown in the default view, and instead, adds a suggestion for how to use the filter field.

### DIFF
--- a/bundle-size-chart/static/index.html
+++ b/bundle-size-chart/static/index.html
@@ -30,7 +30,7 @@
 <body>
   <h1><a href="https://github.com/ampproject/amphtml">AMPHTML</a> bundle-size history</h1>
   <div id="chart"></div>
-  <label for="filter">Filter</label> <input type="text" id="filter" value="v0.js">
+  <label for="filter">Filter (for example, "v0.js")</label> <input type="text" id="filter" value="">
   <script src="https://d3js.org/d3.v5.min.js" charset="utf-8"></script>
   <script src="/index.js"></script>
 </body>


bundle-size-chart/static/index.html
```diff
@@ -30,7 +30,7 @@
 <body>
   <h1><a href="https://github.com/ampproject/amphtml">AMPHTML</a> bundle-size history</h1>
   <div id="chart"></div>
-  <label for="filter">Filter</label> <input type="text" id="filter" value="v0.js">
+  <label for="filter">Filter (for example, "v0.js")</label> <input type="text" id="filter" value="">
```
danielrozenberg: This is gonna make the chart unreadable in the default view...

rsimha: I don't entirely agree. Right now, most information in the chart is hidden.

With this change: For interactive sessions (opening it in a browser), hovering over the chart will highlight different lines, and for non-interactive sessions (showy), you will still see more useful information, particularly inflections in various lines.

WDYT?

bhL3UqsWDLH

rsimha: Bump.

rsimha: @danielrozenberg ^

danielrozenberg: Maybe it makes sense to link this with https://github.com/ampproject/amphtml/blob/master/build-system/tasks/bundle-size/APPROVERS.json and only display those files that are there, to encourage WGs to add their files in there ;)

rsimha: I like your idea much more than what I was trying to do in this PR. Will abandon this.

PR #646, opened 2019-12-04
[bundle-size] Move new reviewer selection logic into the `addBundleSizeReviewer` and move that method into `GitHubUtils`

danielrozenberg:
#617

This will make it easier to switch to the new selection algorithm once the logs show that we're getting it right on the "dry-runs"

Tag-along changes:
* Update .env.example with a more accurate set of super-users
* Remove extended timeout for tests since they now pass fairly quickly

### DIFF
--- a/bundle-size/github-utils.js
+++ b/bundle-size/github-utils.js
@@ -191,11 +191,13 @@ class GitHubUtils {
   /**
    * Get a list of all unique team members by team ids.
    *
-   * @param {!Array<number>} teamIds the team ids.
+   * @param {!Array<string>} teamNames names of full team slug namesto get
+   *   members of.
    * @return {!Array<string>} all members of the provided teams.
    */
-  async getTeamMembers_(teamIds) {
-    const allTeamMembersPromises = teamIds.map(async teamId => {
+  async getTeamMembers_(teamNames) {
+    const allTeamMembersPromises = teamNames.map(async teamName => {
+      const teamId = await this.getTeamId_(teamName);
       const cacheKey = `${CACHE_TEAM_MEMBERS_KEY}/${teamId}`;
       let teamMembers = this.cache.get(cacheKey);
       if (!teamMembers) {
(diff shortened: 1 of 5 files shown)


bundle-size/github-utils.js
```diff
@@ -218,21 +220,13 @@ class GitHubUtils {
...
-    const teamIds = await Promise.all(
-      potentialApproverTeams.map(this.getTeamId_, this)
-    );
-    this.log(
-      `Reviewer for pull request ${pullRequestId} will be chosen from ` +
-        `[${potentialApproverTeams.join(', ')}] (team ids are ` +
-        `[${teamIds.join(', ')}])`
-    );
-    const potentialReviewers = await this.getTeamMembers_(teamIds);
+  async getRandomReviewer_(pullRequestId, potentialReviewers) {
```
rcebulko: This may not be new, but it's the first time I'm noticing it. I don't think adding an argument to the function just to include it in a log is worthwhile, and it's a bit misleading (I saw it and immediately assumed it was making some GitHub API call, esp. since this method is `async` (also, should it be?)). This logging can happen in calling code

bundle-size/github-utils.js
```diff
@@ -242,6 +236,73 @@ class GitHubUtils {
...
+      ...requestedReviewersResponse.data.users.map(user => user.login),
+      ...reviewsResponse.data.map(review => review.user.login),
+    ]);
+
+    const potentialReviewers = await this.getTeamMembers_(approverTeams);
+    if (
+      potentialReviewers.length &&
+      potentialReviewers.every(
+        potentialReviewer => !existingReviewers.has(potentialReviewer)
+      )
```
rcebulko: `[].some(...) ==> false`, so you can remove the length check and move the negation out to the top level. ie:
`if (!potentialReviewers.some(existingReviewers.has, existingReviewers))` (or the arrow-notation version).

bundle-size/github-utils.js
danielrozenberg: Makes sense. This was probably a side effect of refactoring it as a method extraction

bundle-size/github-utils.js
danielrozenberg: I got mixed up so much in the logic here that I ended up with this abomination above. Fixed.

bundle-size/github-utils.js
```diff
@@ -220,21 +220,15 @@ class GitHubUtils {
...
-    this.log(
-      `Chose reviewer ${reviewer} from all of ` +
-        `[${potentialReviewers.join(', ')}] for pull request ${pullRequestId}`
-    );
-
-    return reviewer;
+  async getRandomReviewer_(potentialReviewers) {
+    return potentialReviewers[
+      Math.floor(Math.random() * potentialReviewers.length)
+    ];
```
rcebulko: I can't believe Node doesn't have a builtin like Python's `random.choice` =/

danielrozenberg: Wholeheartedly agree

PR #649, opened 2019-12-05
[bundle-size] Verify whether a pull request is approved by a reviewer using the `approving_teams` field

danielrozenberg:
#617

For now this only logs the results, but does not assign or resolve the check - checks are resolved using the legacy logic until we're satisfied with the logs

### DIFF
--- a/bundle-size/webhooks.js
+++ b/bundle-size/webhooks.js
@@ -80,52 +80,55 @@ exports.installGitHubWebhooks = (app, db, githubUtils) => {
     const pullRequestId = context.payload.pull_request.number;
     const headSha = context.payload.pull_request.head.sha;
 
-    if (
-      context.payload.review.state == 'approved' &&
-      (await githubUtils.isBundleSizeApproverLegacy(approver))
-    ) {
-      context.log(
-        `Pull request ${pullRequestId} approved by a bundle-size keeper`
-      );
-
-      const check = await getCheckFromDatabase(db, headSha);
-      if (!check) {
-        context.log(
-          `Check ID for pull request ${pullRequestId} with head ` +
-            `SHA ${headSha} was not found in the database`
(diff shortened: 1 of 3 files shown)


bundle-size/webhooks.js
```diff
@@ -80,52 +80,74 @@ exports.installGitHubWebhooks = (app, db, githubUtils) => {
...
+    }
 
-      let approvalMessagePrefix;
-      if (check.delta === null) {
-        context.log(
-          'Pull requests can no longer be preemptively approved for ' +
-            'bundle-size changes'
-        );
+    context.log(
+      `Pull request ${pullRequestId} approved by a bundle-size keeper`
```
rcebulko: Where did "keeper" come from? :soccer:

bundle-size/webhooks.js
```diff
@@ -80,52 +80,74 @@ exports.installGitHubWebhooks = (app, db, githubUtils) => {
...
+    await context.github.checks.update({
+      owner: check.owner,
+      repo: check.repo,
+      check_run_id: check.check_run_id,
+      conclusion: 'success',
+      completed_at: new Date().toISOString(),
+      output: {
+        title: `${approvalMessagePrefix} | approved by @${approver}`,
+        summary:
+          'The bundle size (brotli compressed size of `v0.js`) of this pull ' +
```
rcebulko: Doesn't this apply to other binaries (ie. not just `v0.js`)?

bundle-size/webhooks.js
danielrozenberg: Old text that will be changed when we move to the new model

bundle-size/webhooks.js
danielrozenberg: Same

PR #650, opened 2019-12-05
[bundle-size] Also include super users when determining bundle size approvers

danielrozenberg:
#617

### DIFF
--- a/bundle-size/github-utils.js
+++ b/bundle-size/github-utils.js
@@ -165,7 +165,7 @@ class GitHubUtils {
   }
 
   /**
-   * Check whether the user is allowed to approve *all* bundle size change.
+   * Check whether the user is allowed to approve *all* bundle size changes.
    *
    * @param {string} username the username to check.
    * @return {boolean} true if the user is allowed to approve *all* bundle size
--- a/bundle-size/github-utils.js
+++ b/bundle-size/github-utils.js
@@ -173,20 +173,8 @@ class GitHubUtils {
    */
   async isBundleSizeSuperApprover(username) {
     return (
-      await this.getTeamMembers_(process.env.SUPER_USER_TEAMS.split(','))
+      await this.getTeamMembers(process.env.SUPER_USER_TEAMS.split(','))
     ).includes(username);
-  }
-
-  /**
-   * Check whether the user is allowed to approve a bundle size change.
-   *
-   * @param {string} username the username to check.
-   * @param {!Array<string>} approverTeams list of all the teams whose members
-   *   can approve the bundle-size change of this pull request.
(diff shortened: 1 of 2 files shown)


danielrozenberg:
Thinking about this, this is not how I want this to work. Closing for now

danielrozenberg:
Try now

bundle-size/github-utils.js
```diff
@@ -164,6 +164,19 @@ class GitHubUtils {
     return false;
   }
 
+  /**
+   * Check whether the user is allowed to approve *all* bundle size change.
```
rcebulko: Nit: /change/changes/s

Also, `*all*` makes me read this comment with jazz hands :open_hands:

rcebulko: I think rather than emphasizing "all" here, it might be better to clarify in comments for `isBunderSizeApprover` that it only applies to...something. As I look closer and try to complete that sentence, it makes me think `isBundleSizeApprover` might be better named `isMemberOfTeam` or something similar.

Irrespective of naming, you could save/reuse a little bit of work here by sayng

```javascript
async isBundleSizeSuperApprover(username) {
  return await isBundleSizeApprover(username, process.env.SUPER_USER_TEAMS.split(',')); 
}
```

bundle-size/webhooks.js
```diff
@@ -93,12 +93,26 @@ exports.installGitHubWebhooks = (app, db, githubUtils) => {
       return;
     }
 
+    const isBundleSizeSuperApprover = await githubUtils.isBundleSizeSuperApprover(
```
rcebulko: Nit: `BundleSize` is obvious here. If it were just `isSuperApprover` it would fit on one line :laughing:

bundle-size/webhooks.js
```diff
@@ -120,20 +134,11 @@ exports.installGitHubWebhooks = (app, db, githubUtils) => {
...
     );
 
-    let approvalMessagePrefix;
-    if (check.delta === null) {
-      context.log(
-        'Pull requests can no longer be preemptively approved for ' +
-          'bundle-size changes'
-      );
+    const bundleSizeDelta = parseFloat(check.delta);
+    if (bundleSizeDelta <= process.env['MAX_ALLOWED_INCREASE']) {
```
rcebulko: Why use a string key here versus dot-access notation like above?

rcebulko: Oh it's old code

bundle-size/github-utils.js
danielrozenberg: Makes sense. Done.

bundle-size/webhooks.js
danielrozenberg: Done.

PR #657, opened 2019-12-16
[bundle-size] Use the app-based GitHub token, not the user-based one, to add reviewers

danielrozenberg:
When I deployed the app after the changes in #646 this action threw permissions errors, I believe the user-based token can't add reviewers. In any case, it makes sense to do this via the app as this will say which bot added the reviewer, as opposed to just @ampprojectbot

Tag-along change:
* Remove node cache's standard TTL, as each cached object has its own TTL

### DIFF
--- a/bundle-size/api.js
+++ b/bundle-size/api.js
@@ -172,6 +172,44 @@ function choosePotentialApproverTeams(
  */
 exports.installApiRouter = (app, db, githubUtils) => {
   /**
+   * Add a bundle size reviewer to add to the pull request.
+   *
+   * Does nothing if there is already a reviewer that can approve the bundle
+   * size change.
+   *
+   * @param {!Octokit} github an authenticated GitHub API object.
+   * @param {!Octokit.PullsListReviewRequestsParams} pullRequest GitHub Pull
+   *   Request params.
+   * @param {!Array<string>} approverTeams list of all the teams whose members
+   *   can approve the bundle-size change of this pull request.
+   * @return {!Octokit.Response<Octokit.PullsCreateReviewRequestResponse>}
+   *   response from GitHub API.
+   */
+  async function addReviewer_(github, pullRequest, approverTeams) {
+    const newReviewer = await githubUtils.chooseReviewer(
+      pullRequest,
+      approverTeams
+    );
+    if (newReviewer !== null) {
+      try {
+        // Choose a random capable username and add them as a reviewer to the pull
+        // request.
+        return await github.pullRequests.createReviewRequest({
+          reviewers: [newReviewer],
+          ...pullRequest,
+        });
+      } catch (error) {
+        app.log.error(
+          'ERROR: Failed to add a reviewer to pull request ' +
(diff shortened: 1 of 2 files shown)


bundle-size/github-utils.js
```diff
@@ -40,7 +40,7 @@ function getBuildArtifactsFileParams_(filename) {
 }
 
 /**
- * Utils for GitHub actions.
+ * Utils for GitHub actions that are performed by the user based token.
```
rcebulko: Nit: "performed with a user-authenticated token" or something similar is a bit clearer to people who aren't as steeped in the types of tokens. A link to docs with the two types, here or somewhere else, might be useful.

bundle-size/github-utils.js
```diff
@@ -51,12 +51,7 @@ class GitHubUtils {
   constructor(github, log, cache) {
     this.github = github;
     this.log = log;
-    this.cache =
-      cache ||
-      new NodeCache({
-        'stdTTL': CACHE_APPROVERS_TTL_SECONDS,
-        'checkperiod': CACHE_CHECK_SECONDS,
-      });
+    this.cache = cache || new NodeCache({'checkperiod': CACHE_CHECK_SECONDS});
```
rcebulko: Intentional removal of `stdTTL`?

bundle-size/github-utils.js
estherkim: Which account does this token  `process.env.ACCESS_TOKEN` belong to?

bundle-size/api.js
```diff
@@ -171,6 +171,44 @@ function choosePotentialApproverTeams(
  * @param {!GitHubUtils} githubUtils GitHubUtils instance.
  */
 exports.installApiRouter = (app, db, githubUtils) => {
```
estherkim: Just to clarify, `api.js` uses a GitHub API client with ampprojectbot auth, and `GitHubUtils` uses another GitHub API client with this user-based auth? Why do we need 2 different auths?

bundle-size/github-utils.js
danielrozenberg: @ampprojectbot

danielrozenberg: Done

bundle-size/github-utils.js
danielrozenberg: Yes, it's a relic from the earlier implementation. There is no standard TTL for this cache, as each cached object has its own TTL. Updating the PR description

bundle-size/api.js
danielrozenberg: As we are all very well aware, the GitHub API documentation is lacking... There are certain actions that app-based tokens can't perform, no matter the permissions we give them - any action relating to Organization (such as translating team names to team numbers and members) can only be performed with a user-based token.
Now it turns out that creating a review request cannot be performed with the user-based token, so I'm switching it back to the app-based token.
I wish it wasn't so. Or at least better documented...

PR #661, opened 2019-12-19
Add documentation to explain the use of multiple GitHub access tokens

danielrozenberg:


GITHUB_API_NUANCE.md
```diff
@@ -0,0 +1,80 @@
+# Nuance in the GitHub API with Probot
+
+Many of the applications in this repository use the [Probot](https://probot.github.io/) framework to power GitHub apps.
+In the source code of some of these apps (e.g., `bundle-size`) you will notice that we define an environment variable
+called `GITHUB_ACCESS_TOKEN`, which should be set to a [personal access token](https://github.com/settings/tokens)
+generated by the user:
+
+###### [bundle-size/.env.example](https://github.com/ampproject/amp-github-apps/blob/33969c884f51a5fcfe342202b1935145b64efd1b/bundle-size/.env.example#L25-L26)
+```sh
+…
```
rsimha: nit: The ellipses here and elsewhere in the doc look odd in MD view. Maybe remove?

GITHUB_API_NUANCE.md
```diff
@@ -0,0 +1,80 @@
...
+
+## So why the duality?
+
+Unfortunately, the GitHub API permission scopes are confusing and not always in line with the documentation.
+
+There is currently no way, for example, to get read-only access to organization teams and their members as a GitHub App
+(only full administration permissions can be given to an app), but it is possible using a personal access token of a
+member of the organization, if that access token was given the `read:org` permission.
+
+Contrary, it is not currently possible to assign a reviewer to a pull request with a personal access token, but it is
```
rsimha: nit: Should this be `On the contrary,`? Or is `Contrary,` okay? (I'm not sure :))

GITHUB_API_NUANCE.md
```diff
@@ -0,0 +1,80 @@
+# Nuance in the GitHub API with Probot
```
rcebulko: Nit: I feel like `github-access-tokens.md` or something similar might be clearer than `GITHUB_API_NUANCE.md`; similar for the title, highlighting that it's more about tokens than the API

GITHUB_API_NUANCE.md
rcebulko: +1

GITHUB_API_NUANCE.md
```diff
@@ -0,0 +1,80 @@
...
+
+Many of the applications in this repository use the [Probot](https://probot.github.io/) framework to power GitHub apps.
+In the source code of some of these apps (e.g., `bundle-size`) you will notice that we define an environment variable
+called `GITHUB_ACCESS_TOKEN`, which should be set to a [personal access token](https://github.com/settings/tokens)
+generated by the user:
+
+###### [bundle-size/.env.example](https://github.com/ampproject/amp-github-apps/blob/33969c884f51a5fcfe342202b1935145b64efd1b/bundle-size/.env.example#L25-L26)
```
rcebulko: Nit: This doesn't really need to be a heading, can just be `*[bundle-size/.env.example](...)*`

GITHUB_API_NUANCE.md
```diff
@@ -0,0 +1,80 @@
...
+  const userBasedGithub = new Octokit({
+    'auth': process.env.ACCESS_TOKEN,
+  });
+
+  const githubUtils = new GitHubUtils(userBasedGithub, app.log);
+…
+```
+
+An experienced GitHub Apps developer would wonder why this is necessary, as GitHub Apps have their own way to
+authenticate themselves with the GitHub API, using the App ID and its private key, which we also use in as environment
```
rcebulko: `/also use in as/also provide as/s`

GITHUB_API_NUANCE.md
```diff
@@ -0,0 +1,80 @@
...
+
+Inside Probot Webhook handlers, the `context` object has a `github` field that is already authenticated to perform API
+calls with the permissions given to the app:
+
+###### [bundle-size/webhooks.js](https://github.com/ampproject/amp-github-apps/blob/33969c884f51a5fcfe342202b1935145b64efd1b/bundle-size/webhooks.js#L42)
+
+```javascript
+…
+      },
+    });
```
rcebulko: I know these are actual code samples, but I think some of the surrounding bits don't add as much clarity as simpler/stripped down examples would. These braces definitely done add much :stuck_out_tongue:

GITHUB_API_NUANCE.md
```diff
@@ -0,0 +1,80 @@
...
+Whereas in other contexts we use the `githubUtils` object that we authenticated using the personal access token:
+
+###### [bundle-size/app.js](https://github.com/ampproject/amp-github-apps/blob/33969c884f51a5fcfe342202b1935145b64efd1b/bundle-size/webhooks.js#L96)
+```javascript
+…
+    }
+
+    const isSuperApprover = await githubUtils.isSuperApprover(approver);
+    context.log(
+      `Approving user ${approver} of pull request ${pullRequestId}`,
```
rcebulko: Similar to above, this code doesn't provide a very clear view of the difference in general, since `githubUtils` is something unique to the `bundle-size` project (by contrast, `owners` has a single `GitHub` class which accepts an Octokit instance from the caller). I feel like more generic examples, possibly showing the Octokit instance from a Probot `context` versus manually creating one, would be more illustrative.

GITHUB_API_NUANCE.md
danielrozenberg: Done

GITHUB_API_NUANCE.md
danielrozenberg: Both are fine afaik

GITHUB_API_NUANCE.md
danielrozenberg: Done

GITHUB_API_NUANCE.md
danielrozenberg: This is a hack for formatting :)

GITHUB_API_NUANCE.md
danielrozenberg: Done

GITHUB_API_NUANCE.md
danielrozenberg: Done

GITHUB_API_NUANCE.md
danielrozenberg: Done

danielrozenberg:
> I like the idea behind this, but I worry that A) most of the code examples are very specific to `bundle-size` and B) this doesn't provide much beyond what https://developer.github.com/apps/differences-between-apps/ documents. I'm not sure if this deserves its own doc, as opposed to a small section in the top-level repo README linking to the GitHub docs on auth types

* A) addressed
* B) none of this information is in the link you provided. This isn't something that seems to be documented correctly in their official docs :(

danielrozenberg:
PTAL at recent changes

github-access-tokens.md
```diff
@@ -0,0 +1,79 @@
...
+
+```javascript
+// app.js
+
+module.exports = app => {
+  const userBasedGithub = new Octokit({
+    'auth': process.env.ACCESS_TOKEN,
+  });
+
+  app.on('pull_request.opened', /* ... */);
```
rcebulko: Nit: drop this maybe? Doesn't add much imo

danielrozenberg: Dropping this could confuse people who follow this as if this is a code-lab :)

rcebulko: I'm not sure what that means, but w.e.

Issue #664, opened 2020-01-08
I2I: Automated GitHub App deployments


rcebulko:
## Overview

While the code for these apps is open-source, deploying them often requires the author to execute a command which deploys the app using a `.env` file on their local system. This makes it difficult or impossible for other team members to safely deploy or update apps by other members. We can't just check in the actual `.env` files, as they contain sensitive secret keys, OAuth tokens, etc.

I propose using GCloud's Cloud Builds feature to automate deployment. Cloud Builds can be configured to automatically trigger in response to pushes to predetermined (and restricted) branches.

## How to configure auto-deployment for an app

Setting aside a few bookkeeping steps involved with granting API access to various service accounts, there are only a handful of steps required to enable Cloud Build for a given project.

### 1) Encrypt the environment file via Cloud KMS

We can use a Cloud KMS keyring to locally encrpyt the `.env` file and check it into the repo as `.env.enc` (this is the recommended practice). A build step is then able to use Cloud KMS to decrypt that `.env` file and make it available for the app to access, without making it visible within the public-facing repo.

We could optionally make some steps easier by providing top-level npm scripts such as `npm run encrypt-env bundle-size-chart/.env` and the equivalent for decrypting. Note that contributors with access to the Cloud KMS keyring would still be able to decrypt and read the original `.env` file.

> It's worth noting that it's also possible within Cloud Build/Cloud KMS to encrypt environment variables, but doing so requires encrypting each environment key and then providing them all in the cloud build configuration, which would be more work and would diverge from the local development workflow of maintaining a single `.env` file.

### 2) Define the Cloud Build configuration

We create a `cloud_build.yaml` file with the required build and deploy steps. Take the `bundle-size-chart` app as an example. The `cloud_build.yaml` file looks like:

```yaml
steps:
- name: gcr.io/cloud-builders/npm
  dir: bundle-size-chart
  args:
    - install
- name: gcr.io/cloud-builders/gcloud
  dir: bundle-size-chart
  args:
  - kms
  - decrypt
  - --ciphertext-file=.env.enc
  - --plaintext-file=.env
  - --location=global
  - --keyring=amp-github-apps-keyring
  - --key=bundle-size-chart-env-key
- name: gcr.io/cloud-builders/gcloud
  dir: bundle-size-chart
  args:
    - --project
    - amp-bundle-size-chart-test
    - app
    - deploy
- name: gcr.io/cloud-builders/gcloud
  dir: bundle-size-chart
  args:
[...]

PR #665, opened 2020-01-09
bundle-size-chart auto-deploy configuration + helpers

rcebulko:
Demonstrates the set-up required for configuring an app to auto-deploy through some trigger (exact trigger TBD, but is independent of this config).

Also includes a helper script to encrypt a local `.env` file. Note that the individual keys need to be created.

Part of #664


rcebulko:
Obsolete; closing

PR #666, opened 2020-01-10
Silence expected logging during `bundle-size` and  `test-status` tests

rcebulko:
Sets the standard log_level for tests to WARN
Drops log level to FATAL when testing operations that are expected to raise+log an error
Also fixes an `async` issue that kept Jest from exiting at the end of a test run

Drops the number of output lines for `bundle-size` tests from 436 to 4.
Drops the number of output lines for `test-status` tests from 66 to 4.

/cc @ampproject/wg-infra

### DIFF
--- a/bundle-size/package.json
+++ b/bundle-size/package.json
@@ -17,7 +17,7 @@
     "deploy": "gcloud --project amp-bundle-size-bot app deploy",
     "setup-db": "node ./setup-db.js",
     "start": "probot run ./app.js",
-    "test": "LOG_LEVEL=warn jest"
+    "test": "jest"
   },
   "dependencies": {
     "@octokit/rest": "16.35.2",
--- a/bundle-size/package.json
+++ b/bundle-size/package.json
@@ -17,7 +17,7 @@
     "deploy": "gcloud --project amp-bundle-size-bot app deploy",
     "setup-db": "node ./setup-db.js",
     "start": "probot run ./app.js",
-    "test": "jest"
+    "test": "LOG_LEVEL=warn jest"
   },
   "dependencies": {
     "@octokit/rest": "16.35.2",
--- a/bundle-size/test/api.test.js
+++ b/bundle-size/test/api.test.js
@@ -679,9 +679,8 @@ describe('bundle-size api', () => {
         )
         .reply(418, 'I am a tea pot');
 
-      // Silence expected error messages during testing.
-      const logLevel = probot.logger.level();
-      probot.logger.level('fatal');
+      // Stub app.log.error to silence test log noise for expected errors
+      const logErrorSpy = jest.spyOn(app.log, 'error').mockImplementation();
 
       await request(probot.server)
         .post('/v0/commit/5f27002526a808c5c1ad5d0f1ab1cec471af0a33/store')
@@ -691,7 +690,10 @@ describe('bundle-size api', () => {
         .expect(500, /I am a tea pot/);
       nocks.done();
 
-      probot.logger.level(logLevel);
+      expect(logErrorSpy).toHaveBeenCalledWith(
+        expect.stringContaining('Failed to create the bundle-size/5f27002526a808c5c1ad5d0f1ab1cec471af0a33.json file'),
+        expect.any(Error),
+      );
     });
 
     test('fail on non-numeric values when called to store bundle-size', async () => {
--- a/bundle-size/test/api.test.js
+++ b/bundle-size/test/api.test.js
@@ -34,6 +34,7 @@ describe('bundle-size api', () => {
   let app;
   const db = dbConnect();
   const nodeCache = new NodeCache();
+  let logWarnSpy;
 
   beforeAll(async () => {
     await setupDb(db);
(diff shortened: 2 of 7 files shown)


bundle-size/package.json
```diff
@@ -17,7 +17,7 @@
     "deploy": "gcloud --project amp-bundle-size-bot app deploy",
     "setup-db": "node ./setup-db.js",
     "start": "probot run ./app.js",
-    "test": "jest"
+    "test": "LOG_LEVEL=warn jest"
```
danielrozenberg: That's not good, this means if I run `npm test` locally it will ignore the value from the environment. Remove this and set the LOG_LEVEL in Travis

bundle-size/test/api.test.js
```diff
@@ -33,25 +33,29 @@ describe('bundle-size api', () => {
   let probot;
   let app;
   const db = dbConnect();
-  const nodeCache = new NodeCache();
+  let nodeCache;
+  let logWarnSpy;
 
-  beforeAll(async () => {
+  beforeAll(async done => {
```
danielrozenberg: Why was this change needed?

bundle-size/test/api.test.js
```diff
@@ -33,25 +33,29 @@ describe('bundle-size api', () => {
...
     await setupDb(db);
 
     probot = new Probot({});
     app = probot.load(app => {
       const githubUtils = new GitHubUtils(new Octokit(), app.log, nodeCache);
       installApiRouter(app, db, githubUtils);
     });
+    // Stub app.log.warn to silence test log noise
+    logWarnSpy = jest.spyOn(app.log, 'warn').mockImplementation();
```
danielrozenberg: Why not just set LOG_LEVEL to be higher than `warn` in that case?

bundle-size/test/api.test.js
```diff
@@ -33,25 +33,29 @@ describe('bundle-size api', () => {
...
     app.app = {
       getInstallationAccessToken: () => Promise.resolve('test'),
     };
+    done();
   });
 
-  beforeEach(async () => {
-    nodeCache.flushAll();
+  beforeEach(() => {
+    nodeCache = new NodeCache();
```
danielrozenberg: 🤔

bundle-size/test/api.test.js
```diff
@@ -83,13 +87,17 @@ describe('bundle-size api', () => {
       .reply(200, getFixture('teams.listMembers.3065813'));
   });
 
-  afterEach(async () => {
+  afterEach(async done => {
```
danielrozenberg: Same, why do you need `done`?

bundle-size/test/api.test.js
```diff
@@ -673,13 +681,19 @@ describe('bundle-size api', () => {
         )
         .reply(418, 'I am a tea pot');
 
+      // Silence expected error messages during testing.
+      const logLevel = probot.logger.level();
```
danielrozenberg: *Here* I would prefer that you mock it instead of changing the log level (so that if something changes in the code and the test fails, we can still get some idea as to why - i.e. by expecting the log to be have been called)

bundle-size/test/webhooks.test.js
```diff
@@ -221,6 +221,11 @@ describe('bundle-size webhooks', () => {
         name: 'pull_request',
         payload: pullRequestPayload,
       });
+
+      // Silence expected error messages during testing.
+      const logLevel = probot.logger.level();
```
danielrozenberg: ^

test-status/package.json
```diff
@@ -17,7 +17,7 @@
     "deploy": "gcloud --project amp-test-status-bot app deploy",
     "setup-db": "node ./setup-db.js",
     "start": "probot run ./app.js",
-    "test": "jest"
+    "test": "LOG_LEVEL=warn jest"
```
danielrozenberg: ^

bundle-size/package.json
rcebulko: Ahh good to know, done.

bundle-size/test/api.test.js
rcebulko: `async` was not necessary; there's no `await`.
The `nodeCache` instance was leaving dangling `setTimeouts` the way it was being reused (Jest gave me warnings about this), and there's no need to share a cache between the tests

bundle-size/test/api.test.js
rcebulko: Made these tweaks when I was trying to find the culprit preventing Jest from exiting properly. Reverting (the node cache was the issue)

bundle-size/test/api.test.js
rcebulko: See above; reverted.

test-status/package.json
rcebulko: Done

bundle-size/test/api.test.js
rcebulko: Done.

bundle-size/test/webhooks.test.js
rcebulko: In this case, since you want the error to propagate outside the Probot handler, the message is automatically logged by Probot's error event handler (as opposed to the app/context/request loggers). As best as I can tell, there's no direct access to these methods (they're either private or non-explicit, somehow embedded in Probot error catching code). The change in log level still takes effect, though.

bundle-size/test/api.test.js
rcebulko: I opted for this because every single test logs a warning about the bundle size not existing, but if you had logging enabled for debugging purposes (at INFO level), you'd still be able to see all the info logs.

danielrozenberg: Ack

bundle-size/test/webhooks.test.js
danielrozenberg: Ack

PR #673, opened 2020-01-28
Add helper script to generate the unredacted .env file

rcebulko:


bin/unredact-env.js
```diff
@@ -0,0 +1,106 @@
...
+function main(appName) {
+  if (!appName) {
+    throw new Error('Must specify app argument to replace-secrets');
+  }
+  const appDir = path.resolve(appName);
+
+  if (!fs.existsSync(appDir)) {
+    throw new Error(`Invalid app "${appName}"; directory does not exist`);
+  }
+
```
danielrozenberg: nicetohave: "This will overwrite existing .env file. Continue? " prompt (not a blocker)

rcebulko: Added a semi-hacky bash prompt into the NPM script, since actually implementing it in node would either mean a non-trivial amount of i/o code with stdin or a new dependency.

PR #674, opened 2020-01-29
Configure auto-deployment for owners app

rcebulko:


bundle-size-chart/cloud_build.yaml
```diff
@@ -10,7 +10,8 @@ steps:
       - replace-secrets
       - --
       - bundle-size-chart
-    secretEnv: ['ACCESS_TOKEN']
+    secretEnv:
+      - ACCESS_TOKEN
```
danielrozenberg: Was wondering about that initial choice 🙃

owners/cloud_build.yaml
```diff
@@ -0,0 +1,44 @@
...
+      - app
+      - deploy
+      - cron.yaml
+
+# These secrets are the base64-encoded form of encrypted secret values. They are
+# automatically decrypted and added to the `.env` environment at build time.
+secrets:
+  - kmsKeyName: projects/ampproject-owners-bot/locations/global/keyRings/amp-github-apps-keyring/cryptoKeys/app-env-key
+    secretEnv:
+      WEBHOOK_SECRET: CiQAs8/dw/BTKMPqcNPIpQZVCdYi6TpBEwAaTuoHbtjfdJWHv28SOQC3YjA4n5HlcR/fSVmVNlB6m1W8lY5FE6oNJFsuqRqtVgirqFQjIN3RVKbiR5NmYKr2OrP0CMA33A==
```
danielrozenberg: Interesting observation - the base64-encoding of all the secrets in all the apps start with `CiQA`. pre-push hook to ban values that don't start with it from being pushed?

owners/redacted.env
```diff
@@ -0,0 +1,16 @@
+NODE_ENV=production
+LOG_LEVEL=info
+PORT=8080
+APP_ID=22611
+
+CLOUD_STORAGE_BUCKET=ampproject-owners-bot.appspot.com
+# GOOGLE_APPLICATION_CREDENTIALS=/path/to/app-credentials-abcdef12345.json
+
+WEBHOOK_SECRET=[REDACTED]
```
danielrozenberg: Another possible pre-push/pr-check: verify that there's a 1:1 matching between the redacted variables in `redacted.env`, and the `steps.secretEnv`/`secrets.secretEnv` keys

owners/cloud_build.yaml
rcebulko: Indeed an interesting observation, but I'm not sure if that's a bit overboard to have a pre-push hook. Why do you think it's worthwhile here while, say, the environment files would be an easier place to accidentally include a secret?

owners/redacted.env
rcebulko: Same opinion as above. There's a lot of room for validation and automation of various parts of this, but given that changes to these files are so few and far between, I'm not sure if the effort would be worthwhile. Merging this PR, but feel free to continue this or the above thread.

owners/cloud_build.yaml
danielrozenberg: I can see a situation in which one pastes an already base64-encoded value (like the PRIVATE_KEY) without passing it through the encrypter. Extremely minor though...

owners/redacted.env
danielrozenberg: Fair. P3 fixit week? :D

rcebulko: ¯\_(ツ)_/¯

PR #679, opened 2020-02-04
Update DEPLOYMENT.md

danielrozenberg:
Nit


DEPLOYMENT.md
```diff
@@ -57,7 +57,7 @@
     - For the tag pattern, use `deploy-{your-app-name}-\d{14}` (ex. `deploy-bundle-size-20200122154000` would be the deploy tag for the `bundle-size` app, with a timestamp of 15:40:00 on 2020-01-22). For consistency, deploy using the timestamp in the UTC timezone.
     - Under _Build Configuration_, select _Cloud Build configuration file_ and provide the path to your Cloud Build file, ex. `bundle-size/cloud_build.yaml`
 10. Add the NPM script `deploy-tag`, which creates a git tag in the proper tag format for your app
-    - Ex. ``git tag 'deploy-bundle-size-'`date --utc '+%Y%m%d%H%M%S'` ``
+    - Ex. ``git tag 'deploy-{your-app-name}-'`date --utc '+%Y%m%d%H%M%S'` ``
```
rcebulko: I already have that above with "For the tag pattern, use `deploy-{your-app-name}-\d{14}`. This line is an example, not a template. If you don't like having "bundle-size" in the example, I would at least want to make it `deploy-your-app-name-`

danielrozenberg: Done

PR #684, opened 2020-02-04
[multiple apps] Fix Octokit related deprecation warnings

danielrozenberg:
Mostly small fixes, except in bundle-size/github-utils.js where there was some logic change required. If this PR looks too large, review it commit-by-commit.

Note that the remaining deprecation warnings in Travis are from within the Probot library and I can only assume will be fixed soon


bundle-size/github-utils.js
```diff
@@ -129,50 +127,27 @@ class GitHubUtils {
...
-    return teamId;
-  }
-
   /**
    * Get a list of all unique team members by team names.
    *
-   * @param {!Array<string>} teamNames names of full team slug namesto get
-   *   members of.
+   * @param {!Array<string>} teamNames names of full team slug names
+   *   ("org/team_slug") format to get members of.
```
rcebulko: This is really hard to parse in the first couple tries (at least for me). Maybe something like
`@param {!Array<string>} teamNames GitHub team slug ex. "myorg/team_slug"`
"to get members of" probably doesn't add much

bundle-size/github-utils.js
```diff
@@ -129,50 +127,27 @@ class GitHubUtils {
...
-      const teamId = await this.getTeamId_(teamName);
-      const cacheKey = `${CACHE_TEAM_MEMBERS_KEY}/${teamId}`;
+      const cacheKey = `${CACHE_TEAM_MEMBERS_KEY}/${teamName}`;
       let teamMembers = this.cache.get(cacheKey);
       if (!teamMembers) {
         this.log(`Cache miss for ${cacheKey}. Fetching from GitHub...`);
+        const [org, teamSlug] = teamName.split('/', 2);
         teamMembers = await this.github.teams
-          .listMembers({team_id: teamId})
+          .listMembersInOrg({org, team_slug: teamSlug})
```
rcebulko: Cool that this exists now

bundle-size/github-utils.js
```diff
@@ -129,50 +127,27 @@ class GitHubUtils {
...
           .then(result => result.data.map(user => user.login));
         this.log(
-          `Fetched team members [${teamMembers.join(', ')}] for team id ` +
-            `${teamId}. Caching for ${CACHE_TEAM_MEMBERS_TTL_SECONDS} seconds.`
+          `Fetched team members [${teamMembers.join(', ')}] for team ` +
+            `${teamName}. Caching for ${CACHE_TEAM_MEMBERS_TTL_SECONDS} ` +
+            'seconds.'
```
rcebulko: Note to nobody: This could be shared infra, owners does its own cache-refresh pattern with teams too

owners/src/api/github.js
```diff
@@ -231,6 +231,11 @@ class GitHub {
   async getTeamMembers(teamId) {
     this.logger.info(`Fetching team members for team with ID ${teamId}`);
 
+    // TODO(rcebulko): teams.listMembers is deprecated, replace this with
```
rcebulko:
```suggestion
    // TODO(#685): teams.listMembers is deprecated, replace this with
```

bundle-size/github-utils.js
danielrozenberg: Ack

bundle-size/github-utils.js
danielrozenberg: Yes :)

bundle-size/github-utils.js
danielrozenberg: Changed this. wdyt?

bundle-size/github-utils.js
```diff
@@ -130,8 +130,8 @@ class GitHubUtils {
   /**
    * Get a list of all unique team members by team names.
    *
-   * @param {!Array<string>} teamNames names of full team slug names
-   *   ("org/team_slug") format to get members of.
+   * @param {!Array<string>} teamNames list of GitHub team (full) slugs,
```
rcebulko: LGTM, would drop `(full)` since I don't think it adds anything esp with the examples. If you wanted to clarify it further, you could explicitly say org-namespaced slugs, but I think it's overkill.

danielrozenberg: Dropped it.

PR #692, opened 2020-02-10
Add method stubs and tests for GitHub API interface

rcebulko:
Taking a test-driven approach, defining all the class interfaces and expectations first, then filling in the implementations, so as to allow for better parallelism

### DIFF
--- a/invite/src/github.ts
+++ b/invite/src/github.ts
@@ -1,0 +1,72 @@
+/**
+ * Copyright 2020 The AMP HTML Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS-IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {Octokit} from '@octokit/rest';
+import {GitHubAPI} from 'probot/lib/github';
+
+import {ILogger} from './types';
+
+/**
+ * Interface for working with the GitHub API.
+ */
+export class GitHub {
+  private client: Octokit;
+  private org: string;
+  private logger: ILogger;
+
+  /**
+   * Constructor.
+   */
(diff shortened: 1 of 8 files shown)


invite/src/github.ts
```diff
@@ -0,0 +1,72 @@
...
+   * Constructor.
+   */
+  constructor(client: Octokit, org: string, logger: ILogger = console) {
+    Object.assign(this, {client, org, logger});
+  }
+
+  /**
+   * Checks whether a user is a member of the organization.
+   */
+  async userIsMember(username: string): Promise<boolean> {
```
danielrozenberg: Curious - `async` methods still need to declare their return type as `Promise<something>` instead of just `something`?

invite/test/github.test.ts
```diff
@@ -0,0 +1,137 @@
...
+
+  afterEach(() => {
+    // Fail the test if there were unused nocks.
+    if(!nock.isDone()) {
+      throw new Error('Not all nock interceptors were used!');
+      nock.cleanAll();
+    }
+  });
+
+  describe('userIsMember', () => {
```
danielrozenberg: Change the tests so that they'll expect the correct value for each case, then change the `it` to `it.skip`. Better have the real test at the ready

invite/test/github.test.ts
```diff
@@ -0,0 +1,137 @@
...
+  });
+
+  beforeEach(() => {
+    github = new GitHub(githubClient, 'test_org');
+  });
+
+  afterEach(() => {
+    // Fail the test if there were unused nocks.
+    if(!nock.isDone()) {
+      throw new Error('Not all nock interceptors were used!');
```
danielrozenberg: I believe `nock.done()` throws the error for you if it is not done

invite/src/github.ts
rcebulko: In Typescript, yes

invite/test/github.test.ts
rcebulko: I get an error saying `done` does not exist on nock, so gonna leave it for now :upside_down_face:

invite/test/github.test.ts
rcebulko: These tests already are expecting the correct value (they currently fail). Added the `.skip` to the top `describe`.

invite/test/github.test.ts
danielrozenberg: Oh it must be a nock scope thing then...

rcebulko: `¯\_(ツ)_/¯`

PR #693, opened 2020-02-10
[invite] Add method stubs and tests for invitation record

rcebulko:
Taking a test-driven approach, defining all the class interfaces and expectations first, then filling in the implementations, so as to allow for better parallelism

This invitation record will eventually contain an implementation backed either by Cloud SQL or Cloud Datastore, which will record invites as they are sent so that the bot can later update the threads and assign issues if/when the invites are accepted.

```
invitation record
  recordInvite
    ✕ records an invite (1ms)
    ✕ sets `archived = false`
  getInvites
    if no invite record exists for the user
      ✓ returns an empty list (1ms)
    if records exists of invites to the user
      ✕ returns the invites (3ms)
    if only archived invite records exists for the user
      ✓ returns an empty list
  archiveInvites
    if records exists of any invites to the user
      ✕ updates the invite records (1ms)
```

### DIFF
--- a/invite/test/invitation_record.test.ts
+++ b/invite/test/invitation_record.test.ts
@@ -17,7 +17,7 @@
 import {Invite, InviteAction} from '../src/types';
 import {InvitationRecord} from '../src/invitation_record';
 
-describe('invitation record', () => {
+describe.skip('invitation record', () => {
   let record: InvitationRecord;
   const invite: Invite = {
     username: 'someone',
--- a/invite/test/invitation_record.test.ts
+++ b/invite/test/invitation_record.test.ts
@@ -32,15 +32,8 @@ describe.skip('invitation record', () => {
   });
 
   describe('recordInvite', () => {
-    // TODO: Update once the storage interface is determined.
-    it('records an invite', async done => {
-      throw new Error('Not implemented!');
-      done();
-    });
-    it('sets `archived = false`', async done => {
-      throw new Error('Not implemented!');
-      done();
-    });
+    it.todo('records an invite');
+    it.todo('sets `archived = false`');
   });
 
   describe('getInvites', () => {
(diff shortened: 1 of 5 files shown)


invite/test/invitation_record.test.ts
```diff
@@ -0,0 +1,100 @@
...
+  };
+  const archivedInvite: Invite = Object.assign({archived: true}, invite);
+
+  beforeEach(() => {
+    record = new InvitationRecord();
+  });
+
+  describe('recordInvite', () => {
+    // TODO: Update once the storage interface is determined.
+    it('records an invite', async done => {
```
danielrozenberg: `it.todo('records an invite');`

rcebulko: > Is storage really necessary? You can probably query the API to see whether a user is pending an invitation, right?

This is to record invites triggered by comments. For example, suppose Joey comments `/tryassign @somebody` on two issues and `/invite @somebody` on a third. When that invite is eventually accepted, we want to assign those two issues to `@somebody` and add comments on all three updating the thread that the invited user has joined. This metadata isn't available in the invite, and so the mapping to issues from which users were invited needs to be persisted.

danielrozenberg: Counterpoint - can you use the API to search for all comments containing `/invite @somebody` and `/tryassign @somebody` and work your way from there? 🙃
Stateless code is code that's easier to debug and reason about, so I always prefer to try that

rcebulko: a) Search is limited to one repository, while this invite-bot will operate organization-wide
b) Search API is ugly and nondeterministic and has no consistent ordering, which completely breaks pagination (to find all OWNERS files with the search API, I literally just have to try multiple times and hope I got them all)
c) Persisting the invites to act on also acts as a record log for debugging and reasoning about the actions the bot takes

danielrozenberg: Fair. Proceed!

rcebulko: I shall! Iff you approve :upside_down_face:

PR #695, opened 2020-02-11
[invite] Implement GitHub API interface methods

rcebulko:
```
GitHub interface
  inviteUser
    ✓ PUTs to /orgs/:org/memberships/:username (19ms)
    ✓ returns true when the user is invited (2ms)
    ✓ returns false when the user is already a member (1ms)
  addComment
    ✓ POSTs comment to /repos/:owner/:repo/issues/:issue_number/comments (2ms)
  assignIssue
    ✓ POSTs assignee to /repos/:owner/:repo/issues/:issue_number/assignees (1ms)
```

### DIFF
--- a/invite/src/github.ts
+++ b/invite/src/github.ts
@@ -70,8 +70,12 @@ export class GitHub {
     issue_number: number,
     assignee: string
   ): Promise<void> {
-    // https://octokit.github.io/rest.js/#octokit-routes-issues-add-assignee
-    // octokit.issues.addAssignees({owner, repo, issue_number, assignees})
+    await this.client.issues.addAssignees({
+      owner: this.org,
+      repo,
+      issue_number,
+      assignees: [assignee],
+    });
   }
 }
 
--- a/invite/src/github.ts
+++ b/invite/src/github.ts
@@ -25,24 +25,25 @@ import {ILogger} from './types';
  * Interface for working with the GitHub API.
  */
 export class GitHub {
-  private client: Octokit;
-  private org: string;
-  private logger: ILogger;
-
   /**
    * Constructor.
    */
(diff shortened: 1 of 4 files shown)


invite/test/github.test.ts
```diff
@@ -48,35 +48,6 @@ describe.skip('GitHub interface', () => {
     }
   });
 
-  describe('userIsMember', () => {
```
danielrozenberg: Why did you remove these tests?

invite/src/github.ts
```diff
@@ -62,8 +55,12 @@ export class GitHub {
     issue_number: number,
     comment: string
   ): Promise<void> {
-    // https://octokit.github.io/rest.js/#octokit-routes-issues-create-comment
-    // octokit.issues.createComment({owner, repo, issue_number, body})
+    await this.client.issues.createComment({
```
danielrozenberg: Why would this bot create comments? (genuine q)

invite/test/github.test.ts
rcebulko: The send-invite API endpoint responds with a field indicating whether the invite was sent (state: pending) or the user was already a member (state: active). So, there's no point to checking if the user is a member, we can just send an invite and GitHub will ignore it if they're a member.

When the thread on limiting invites to org members or group members is resolved, this may be added for that check, but for now it can go :wastebasket:

invite/src/github.ts
rcebulko: To update threads that an invite was sent/accepted, or in the event of an error that the invite was not sent because of an issue. Without some kinda feedback in the form of a comment, there's no indication that the bot actually did anything.

PR #697, opened 2020-02-11
[invite] Add knex database access and relevant tests

rcebulko:
Stole much of the boilerplate from you, thanks!


invite/README.md
```diff
@@ -60,19 +60,28 @@ Setup
    take note of the App ID and app Client Secret.
 6. Install the application on a GitHub repository that you want to use for
    testing. You might want to fork the [ampproject/amphtml](https://github.com/ampproject/amphtml) repository or create a new repository for this purpose.
-7. Copy the `redacted.env` file to `.env` and modify the fields based the values from the GitHub App page:
+7. Run a local instance of PostgreSQL, or use the [Cloud SQL Proxy](https://cloud.google.com/sql/docs/postgres/sql-proxy) to connect to a[Cloud SQL instance](https://pantheon.corp.google.com/sql/choose-instance-engine?project=ampproject-invite-bot)
```
danielrozenberg: nit: space between `a` and `[Cloud`

invite/README.md
```diff
@@ -60,19 +60,28 @@ Setup
...
+9. Copy the `redacted.env` file to `.env` and modify the fields based the values from the GitHub App page:
    * The value for the `APP_ID` field is the App ID from Step 5.
    * The value for `WEBHOOK_SECRET` is the secret you set when creating the GitHub app.
    * The value for the `PRIVATE_KEY` field is a base64 representation of the
      `.pem` file you downloaded from the GitHub App page in Step 5. On Linux/Mac you can convert that file by running `cat private-key-file.pem | base64` in a command line.
    * The value for `GITHUB_ACCESS_TOKEN` is the Client Secret obtained in Step 5.
-   * Set `GITHUB_ORG` to the name of the GitHub organization the bot will be inviting users to.
+   * The value for `GITHUB_ORG` to the name of the GitHub organization the bot will be inviting users to.
+   * The values for `DB_USER`, `DB_PASSWORD`, and `DB_NAME` will be the values used or created during step 7.
+   * The value for `DB_UNIX_SOCKET` will be `/cloudsql/[CLOUD_SQL_INSTANCE_NAME]`; the instance name can be found on the [Instances page](https://pantheon.corp.google.com/sql/instances) and is of the form `project-name:region:database-instance-name`
```
danielrozenberg: Use a database connection string, it's much more dynamic and supports local and remote development workflows

invite/src/setup_db.ts
```diff
@@ -0,0 +1,50 @@
...
+import {Database, dbConnect} from './db';
+
+/** Set up the database table. */
+export function setupDb(db: Database) {
+  return db.schema.createTable('invites', table => {
+    table.increments('id').primary();
+    table.string('username', 40);
+    table.string('repo', 100);
+    table.integer('issue_number');
+    table.enum('action', Object.values(InviteAction));
```
danielrozenberg: This means you'll have to change the schema every time you modify the enum in code. Use an integer and have explicit conversion when you read/write

invite/src/setup_db.ts
```diff
@@ -0,0 +1,50 @@
...
+    table.boolean('archived').defaultTo(false);
```
danielrozenberg: Defaults that are not `null` are okay as fallbacks but make sure the code explicitly sets when you first create a row :)

invite/test/invitation_record.test.ts
```diff
@@ -14,10 +14,15 @@
  * limitations under the License.
  */
 
+import {dbConnect, Database} from '../src/db';
+import {setupDb} from '../src/setup_db';
 import {Invite, InviteAction} from '../src/types';
 import {InvitationRecord} from '../src/invitation_record';
 
+jest.mock('../src/db', () => require('./__mocks__/db'));
```
danielrozenberg: Don't mock, you can inline the in-memory db initialization in the `beforeAll()` and make the behaviour more transparent

invite/test/invitation_record.test.ts
```diff
@@ -27,13 +32,35 @@ describe.skip('invitation record', () => {
   };
   const archivedInvite: Invite = Object.assign({archived: true}, invite);
 
+  beforeAll(async () => setupDb(db));
```
danielrozenberg: Also it's really not such a big deal to create and destroy an in-memory db in `before/afterEach` instead of `before/afterAll` :)

invite/README.md
rcebulko: It doesn't play nice with `knex` when using a socket, which is required for Cloud SQL

invite/test/invitation_record.test.ts
rcebulko: a) This is literally exactly what you did in bundle-size
b) This mock will be useful for e2e tests as well
c) Mocking is not evil

invite/test/invitation_record.test.ts
rcebulko: ¯\_(ツ)_/¯

invite/test/invitation_record.test.ts
danielrozenberg: It's not evil, but it can be confusing and adds; I did that in bundle-size because the implementation was much different at first, and when I changed how the implementation works I guess I never followed up with a refactor. The reason I did the refactor though, was that I _couldn't_ write the test without a mock - which is a sign of tight-coupling

invite/README.md
rcebulko: Done

invite/src/setup_db.ts
rcebulko: The underlying type of the enum is string, so using that.

invite/src/setup_db.ts
rcebulko: Ack

invite/test/invitation_record.test.ts
rcebulko: Done

invite/src/setup_db.ts
danielrozenberg: (something something about being database-engine specific but we'll _never_ switch away from postgres so... meh)

danielrozenberg: (oh you meant you're switching this to use `string`, not `int`, got it)

PR #699, opened 2020-02-11
[invite] Add implementation for invitation record methods

rcebulko:
```
 PASS  test/invitation_record.test.ts
  invitation record
    recordInvite
      ✓ records an invite (5ms)
      ✓ sets `archived = false` (1ms)
    getInvites
      if no invite record exists for the user
        ✓ returns an empty list (2ms)
      if records exists of invites to the user
        ✓ returns the invites (1ms)
      if only archived invite records exists for the user
        ✓ returns an empty list (1ms)
    archiveInvites
      if records exists of any invites to the user
        ✓ updates the invite records (3ms)
```

### DIFF
--- a/invite/src/types.ts
+++ b/invite/src/types.ts
@@ -40,5 +40,5 @@ export interface Invite {
   repo: string;
   issue_number: number;
   action: InviteAction;
-  archived?: boolean|number;  // stored as a TINYINT in PostgresQL
+  archived?: boolean | number; // stored as a TINYINT in PostgresQL
 }
--- a/invite/src/types.ts
+++ b/invite/src/types.ts
@@ -40,5 +40,5 @@ export interface Invite {
   repo: string;
   issue_number: number;
   action: InviteAction;
-  archived?: boolean | number; // stored as a TINYINT in PostgresQL
+  archived?: boolean;
 }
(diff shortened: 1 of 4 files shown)


invite/src/types.ts
```diff
@@ -40,5 +40,5 @@ export interface Invite {
   repo: string;
   issue_number: number;
   action: InviteAction;
-  archived?: boolean;
+  archived?: boolean | number; // stored as a TINYINT in PostgresQL
```
estherkim: This is a little scary! The interface should stay as a boolean but be converted when retrieving or storing SQL data.

rcebulko: I've dropped the `number` type and added explicit casting in the method that actually fetches from the DB.

Thanks for the link! Would it be alright with you if I move this to a declaration file in a separate PR? There are a handful of files that import it directly, and `Database` type could be moved into the type declaration file as well. I'd prefer to do that as a separate set of changes to minimize rebase headaches

rcebulko: https://github.com/ampproject/amp-github-apps/pull/703

estherkim: Sounds good!

rcebulko: Given that, is this PR LGTY?

PR #702, opened 2020-02-12
[invite] Implement core bot actions for invites and assigning issues

rcebulko:
You can ignore all the changes to `test/fixtures` files. I had manually scrubbed the files, replacing a real GitHub username with `someone` and `someone_else`. When I got around to implementing the macro parsing and implemented the regex for usernames, I realized underscores are actually disallowed, so I dropped the underscore everywhere `someone_else` was used.

Note that the bot will comment twice for each invite macro: Once when the macro is triggered (to indicate that it's working/sending the invite) and once when the invite is accepted (to update the thread and possibly note that the issue has been assigned). This may at first seem a bit verbose. My reasoning was that it's important to be able to assure users that the bot is in fact working, and given that the invocations of these macros is likely to be relatively infrequent, any potential "annoyance" the verbosity could cause should be outweighed by its benefits in clarity and understand-ability of the bot's actions. 

```
 PASS  test/invite_bot.test.ts
  Invite Bot
    parseMacros
      ✓ handles mix of `/invite` and `/tryassign` macros (1ms)
      ✓ ignores `whatever/invite` and `something/tryassign` non-macros
      comments without macros
        ✓ returns no matches (23ms)
        ✓ ignores macros without usernames
      comments with `/invite @user`
        ✓ returns list including Invite action for @user (1ms)
        ✓ handles multiple `/invite` macros
      comments with `/tryassign @user`
        ✓ returns list including InviteAndAssign action for @user (1ms)
        ✓ handles multiple `/tryassign` macros
    tryInvite
      when the user has a pending invite from the bot
        ✓ it does not attempt to send an invite (1ms)
        ✓ records the requested invite (1ms)
        ✓ comments that there is already an invite pending (1ms)
      when the user is a member
        ✓ attempts to send an invite
        ✓ does not record the requested invite (1ms)
        ✓ comments that the user is already a member
      when the user is not a member
        ✓ attempts to send an invite (1ms)
        ✓ records the requested invite (5ms)
        ✓ comments that the user was invited (1ms)
      when the invite fails
        ✓ logs an error (1ms)
        ✓ comments about the error
        ✓ re-throws the error (4ms)
    tryAssign
      ✓ assigns the user to the issue
      when @someone just accepted the invitation
        ✓ comments that the issue was assigned (1ms)
      when @someone was already a member of the org
        ✓ comments that the issue was assigned
```


invite/src/invite_bot.ts
```diff
@@ -16,12 +16,24 @@
...
+  invite: InviteAction.INVITE,
+  tryassign: InviteAction.INVITE_AND_ASSIGN,
+};
+
+// Regex source: https://github.com/shinnn/github-username-regex
+const USERNAME_PATTERN: string = '[a-z\\d](?:[a-z\\d]|-(?=[a-z\\d])){0,38}';
+const MACRO_PATTERN: string = `/(${Object.keys(INVITE_MACROS).join('|')})`;
+const FULL_MACRO_REGEX: RegExp = new RegExp(
+  // (?<!\\s) ensures the macro is not preceded by a non-space character.
+  `(?<!\\S)${MACRO_PATTERN} @${USERNAME_PATTERN}`,
```
danielrozenberg: You can use `\b` instead of `(?<!\S)`. e.g., this'll catch "I think Ryan can help here./tryassign @rcebulko" - which is much more likely an unintentional missing space than it is a non-intention to invoke the bot

(if you still prefer to use the negative lookbehind thingy, fix the comment that says `\s` instead of `\S`)

danielrozenberg: Also, consider using match groups instead of string.split down the line

invite/src/invite_bot.ts
```diff
@@ -57,7 +73,17 @@ export class InviteBot {
...
    */
   parseMacros(comment: string): Record<string, InviteAction> {
-    return {};
+    const macros: Record<string, InviteAction> = {};
+    const matches = comment.match(FULL_MACRO_REGEX);
+
+    if (matches) {
+      matches.forEach((macro: string) => {
+        const [macroString, username] = macro.substr(1).split(' @');
+        macros[username] = INVITE_MACROS[macroString.toLowerCase()];
```
danielrozenberg: Isn't macroString here going to be '/invite' instead of 'invite'?

invite/src/invite_bot.ts
```diff
@@ -67,8 +93,76 @@ export class InviteBot {
...
+  /** Handle case where the there was an error sending the invite. */
+  private async handleErrorSendingInvite(invite: Invite, error: Error): Promise<void> {
+    this.logger.error(
+      `Failed to send an invite to \`@${invite.username}\`: ${error}`
+    );
+    await this.github.addComment(
+      invite.repo,
+      invite.issue_number,
+      `You asked me to send an invite to \`@${invite.username}\`, but I ` +
+        'ran into an error when I tried. Try sending the invite manually.'
```
danielrozenberg: How about :%s/Try sending the invite manually./You can try again later ask @ampproject/wg-infra for help./

And because you want to make this repo-agnostic, you can make the helping tag name an env variable :)

invite/src/invite_bot.ts
```diff
@@ -78,5 +172,20 @@ export class InviteBot {
    * Should be called in response to accepted invitations in order to update the
    * thread(s) from which the user was invited.
    */
-  async tryAssign(invite: Invite, accepted: boolean): Promise<void> {}
+  async tryAssign(invite: Invite, accepted: boolean): Promise<void> {
+    const commentStart = accepted ?
+      `The invitation to \`@${invite.username}\` was accepted!` :
+      `It looks like \`@${invite.username}\` is already a member of ` +
```
danielrozenberg: If the invitee is already a member skip this prefix and have the whole comment just say "I've assigned this issue to @username", no need to say that they're already a member in that case

invite/src/invite_bot.ts
rcebulko: `\b` actually matches anything that precedes `/`, since it is considered a word break. I fiddled with this for a while lol. I could explicitly add punctiation I suppose, but my concern was that there's also the case where suppose we're commenting on this repo and we said "I'm planning to move ./*.js to ./invite @danielrozenberg" . ie, would rather make it not trigger on a typo than accidentally trigger on some random comment. WDYT?

rcebulko: Considered it, doesn't play nice with Typescript. Match doesn't work with capture groups when using the global flag. I would like to use `matchAll, but `String.prototype.matchAll` is relatively new, and Typescript doesn't recognize that it exists. I tried for about 20 minutes to adjust the TS config so it would work, but none of the fixes I found online worked, so I just gave up. At least the fact that it matches the regex ensures that by the time it reaches the string.split, it's in the expected format.

invite/src/invite_bot.ts
rcebulko: `.substr(1)`

invite/src/invite_bot.ts
rcebulko: Done.

invite/src/invite_bot.ts
danielrozenberg: Fair

invite/src/invite_bot.ts
danielrozenberg: lol

invite/src/invite_bot.ts
danielrozenberg: Add a TODO then :)

invite/src/invite_bot.ts
rcebulko: How about "You can try sending the invite manually, or ask `@ampproject/wg-infra` for help."

rcebulko: Done; resolving.

danielrozenberg: sg :P

Issue #687, opened 2020-02-07
Create a new automated workflow to invite new contributors to `ampproject/amphtml`


rsimha:
For context, see https://github.com/ampproject/meta-tsc/issues/28

> The TSC discussed how to balance efficiency for assignments against not being spammy. It seems to come down to not wanting to be blocked by lack of org membership when wanting to perform the assignment. So, can we directly confront this issue?
> 
> What we'd like to consider is a bot that detects a comment like "/tryassign @example". If @example is not yet in the AMP org, they get an invite. If they accept the invitation, the bot can assign the issue/PR to them. This could occur on a later pass through issues, on the event of the acceptance, or through any other method.

/cc @ampproject/wg-infra

rcebulko:
Given the nature of this goal, it would be a good candidate for [serverless deployment](https://probot.github.io/docs/serverless-deployment/) with Google Cloud Functions.

## Interface

Building off the interface suggested by @davidstrauss  in https://github.com/ampproject/meta-tsc/issues/28#issuecomment-567227934, a stateless Probot app can listen for the [issue comment](https://developer.github.com/v3/activity/events/types/#issuecommentevent) webhook event (and optionally [commit comments](https://developer.github.com/v3/activity/events/types/#commitcommentevent)) and look for:
- `/invite @user`: this will trigger the app sending an [invite to the organization](https://developer.github.com/v3/orgs/members/#add-or-update-organization-membership) (**not** just the repository) and posting a comment indicating the invite was sent
  - by also listening for [member events](https://developer.github.com/v3/activity/events/types/#memberevent), the bot can post another comment updating the thread when the user accepts the invitation
- `/tryassign @user`: this will do the same as above (ie. send an invite); if the user accepts the invite, it will not only comment, but will also try to assign the issue/PR to the user

## Questions to answer

1. If the issue/PR is already assigned to someone else, should the `/tryassign` user be assigned in their place, or should `/tryassign` only assign an issue if there is no assignee yet?

Proposal: Only assign if unassigned. Consider the case where a user is `/tryassign`-ed in an issue, but does not accept the invite right away. Other devs take over the issue. Months later, the user accepts the invite that has been sitting in her inbox; she should not be assigned the issue.

2. Should there be any restriction on who can trigger the `/invite` and `/tryassign` macros?

[...]

PR #703, opened 2020-02-12
[invite] Migrate module types to a type declaration file

rcebulko:
Addresses #687 and https://github.com/ampproject/meta-tsc/issues/28

### DIFF
--- a/invite/src/db.ts
+++ b/invite/src/db.ts
@@ -17,9 +17,8 @@ require('dotenv').config();
 
 import Knex from 'knex';
 
-export type Database = Knex;
 /** Connect to a database instance. */
-export function dbConnect(): Database {
+export function dbConnect(): Knex {
   return Knex({
     client: 'pg',
     connection: {
--- a/invite/src/invitation_record.ts
+++ b/invite/src/invitation_record.ts
@@ -14,7 +14,8 @@
  * limitations under the License.
  */
 
-import {Database} from './db';
+import Knex from 'knex';
+
 import {ILogger, Invite} from './types';
 
 /**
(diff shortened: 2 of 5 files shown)


invite/src/db.ts
```diff
@@ -15,9 +15,10 @@
 
 require('dotenv').config();
 
+import {Database} from '.';
```
estherkim: What is this importing?

rcebulko: The type

invite/src/index.d.ts
```diff
@@ -14,27 +14,26 @@
  * limitations under the License.
  */
 
-/**
- * A standard logging interface.
- */
+import Knex from 'knex';
+
+/** A Knex database interface. */
+export type Database = Knex;
```
estherkim: knex provides its own type defs, no? This shouldn't need to be imported

rcebulko: I'm trying to alias `Knex` to `Database` so that the other files don't have to explicitly know about `Knex`, but I'm not familiar enough with Typescript to know the right way to do that or if it's even the right thing to do. WDYT?

estherkim: Oh hmm why do you want to alias it? It could make debugging harder.

But if you have a database service, you could have it implement an interface and have other files reference that instead of knex. Looks like that's what `invitation_record.ts` essentially is. (OT but I'd rename that class)

rcebulko: I dropped the alias entirely, and just use Knex as the type now. Using this method with the type declarations, I get no issues compiling the Typescript with `tsc`, but for some reason it fails when I try to actually run the tests. I'm not really sure what's going wrong :disappointed:

rcebulko:
@estherkim I tried out the branch you had, but after rebasing and getting everything else working, there were still issues with actually running the code. I think it may be because imports of `.d.ts` files are ignored at runtime, but the enum type actually defines string values which are used at runtime. I've updated this PR to just drop the Knex alias for now, and the modules still import from `src/types.ts`

rcebulko:
Closing this, as the relevant work was done in other linting/cleanup PRs.

PR #705, opened 2020-02-13
[invite] Add team membership check to GitHub API interface

rcebulko:
Will be used to validate that the comment author is in `ampproject/reviewers-amphtml` (similar to Owners Bot) in order to trigger a macro.

Addresses #687

### DIFF
--- a/invite/app.ts
+++ b/invite/app.ts
@@ -26,6 +26,7 @@ export default (app: Application) => {
     require('dotenv').config();
   }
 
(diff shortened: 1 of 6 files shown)


rcebulko:
Per https://github.com/ampproject/meta-tsc/issues/28#issuecomment-585868989 we're going to start by checking for membership in `ampproject/reviewers-amphtml` instead and open it up from there if needed later on. Closing this PR and will implement team membership check separately.

rcebulko:
Updated to use team membership instead of org membership; re-opened.

PR #706, opened 2020-02-13
[invite] Implement methods to process comments and accepted invites

rcebulko:
View commit-by-commit. First couple commits are cleaning up rebase messiness. The rest implements the two methods to actually process a new comment or an accepted invite using the already-defined `parseMacros`, `tryInvite`, and `tryAssign` methods.

With this PR, all end-to-end tests now pass.

```
 PASS  test/invite_bot.test.ts
  Invite Bot
    processComment
      ✓ parses the comment for macros (3ms)
      when macros are present
        ✓ tries to send invites (4ms)
        for /tryassign macros
          ✓ tries to assign the issue (2ms)
      when no macros are found
        ✓ does not try to send any invites (1ms)
    processAcceptedInvite
      ✓ checks the record for invites to the user (4ms)
      when there are recorded invites
        with Invite action
          ✓ does not try to assign any issues (3ms)
          ✓ comments on the issues that the invite was accepted (3ms)
        with InviteAndAssign action
          ✓ tries to assign the issues (4ms)
      when there are no recorded invites to the user
        ✓ does not try to assign any issues (2ms)
        ✓ does not comment on any issues (2ms)

 PASS  test/e2e.test.ts
  end-to-end
    when a comment includes "/invite @someone"
      when @someone is a member of the org
        ✓ comments, doesn't record (74ms)
      when @someone is not a member of the org
        ✓ invites, records, comments (8ms)
        once the invite is accepted
          ✓ comments, archives (7ms)
    when a comment includes "/tryassign @someone"
      when @someone is a member of the org
        ✓ assigns, comments, doesn't record (7ms)
      when @someone is not a member of the org
        ✓ invites, records, comments (7ms)
        once the invite is accepted
          ✓ assigns, comments, archives (8ms)
    when a comment includes no macros
      ✓ ignores it (2ms)
    when someone joins without a recorded invitation
      ✓ ignores it (3ms)
```

Addresses #687


invite/test/e2e.test.ts
```diff
@@ -63,14 +63,15 @@ describe('end-to-end', () => {
     db.destroy();
   });
 
+  beforeEach(() => nock.cleanAll());
```
rcebulko: If the `cleanAll` call lives in the `afterEach` block, there are cases where a test can fail and the error throws itself out of the test before the `cleanAll` call executes, which leaves un-requested nocks in place and fails all tests that follow. Doing it before each test is more reliable in the event of an error within a test.

invite/test/e2e.test.ts
```diff
@@ -91,7 +100,7 @@ describe('end-to-end', () => {
           .reply(200);
 
         await triggerWebhook(probot, 'trigger_invite.issue_comment.created');
-        expect(await db('invites').select().first()).toBeUndefined();
+        expect(await record.getInvites('someone')).toEqual([]);
```
rcebulko: Using the `InvitationRecord` interface here simplifies the checks, since that class handles some typecasting nonsense from Postgres

invite/src/invite_bot.ts
```diff
@@ -115,16 +115,23 @@ export class InviteBot {
     // The user has not been invited by the bot yet, so try and send an invite.
     let invited = false;
     try {
-      invited = await this.github.inviteUser(invite.username);
+      invited = await this.github.inviteUser(invite.username).catch(error => {
```
danielrozenberg: `await`ing a `promise` that rejects will simply throw that rejection as an `Error`, so basically what you wrote here is:
```typescript
try {
  try {
    invited = await this.github.inviteUser(invite.username);
  } catch (error) {
    throw error;
  }
} catch (error) {
  await this.handleError...
```

(TL;DR: don't add the `.catch(...)` part)

invite/test/invite_bot.test.ts
```diff
@@ -14,33 +14,54 @@
...
 import {mocked} from 'ts-jest/utils';
 
+import {setupDb} from '../src/setup_db';
+import {dbConnect} from '../src/db';
 import {Invite, InviteAction} from '../src/types';
 import {InvitationRecord} from '../src/invitation_record';
 import {InviteBot} from '../src/invite_bot';
 import {GitHub} from '../src/github';
 
+jest.mock('../src/db', () => {
```
danielrozenberg: Again, not really needed for unit tests. Makes sense in e2e tests though

invite/src/invite_bot.ts
```diff
@@ -73,12 +73,34 @@ export class InviteBot {
...
 
   /**
    * Process an accepted invite by adding comments and assigning issues.
    */
-  async processAcceptedInvite(username: string): Promise<void> {}
+  async processAcceptedInvite(username: string): Promise<void> {
+    const recordedInvites = await this.record.getInvites(username);
+
+    for (const invite of recordedInvites) {
+      if (invite.action === InviteAction.INVITE) {
```
danielrozenberg: Consider using `switch` here, with `default: throw new RangeError('Unimplemented action');` (it's more future-proof)

invite/test/e2e.test.ts
danielrozenberg: A cleaner solution is to put this in the `afterEach`:
```typescript
// My preferred syntax
try {
  if (!nock.isDone()) {
    throw new Error('Not all nock interceptors were used!');
  }
} finally {
  nock.cleanAll();
}
```

or
```typescript
// Arguably more straightforward, arguably more confusing syntax
const nockWasDone = nock.isDone();
nock.cleanAll();
if (!nockWasDone) {
  throw...
}
```

(here and in the other test file)
This way you don't have to double-clean the nocks :)

invite/test/e2e.test.ts
```diff
@@ -94,7 +95,7 @@ describe('end-to-end', () => {
         done();
       });
     });
-
```
danielrozenberg: 😱

invite/test/e2e.test.ts
danielrozenberg: Nit: I prefer using `expect(foo).resolves.bar` instead of `expect(await foo).bar`:
```typescript
expect(record.getInvites('someone')).resolves.toBe([]);
```

(yes, I am aware my old tests aren't using this. I learned about this after writing them 🙃)

danielrozenberg: > Using the `InvitationRecord` interface here simplifies the checks, since that class handles some typecasting nonsense from Postgres

+1

invite/test/e2e.test.ts
```diff
@@ -113,22 +122,20 @@ describe('end-to-end', () => {
             expect(body).toEqual({
               body: 'An invitation to join `test_org` has been sent to ' +
```
danielrozenberg: Supernit: you don't always have to break long strings into multiple lines, especially if they're simple strings (i.e., not formatter strings) - this whole thing can be in one line and still pass Prettier formatting. Totally your choice of style

invite/test/e2e.test.ts
```diff
@@ -113,22 +122,20 @@ describe('end-to-end', () => {
...
+                'is accepted.'
             });
             return true;
           })
           .reply(200);
 
         await triggerWebhook(probot, 'trigger_invite.issue_comment.created');
-        expect(await db('invites').select().first()).toEqual(recordedInvite);
+        const invites = await record.getInvites('someone');
+        expect(invites[0]).toMatchObject(recordedInvite);
```
danielrozenberg:
```typescript
expect(record.getInvites('someone')).resolves.arrayContaining(recordInvite);
```
(this way if invites is zero length or undefined or not an array, the test won't crash here)

invite/test/e2e.test.ts
```diff
@@ -233,17 +241,21 @@ describe('end-to-end', () => {
...
       done();
     });
   });
 
   describe('when someone joins without a recorded invitation', () => {
     it('ignores it', async done => {
       await triggerWebhook(probot, 'organization.member_added');
-      expect(await db('invites').select().first()).toBeUndefined();
+
+      // The test will fail if any unexpected network requests occur.
```
danielrozenberg: Supernit: could still make this explicit in the code by adding something like
```typescript
expect(nock.isDone()).toBe(true);
```
(since this is _the_ thing you want to assert in this test)

invite/test/invite_bot.test.ts
rcebulko: I think it's necessary because `InviteBot` calls `dbConnect` in its constructor

invite/test/e2e.test.ts
rcebulko: I often have 3~4 vertical code panes, so I like keeping lines as narrow as the style guide will justify :P

invite/test/e2e.test.ts
rcebulko: 1) it would have to be `arrayContaining(expect.matchesObject(recordedInvite))` to ignore the extra `id` field
2) it should crash if the array is empty though :thinking:

invite/test/e2e.test.ts
rcebulko: That's not the thing. There aren't any nocks that I want to assert are done; I want to be sure that there aren't any _un-nocked requests_, which nock already throws an error on.

invite/src/invite_bot.ts
rcebulko: This was the product of a deep rabbit hole of trying to mock, then catch, then assert errors/rejections in test lol. Fixed.

invite/src/invite_bot.ts
rcebulko: I can't write a test for it (because Typescript won't let me pass a non-existent enum value), but done.

invite/test/e2e.test.ts
rcebulko: Modified all tests (in `invite_bot` and elsewhere) to use that form.

invite/test/e2e.test.ts
danielrozenberg: Crash ≠ fail >_<
(they result in the same thing, but you should try to write safe code wherever you can)

rcebulko: Updated it to

```
expect(record.getInvites('someone')).resolves.toEqual([
  expect.objectContaining(recordedInvite)
]);
```

invite/test/e2e.test.ts
rcebulko: I tried that and it didn't work. I fiddled with this for a while, Jest seems to be picky with nocks and errors. It's not double-cleaning the nocks, it's just cleaning before each test instead of after, which seems fine to me, so ¯\\_(ツ)_/¯

PR #707, opened 2020-02-13
[invite] Add check to see if author is allowed to trigger macros

rcebulko:
Adds bot logic around testing if the user who commented is allowed to trigger an invite. Currently it always returns true, but after this and #705 we'll be able to restrict it to members of `ampproject` in a follow-up PR.

Addresses #687


rcebulko:
Whoops, forgot this was dependent on #706 . Closing until that is submitted and this branch is rebased.

rcebulko:
This now asserts that the comment author is a member of the team defined in the `ALLOW_TEAM_SLUG` environment variable. The PR adds the following tests:

```
 PASS  test/e2e.test.ts
  end-to-end
    when the author is not a member of the allow team
      ✓ ignores it (1ms)

 FAIL  test/invite_bot.test.ts
  Invite Bot
    userCanTrigger
      ✓ returns true if user is a member of allow team (1ms)
      ✓ returns false if user is not a member of allow team (3ms)
```

All e2e tests are updated to include the nocked membership test.

invite/src/invite_bot.ts
```diff
@@ -51,10 +51,16 @@ export class InviteBot {
    * event that the invite fails to send for some reason. If left as null, the
    * comment will just say "ask someone for help".
    * Example: 'ampproject/wg-infra'
+   *
```
danielrozenberg: Order the comments on the arguments based on the same order of the arguments :)

invite/src/invite_bot.ts
```diff
@@ -51,10 +51,16 @@ export class InviteBot {
...
+   * The allowTeamSlug parameter restricts the set of users who can trigger an
+   * invite to members of a GitHub team. The value should be the name of the 
+   * team, but should not include the org.
+   * Example: 'wg-infra'
    */
   constructor(
     client: Octokit,
     private org: string,
+    private allowTeamSlug: string,
```
danielrozenberg: Might be worth making this an array or strings, I can 100% see this becoming a feature request ("oh can you also let the TSC and AC invite people?")

invite/src/invite_bot.ts
```diff
@@ -51,10 +51,16 @@ export class InviteBot {
    * event that the invite fails to send for some reason. If left as null, the
    * comment will just say "ask someone for help".
    * Example: 'ampproject/wg-infra'
+   *
+   * The allowTeamSlug parameter restricts the set of users who can trigger an
+   * invite to members of a GitHub team. The value should be the name of the 
+   * team, but should not include the org.
+   * Example: 'wg-infra'
```
danielrozenberg: It's a bit confusing that `allowTeamSlug` requires just the team name, and `helpUsernameOrTag` requires an org/teamname - one of these should change 🙃

invite/test/invite_bot.test.ts
```diff
@@ -66,32 +67,33 @@ describe('Invite Bot', () => {
...
+        /*helpUsernameToTag=*/'test_org/wg-helpme',
       );
-      expect(inviteBot.helpUserTag).toEqual('@test_org/wg-example');
+      expect(inviteBot.helpUserTag).toEqual('@test_org/wg-helpme');
     });
   });
 
   describe('processComment', () => {
     beforeEach(() => {
+      jest.spyOn(inviteBot, 'userCanTrigger');
```
danielrozenberg: Should've noticed this in the previous round of reviews... As a rule of thumb don't spy on or mock things that are inside the class/module that you're testing - spy or mock _only_ its dependencies. You can still `expect(method()).toBe / .resolves.toBe`, but instead of `expect(method()).toBeCalledWith / .not.toBeCalled` you'll have to replace this with its dependencies.

When you spy on your own methods you create a tight coupling between the code under test, and the test itself. When you mock your own methods you are testing that your test is testing itself correctly. When you mock the dependencies you create a tight coupling between the test and the code's dependencies, which is better because ideally you use stable dependencies which themselves shouldn't change :)

(there is, of course, no perfect way to decouple the test from the code entirely, but the less friction between the test and the code that is being tested, the better)

invite/src/invite_bot.ts
rcebulko: If they do that, I'll push back and ask them to create a Team which includes subteams (nesting is allowed). More sensible than doing multiple API checks, and easier to reason about/keep track of a single `ampproject/inviters` team

invite/test/invite_bot.test.ts
rcebulko: This was written before I'd implemented the underlying functionality; I'll swap it out for a spy on the GitHub instance.

invite/src/invite_bot.ts
danielrozenberg: Teams _aren't_ cheap, their listings require constant maintenance - this pushback means you expect Joey to keep adding people to multiple teams all the time. I'll accept the pushback, but urge you to reconsider

invite/src/invite_bot.ts
rcebulko: Done.

invite/src/invite_bot.ts
rcebulko: Done.

invite/test/invite_bot.test.ts
rcebulko: Done.

invite/src/invite_bot.ts
rcebulko: No, I mean groups can own groups. So if he wants `tsc-meta` and `wg-infra` and `wg-approvers` to all have invite powers, he can add the _groups themselves_ to something like `org-inviters`. Then when someone gets added to `wg-infra`, that automatically includes them in `org-inviters`. Do you get what I mean with that?

danielrozenberg: Yep :) alright

PR #712, opened 2020-02-18
[invite] Add helper to build invite-bot instance from Probot context

rcebulko:
The repetitive (long) constructor calls were annoying me, especially with the Probot/Octokit typecast :stuck_out_tongue: 

Addresses https://github.com/ampproject/amp-github-apps/issues/687 and https://github.com/ampproject/meta-tsc/issues/28

### DIFF
--- a/invite/app.ts
+++ b/invite/app.ts
@@ -20,6 +20,13 @@ import {Octokit} from '@octokit/rest';
 
 import {InviteBot} from './src/invite_bot';
 
+type CommentWebhookPayload =
+  | Webhooks.WebhookPayloadIssueComment
+  | Webhooks.WebhookPayloadIssues
+  | Webhooks.WebhookPayloadPullRequest
+  | Webhooks.WebhookPayloadPullRequestReview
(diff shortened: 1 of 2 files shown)


invite/app.ts
```diff
@@ -20,100 +20,86 @@ import {Octokit} from '@octokit/rest';
...
-    {github, event, payload, log}: Context<Webhooks.WebhookPayloadIssueComment>
-  ) => {
-    const inviteBot = new InviteBot(
+  function botFromContext(
+    {github, payload, log}: Context<CommentWebhookPayload>
+  ) {
+    return new InviteBot(
+      // This type-cast is required because Probot exports a separate GitHubAPI
+      // class, even though it's in Octokit instance.
       (github as unknown) as Octokit,
```
danielrozenberg: Did you consider changing the original constructor to take a Context<CommentWebhookPayload> instead of an Octokit? :P

rcebulko: I'd prefer not to tie the InviteBot constructor to Probot

danielrozenberg: Fair 🤷‍♂

PR #713, opened 2020-02-19
Fix cherry-pick issue scraping DB query

rcebulko:
Fixes a query that SQLAlchemy was complaining about in the logs


project-metrics/metrics_service/database/models.py
```diff
@@ -196,7 +196,7 @@ def __repr__(self) -> Text:
 
 
 class Cherrypick(Base):
-  """A cherry-picked commit."""
+  """[DEPRECATED] A cherry-picked commit."""
```
danielrozenberg: Add an explanation why, and what replaces it

rcebulko: Done

PR #714, opened 2020-02-21
Report parser error when encountering singleton brace expansion rule patterns

rcebulko:
The minimatch glob parsing library Owners Bot uses for parsing patterns performs brace-expansion, but only if the braces contain comma-separated patterns. This is not obvious behavior, and has resulted in broken owners rules (see https://github.com/ampproject/amphtml/pull/26899).

Ex. `"{a,b}.js"` becomes `/(a|b)\.js/`
Ex. `"{a}.js"` becomes `/\{a\}\.js/`

This PR identifies rules where there is a single pattern specified within a brace-expansion and rejects the rule, reporting an error. This will block future PRs from submitting rules broken in this way.

/cc @Enriqe

### DIFF
--- a/owners/src/ownership/parser.js
+++ b/owners/src/ownership/parser.js
@@ -261,6 +261,23 @@ class OwnersParser {
       }
     }
 
+    // Check that there are no singleton patterns in brace-expansion format. The
+    // minimatch glob parsing library performs brace-exparnsion, but only if the
+    // braces contain comma-separated patterns. This is not obvious behavior,
+    // and has resulted in broken owners rules.
+    // Ex. `{a,b}.js` becomes /(a|b)\.js/
+    // Ex. `{a}.js` matches /\{a\}\.js/
+    if (/\{[^,]+\}/.test(pattern)) {
+      errors.push(
+        new OwnersParserError(
+          ownersPath,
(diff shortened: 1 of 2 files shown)


owners/src/ownership/parser.js
```diff
@@ -261,6 +261,23 @@ class OwnersParser {
       }
     }
 
+    // Check that there are no singleton patterns in brace-expansion format. The
+    // minimatch glob parsing library performs brace-expansion, but only if the
+    // braces contain comma-separated patterns. This is not obvious behavior,
+    // and has resulted in broken owners rules.
+    // Ex. `{a,b}.js` becomes /(a|b)\.js/
+    // Ex. `{a}.js` becomes /\{a\}\.js/
+    if (/\{[^,]+\}/.test(pattern)) {
```
danielrozenberg: New edge-case introduced in this PR: what if the file actually includes `{` and `}` in its name? 😅

rcebulko: Then it is wrong :laughing:

rcebulko: But actually that should be disallowed and I'm comfortable with not supporting that at present lmao

PR #716, opened 2020-02-25
[invite] Fix error being logged for empty comments

rcebulko:
When the comment field is empty, GitHub sets it to `null`, which results in an error when trying to parse macros. This has no impact except logging errors since empty comments obviously contain no macros. This PR aborts trying to process a comment if it's empty.


rcebulko:
test comment

Issue #722, opened 2020-02-27
[release calendar] Create server layer


estherkim:
- [ ] Start a database initialization script
- [ ] For development, use a local database

PR #732, opened 2020-03-02
[release-calendar] Connect to a MySQL db with TypeORM

estherkim:
Uses [TypeORM](https://github.com/typeorm/typeorm) to interface with a MySQL database. Opted for this over knex for its TypeScript support.

Uses a `.env` file. Updates README for instructions on starting local dev.

Related to https://github.com/ampproject/amp-github-apps/issues/722

### DIFF
--- a/release-calendar/README.md
+++ b/release-calendar/README.md
@@ -8,3 +8,7 @@ To build the app: `npm run build`
 To run the app: `npm run start`
 To run tests: `npm run test`
 To lint: `npm run lint`
+
+#### Local development
+1. Install MySQL for a local database.
+2. Add a `.env` file to the root directory. See `.env.example` for the required variables and replace values with your own.
--- a/release-calendar/src/app.ts
+++ b/release-calendar/src/app.ts
@@ -14,4 +14,31 @@
  * limitations under the License.
  */
 
-console.log('hi');
+import {createConnection, Connection} from 'typeorm';
+import {Release} from './entities/release';
+
+async function main() {
+  const connection = await createConnection({
+    type: 'mysql',
+    host: process.env.HOST,
+    username: process.env.DB_USERNAME,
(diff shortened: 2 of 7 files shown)


release-calendar/README.md
```diff
@@ -7,4 +7,8 @@ To start: `cd release-calendar && npm install`
 To build the app: `npm run build`
 To run the app: `npm run start`
 To run tests: `npm run test`
-To lint: `npm run lint`
\ No newline at end of file
+To lint: `npm run lint`
+
+#### Local development
+1. Install MySQL for a local database.
```
danielrozenberg: Note that we use PostgreSQL in our other apps, so it might be worth switching to that as the default db engine for consistency

release-calendar/src/app.ts
```diff
@@ -14,4 +14,31 @@
...
+
+async function main() {
+  const connection = await createConnection({
+    type: 'mysql',
+    host: process.env.HOST,
+    username: process.env.DB_USERNAME,
+    password: process.env.DB_PASSWORD,
+    database: process.env.DB_NAME,
+    entities: [Release],
+    synchronize: true, //
```
danielrozenberg: //?

release-calendar/src/app.ts
```diff
@@ -14,4 +14,31 @@
...
+    logging: false,
+  }).catch(error => {
+    throw error;
+  }) as Connection;
+
+
+  const releaseRepo = connection.getRepository(Release);
+
+  const release = new Release('1234567891234');
```
danielrozenberg: Debugging leftovers?

release-calendar/README.md
estherkim: I considered it, but MySQL seemed more supported and popular. We won't notice the difference too much but I am using TypeORM instead of knex, which looks very different.

PR #738, opened 2020-03-06
[checklist] Boilerplate

### LINKED ISSUE #735: Automatically post checklist comment for new components
When a PR includes a new folder in `extensions/`, have @ampprojectbot automatically post a checklist:

> It looks like you're submitting a new extension.  👁️📎👁️
> 
> Please make sure you're following the [contribution guidelines](https://go.amp.dev/contribute/code). These steps need to be completed before code approval:
> 
> - [ ] This component has an [intent-to-implement issue](https://go.amp.dev/i2i) that is attached to this pull request. It documents:
>   - [ ] **component features**
>   - [ ] **element attributes**
>   - [ ] **how the page is modified** ("my component creates an `<iframe>` that loads `$URL`").
> - [ ] This I2I has been approved during [a design review](https://github.com/ampproject/amphtml/blob/master/contributing/design-reviews.md).
>
> If your component is an ad or an integration for analytics, please tag `@ampproject/wg-ads-and-analytics`. Otherwise tag `@ampproject/wg-ui-and-a11y`.

/to @ampproject/wg-infra

alanorozco:
Partial for #735


checklist/cloud_build.yaml
```diff
@@ -0,0 +1,60 @@
...
+
+  # Deploy the app.
+  - name: gcr.io/cloud-builders/gcloud
+    dir: checklist
+    args:
+      - functions
+      - deploy
+      - --source
+      - dist/
+      - checklist-bot
```
alanorozco: Unsure about these names or this GCP deployment process.

checklist/README.md
```diff
@@ -0,0 +1,71 @@
+# AMP Contribution Checklist Bot
+
+A GitHub app which posts checklists on pull requests.
```
rcebulko: Nit: add some context/what the checklist is

checklist/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+This app runs as a Google Cloud Functions deployment. It can be configured for deployment to any organization.
+
+## Interface
+
+The AMP Contribution Checklist Bot watches for all newly opened pull requests.
+
+If the pull request includes a new directory in `extensions/`, the bot will:
+
+1. edit the pull request's description to insert a checklist
```
rcebulko:
```suggestion
1. Edit the pull request's description to insert a checklist
```

checklist/README.md
```diff
@@ -0,0 +1,71 @@
...
+2. post a comment on the pull request metioning the added checklist
```
rcebulko:
```suggestion
2. Post a comment on the pull request metioning the added checklist
```

checklist/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Webhooks
+
+The app subscribes to the following GitHub Webhooks:
+
+- [`PullRequestEvent`](https://developer.github.com/v3/activity/events/types/#pullrequestevent)
+  - `opened`: check the PR content and its comments to post checklist
```
rcebulko: How do you want to handle draft PRs or PRs with "DO NOT SUBMIT" or similar things in the title? This is a common question we end up confronting after-the-fact with PR bots, worth answering here. If you want to ignore draft PRs and only comment once the PR is ready, you can filter on `payload.pull_request.draft` or something like that, and then you can listen for the `pull_requests.ready_for_review` event. The Owners Bot `index.js` has some logic like this you may find helpful.

checklist/README.md
```diff
@@ -0,0 +1,71 @@
...
+5. After creating the application, generate and download a private key. Also
+   take note of the App ID and app Client Secret.
+6. Install the application on a GitHub repository that you want to use for
+   testing. You might want to fork the [ampproject/amphtml](https://github.com/ampproject/amphtml) repository or create a new repository for this purpose.
+7. Copy the `redacted.env` file to `.env` and modify the fields based the values from the GitHub App page:
+   - The value for the `APP_ID` field is the App ID from Step 5.
+   - The value for `WEBHOOK_SECRET` is the secret you set when creating the GitHub app.
+   - The value for the `PRIVATE_KEY` field is a base64 representation of the
+     `.pem` file you downloaded from the GitHub App page in Step 5. On Linux/Mac you can convert that file by running `cat private-key-file.pem | base64` in a command line.
+   - The value for `GITHUB_ORG` to the name of the GitHub organization the bot will be inviting users to.
```
rcebulko: This line is from invite-bot, and isn't relevant here. You'll probably want `GITHUB_ORG` and `GITHUB_REPO`

alanorozco:
@rcebulko Suggestions on how to split concretely? Most of this is bot boilerplate.

checklist/app.ts
```diff
@@ -0,0 +1,51 @@
...
+      // This type-cast is required because Probot exports a separate GitHubAPI
+      // class, even though it's in Octokit instance.
+      (github as unknown) as Octokit,
+      payload.repository.owner.login,
+      log
+    );
+  }
+
+  app.on(
+    'pull_request.opened',
```
rcebulko: See other comments on draft PRs

checklist/cloud_build.yaml
rcebulko: https://github.com/ampproject/amp-github-apps/blob/master/DEPLOYMENT.md

rcebulko: The name here is the name of the Cloud Function, which can be whatever as long as it stays constant. This should be fine. See above for guidance on automating the deployment with Cloud Build and for guidance on setting up the shared keyring/encrypted secrets.

checklist/package.json
```diff
@@ -0,0 +1,74 @@
+{
+  "name": "amp-checklist-bot",
```
rcebulko: High-level nit: let's see if we can figure out a clearer name

checklist/package.json
```diff
@@ -0,0 +1,74 @@
...
+    "node": "^12.13.1"
+  },
+  "scripts": {
+    "lint": "eslint **/*.ts",
+    "fix": "eslint **/*.ts --fix",
+    "lint-conflicts": "eslint --print-config app.ts | eslint-config-prettier-check",
+    "build": "tsc",
+    "build:watch": "tsc -w --p tsconfig.json",
+    "start": "probot run ./dist/app.js",
+    "dev": "nodemon",
```
rcebulko: Does this work if you try to run it? :stuck_out_tongue:

checklist/package.json
```diff
@@ -0,0 +1,74 @@
...
+    "deploy-tag": "git tag 'deploy-checklist-'`date --utc '+%Y%m%d%H%M%S'`",
```
rcebulko: This will be relevant to when you go through https://github.com/ampproject/amp-github-apps/blob/master/DEPLOYMENT.md

checklist/package.json
```diff
@@ -0,0 +1,74 @@
...
+    "@typescript-eslint/eslint-plugin": "2.19.2",
+    "@typescript-eslint/parser": "2.19.2",
+    "eslint": "6.8.0",
+    "eslint-config-prettier": "6.10.0",
+    "eslint-plugin-prettier": "3.1.2",
+    "jest": "25.1.0",
+    "nock": "11.7.2",
+    "nodemon": "2.0.2",
+    "prettier": "1.19.1",
+    "sqlite3": "4.1.1",
```
rcebulko: This shouldn't need sqlite :)

checklist/app.ts
```diff
@@ -0,0 +1,51 @@
...
+
+type WebhookPayload = Webhooks.WebhookPayloadPullRequest;
+
+module.exports = (app: Application) => {
+  if (process.env.NODE_ENV !== 'test') {
+    require('dotenv').config();
+  }
+
+  function botFromContext({github, payload, log}: Context<WebhookPayload>) {
+    return new ChecklistBot(
```
danielrozenberg: Ryan created this because he needed to call this method multiple times. No reason why you can't just inline this in the single `app.on` handler that you have here :)

checklist/src/checklist_bot.ts
```diff
@@ -0,0 +1,76 @@
...
+  readonly github: GitHub;
+
+  /** */
+  constructor(
+    client: Octokit,
+    private org: string,
+    private logger: ILogger = console
+  ) {
+    this.github = new GitHub(client, org, logger);
+    this.logger.info(`ChecklistBot initialized for ${this.org}`);
```
rcebulko: Nit: This will probably be a repo-bot rather than an org-wide bot. Can address in a later stage of review.

checklist/src/checklist_bot.ts
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  /**
+   * Process a pull request.
+   * Determines whether a new directory was added to path. If so, it insers a
```
rcebulko:
```suggestion
   * Determines whether a new directory was added to path. If so, it inserts a
```

checklist/app.ts
```diff
@@ -0,0 +1,51 @@
...
+import Webhooks from '@octokit/webhooks';
+import {Octokit} from '@octokit/rest';
+
+import {ChecklistBot} from './src/checklist_bot';
+
+type WebhookPayload = Webhooks.WebhookPayloadPullRequest;
+
+module.exports = (app: Application) => {
+  if (process.env.NODE_ENV !== 'test') {
+    require('dotenv').config();
```
danielrozenberg: I'm pretty sure Probot does this whole thing automatically, you can try removing this condition and see if local dev and tests still work as expected

checklist/src/checklist_bot.ts
```diff
@@ -0,0 +1,76 @@
...
+   * checklist into the pull's description and posts a notifier comment pointing
+   * to it.
+   */
+  async processPullRequest(
+    repo: string,
+    pull_number: number,
```
rcebulko: nit: my preference is to use camelCase rather than let the format of the JSON payload dictate the variable names in our JS

checklist/src/checklist_bot.ts
danielrozenberg: "inserts"

checklist/src/github.ts
```diff
@@ -0,0 +1,103 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {Octokit} from '@octokit/rest';
+
+import {ILogger} from './types';
+
+/** Interface for working with the GitHub API. */
+export class GitHub {
```
rcebulko: Would be great to isolate this class and related tests into their own PR, easy isolated chunk

checklist/README.md
alanorozco: Oh, yeah the description is wrong.

How about we remove both `GITHUB_` environment variables altogether? The webhook context has both the `owner` and the `repo` name, we can just use those rather than setting them statically.

checklist/src/checklist_bot.ts
```diff
@@ -0,0 +1,76 @@
...
+ * limitations under the License.
+ */
+
+import {Octokit} from '@octokit/rest';
+
+import {GitHub} from './github';
+import {ILogger} from './types';
+import {checklistSectionBody, commentBody} from './template';
+
+const path = 'extensions';
```
danielrozenberg: How about making this a regex to begin with, and not have to construct one from string concatenation? This way you can just match the part of the string you want and pull it from the matcher object
```typescript
const PATH_REGEX = /^extensions\/(.+)\/.+/;`

**Comment by rcebulko** [comment]
> @rcebulko Suggestions on how to split concretely? Most of this is bot boilerplate.

Boilerplate, docs, eslintrc, etc. with an empty application function can take care of many of the minor files and slim this down

GitHub API interface + related tests also could be pulled out

File: checklist/src/checklist_bot.ts (hunk 18 above)
Comment (danielrozenberg) [reply checklist/src/checklist_bot.ts]: Also if we want to let others use this for their own repos you can make this an env variable, but let's not jump ahead of ourselves

File: checklist/README.md [hunk 19]
```diff
@@ -0,0 +1,71 @@
...
+   ```
+   gcloud functions deploy [FUNCTION_NAME] \
+     --entry-point probot \
+     --runtime nodejs10 \
+     --trigger-http
+   ```
+   - When deploying after the first time, the `--runtime` and `--trigger-http` flags may be omitted
+3. To configure Cloud Build auto-deployment, follow [this guide](https://github.com/ampproject/amp-github-apps/blob/master/DEPLOYMENT.md)
+
+This GitHub App is deployed as a [Google Cloud Function](https://cloud.google.com/functions/docs/) at the endpoint: https://ampproject-invite-bot.cloudfunctions.net/
```
rcebulko: This line needs to be updated

checklist/src/github.ts
```diff
@@ -0,0 +1,103 @@
...
+    pull_number: number,
+    path: string
+  ): Promise<Array<string>> {
+    const {data: files} = await this.client.pulls.listFiles({
+      owner: this.org,
+      repo,
+      pull_number,
+    });
+    return files
+      .filter(({filename}) => new RegExp(`^${path}/.+/.+`).test(filename))
```
rcebulko: Nit: This creates a new regexp object for every file in the iterable. Instead, I'd define

```javascript
const pathRegex = new RegExp(`^${path}/.+/.+`)
```
and use that. On another note, I wonder if some of this logic should live in the calling code. This function could be `getPullRequestFiles` and just return a list of filenames, and the ChecklistBot can own the logic for using that list of files.

checklist/package.json
alanorozco: Yes :)

checklist/package.json
alanorozco: Good catch!

checklist/package.json
alanorozco: `amp-contribution-checklist-bot` seems unambiguous enough but rather verbose to me. Up to you.

checklist/src/checklist_bot.ts
alanorozco: I don't think we even need the environment variable, since the repo and owner names come from the PR webhooks.

rcebulko: I agree, I like this suggestion

checklist/app.ts
rcebulko: I was under the impression it doesn't cause I had tests acting weird due to dotenv stuff

checklist/app.ts
rcebulko: I agree, but I didn't comment cause I knew you'd want to :)

checklist/package.json
rcebulko: Eh I guess you're right, checklist bot is good enough

build-system/pr-check.js
```diff
@@ -67,9 +67,6 @@ function main() {
   if (buildTargets.has('BUNDLE_SIZE')) {
     runAppTests('bundle-size');
   }
-  if (buildTargets.has('CHECKLIST')) {
-    runAppTests('checklist');
-  }
```
rcebulko: Nit: Throw a TODO in here to come back to :)

PR #755, opened 2020-03-15
[release-calendar] RTVBlock Component

ajwhatson:
To switch from displaying SingleRTVTable to ManyRTVTable, change the mode from true to false, which is within the App Component's state.

The RTVHeader Component has two hierarchies that change depending on if the user is searching for a single RTV history or a single Channels history. When looking for a single RTV, the mode will be true and displayed will be the inner SingleRTVTable Component. When looking for a single Channel, the mode will be false and the display will be the inner ManyRTVTable. The ManyRTVTable also had many RTVRows which are RTVs within the desired Channel. 

Note: all data is static and mostly likely will change down the line as I consider how information will flow through my app. The types I have created represent the information I will need to correctly populate the components.

### DIFF
--- a/release-calendar/src/client/components/ManyRTVTable.tsx
+++ b/release-calendar/src/client/components/ManyRTVTable.tsx
@@ -29,7 +29,6 @@ export class ManyRTVTable extends React.Component<ManyRTVTableProps, {}> {
     const fakeData = this.props.fakeData;
 
     for (let i = 0; i < fakeData.length; i++) {
-      console.log(fakeData[i]);
       rows.push(<RTVRow Rtv={fakeData[i]}></RTVRow>);
     }
     return (
--- a/release-calendar/src/client/components/ManyRTVTable.tsx
+++ b/release-calendar/src/client/components/ManyRTVTable.tsx
@@ -1,0 +1,42 @@
+/**
+ * Copyright 2020 The AMP HTML Authors. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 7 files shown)


release-calendar/src/client/components/SingleRTVBlock.tsx
```diff
@@ -0,0 +1,36 @@
...
+export interface SingleRTVBlockProps {
+  RTV: string;
+}
+
+export class SingleRTVBlock extends React.Component<SingleRTVBlockProps, {}> {
+  render(): JSX.Element {
+    const RTV = this.props.RTV;
+    return (
+      <tr>
+        <ul> {'Selected RTV: ' + RTV} </ul>
```
danielrozenberg: nit: keep the spacing consistent between the lines :)

release-calendar/src/client/components/SingleRTVBlock.tsx
```diff
@@ -0,0 +1,36 @@
...
+        <ul>{'Release Inforamtion... blah blah'}</ul>
```
danielrozenberg: Sprinkle `TODO(ajwhatson): what needs to be done` comments (or `TODO(#123)` or `TODO(ajwhatson, #123)` if you have a related issue) everywhere that you have stuff like this

release-calendar/src/client/components/ManyRTVTable.tsx
```diff
@@ -0,0 +1,42 @@
...
+  fakeData: RTVRowObject[];
+}
+
+export class ManyRTVTable extends React.Component<ManyRTVTableProps, {}> {
+  render(): JSX.Element {
+    const rows = [];
+    const fakeData = this.props.fakeData;
+
+    for (let i = 0; i < fakeData.length; i++) {
+      console.log(fakeData[i]);
```
danielrozenberg: Debugging leftover?

PR #763, opened 2020-03-16
[release-calendar] ChannelTable Component

ajwhatson:
The ChannelTable Component will display the different channels- long term stable, stable, beta 1%, experimental 1%, beta opt-in, experimental opt-in, and nightly- along with newest RTVs that have gone through those channels.


release-calendar/src/client/components/ChannelTable.tsx
```diff
@@ -0,0 +1,48 @@
...
+}
+
+export class ChannelTable extends React.Component<ChannelTableProps, {}> {
+  render(): JSX.Element {
+    const rows = [];
+    const currentReleases = this.props.currentReleases;
+
+    //change this for loop
+    for (let i = 0; i < currentReleases.length; i++) {
+      console.log(currentReleases[i]);
```
danielrozenberg: Debugging leftover?

release-calendar/src/client/components/ChannelRow.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+export class ChannelRow extends React.Component<ChannelRowProps, {}> {
+  render(): JSX.Element {
+    const channel = this.props.channel;
+    const RTV = this.props.RTV;
+    const enlarge = this.props.isDisplayed;
+    if (enlarge) {
+      return (
+        <tr>
+          <td> {'clicked on'}</td>
```
danielrozenberg: nit: consistent spacing

release-calendar/src/client/components/ChannelRow.tsx
```diff
@@ -0,0 +1,47 @@
...
+  channel: string;
+  RTV: string;
+  isDisplayed: boolean;
+}
+
+export class ChannelRow extends React.Component<ChannelRowProps, {}> {
+  render(): JSX.Element {
+    const channel = this.props.channel;
+    const RTV = this.props.RTV;
+    const enlarge = this.props.isDisplayed;
```
danielrozenberg: There's really no need to assign these temp variables - it's okay to use the longer name directly because it's not that long and it doesn't take a longer time to access, and for each branch you only use it once

release-calendar/src/client/components/ChannelTable.tsx
```diff
@@ -0,0 +1,48 @@
...
+        <ChannelRow
+          channel={currentReleases[i].channel}
+          RTV={currentReleases[i].RTV}
+          isDisplayed={currentReleases[i].isDisplayed}
+          key={i}></ChannelRow>,
+      );
+    }
+    return (
+      <table>
+        <thead>Current Releases</thead>
```
danielrozenberg: A `<thead>` element expects a `<tr>` inside it, and only a `<tr>` https://developer.mozilla.org/en-US/docs/Web/HTML/Element/thead
So you'll have to do something like
```tsx
<thead>
  <tr>
    <th colspan='3'>Current Releases</th>
  </tr>
</thead>
```

release-calendar/src/client/components/ChannelRow.tsx
```diff
@@ -0,0 +1,47 @@
...
+        <tr>
+          <td> {'clicked on'}</td>
+          <td>{channel}</td>
+          <td>{RTV}</td>
+        </tr>
+      );
+    } else {
+      return (
+        <tr>
+          <td>{channel}</td>
```
danielrozenberg: You should still add an empty column at the beginning so they will align with the rest of the table, in which case you can collapse the "if" statement to the inside of that column

release-calendar/src/client/models/data.ts
```diff
@@ -18,36 +18,41 @@ import {Channel, Release} from '../../types';
...
+  ),
+  new Release(
+    '7777777777777',
+    Channel.NIGHTLY,
+    new Date('2020-03-07T08:44:29'),
   ),
+  new Release('8888888888888', Channel.LTS, new Date('2020-03-08T08:44:29')),
+  new Release('9999999999999', Channel.LTS, new Date('2020-03-09T08:44:29')),
+  new Release('2111111111111', Channel.LTS, new Date('2020-03-10T08:44:29')),
+  new Release('233333333333', Channel.LTS, new Date('2020-03-11T08:44:29')),
```
estherkim: Release name is missing a digit

release-calendar/src/client/models/release-channel.ts
```diff
@@ -0,0 +1,37 @@
...
+  channelRTVs.push(
+    releases.find(release => release.channel == 'perc-experimental').name,
+  );
+  channelRTVs.push(
+    releases.find(release => release.channel == 'opt-in-beta').name,
+  );
+  channelRTVs.push(
+    releases.find(release => release.channel == 'opt-in-experimental').name,
+  );
+  channelRTVs.push(releases.find(release => release.channel == 'nightly').name);
```
estherkim: So this returns a sorted array by channel? Do you see any potential bugs this could cause?

release-calendar/src/client/components/App.tsx
```diff
@@ -15,39 +15,106 @@
...
+import Card, {
+  CardActionIcons,
+  CardActions,
+  CardPrimaryContent,
+} from '@material/react-card';
 
 interface AppState {
   events: EventSourceInput[];
+  currentRTVs: string[];
+  selectedChannel: number;
```
estherkim: What does this number represent? The index? Ideally it would be a Channel enum type, not number.

release-calendar/src/client/components/Calendar.tsx
```diff
@@ -23,23 +23,41 @@ import timeGridPlugin from '@fullcalendar/timegrid';
...
-            left: 'prev,next',
-            center: 'title',
-            right: 'dayGridMonth,timeGridWeek,listWeek',
-          }}
-          plugins={[dayGridPlugin, timeGridPlugin]}
-          eventSources={this.props.events}
-        />
-      </div>
-    );
+    if (this.props.selectedChannel != null) {
```
estherkim: This feels like unnecessary duplication. Any way `events` can be set before rendering?

release-calendar/src/client/components/App.tsx
```diff
@@ -15,39 +15,106 @@
...
 import {Calendar} from './Calendar';
+import {Cell, Grid, Row} from '@material/react-layout-grid';
+import {ChannelTable} from './ChannelTable';
 //TODO: remove DATA and instead populate with data from ./test/development-data.ts
+import '@material/react-card/index.scss';
+import '@material/react-layout-grid/index.scss';
 import {DATA} from '../models/data';
 import {EventSourceInput} from '@fullcalendar/core/structs/event-source';
 import {Header} from './Header';
+import {getChannelRTVs} from '../models/release-channel';
```
estherkim: Does this have to be called at the `App` level, or can it be moved to the `ChannelTable component`?

release-calendar/src/client/models/release-channel.ts
```diff
@@ -0,0 +1,37 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {Release} from '../../types';
+
+export function getChannelRTVs(releases: Release[]): string[] {
+  const channelRTVs: string[] = [];
+  channelRTVs.push(releases.find(release => release.channel == 'lts').name);
```
estherkim: How does `releases.find()` ensure that it gets the current release?

release-calendar/src/client/models/release-channel.ts
```diff
@@ -0,0 +1,37 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS-IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {Release} from '../../types';
+
+export function getChannelRTVs(releases: Release[]): string[] {
```
estherkim: Rename so it's clear it gets the latest releases.

release-calendar/src/client/components/App.tsx
ajwhatson: on line 61 of ChannelTable, a list component expects and only accepts a number in as a parameter. `selectedIndex={this.props.selectedChannel}`It would be very tricky to have it as the enum and I would have to convert it to a number in order to pass it here anyway, but I can probably still do it.... is it worth it?

release-calendar/src/client/components/Calendar.tsx
ajwhatson: yes it can be, will do!

release-calendar/src/client/models/release-channel.ts
ajwhatson: I was making the assumption that when I query the releases that I can order them my date. `releases.find()` returns the first of that type, which should be the most current. Is that okay to assume that will eventually be something I can make sure of?

release-calendar/src/client/components/App.tsx
ajwhatson: that's a good question. If i were to move it to the `ChannelTable component` I would probably need to sent all the raw DATA as a parameter to the component. do we like that or should I filter the data in some other way beforehand?

release-calendar/src/client/models/release-channel.ts
ajwhatson: I was making the assumption that when I query the releases that I can order them my date. If youre thinking that assumption isn't something I can make then yeah that could be a big bug. Or do you mean that when I push onto the channelRTVs array. From what I can find online, `.push()` always adds to the last index. Is that not true?

release-calendar/src/client/components/App.tsx
```diff
@@ -15,39 +15,106 @@
  */
 import * as React from 'react';
 import {Calendar} from './Calendar';
+import {Cell, Grid, Row} from '@material/react-layout-grid';
+import {ChannelTable} from './ChannelTable';
 //TODO: remove DATA and instead populate with data from ./test/development-data.ts
+import '@material/react-card/index.scss';
```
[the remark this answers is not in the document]
ajwhatson: I can remove the `react-card` from `App` but the `react-layout-grid` does need to stay

release-calendar/package.json
```diff
@@ -28,6 +28,11 @@
     "@fullcalendar/interaction": "4.4.0",
     "@fullcalendar/react": "4.4.0",
     "@fullcalendar/timegrid": "4.4.0",
+    "@material/react-button": "^0.15.0",
```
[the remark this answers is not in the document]
ajwhatson: hehe, my b

release-calendar/src/client/components/App.tsx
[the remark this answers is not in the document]
estherkim: Ok, let's do that.

release-calendar/src/client/components/App.tsx
estherkim: You can create an API call specific to `ChannelTable` that gets current releases for all channels. This would be different from DATA which I assume gets releases by month.

release-calendar/src/client/models/release-channel.ts
estherkim: This comment has to do the the push. The logic that `array[x]` gives the `y` channel is not maintainable because `array` can be updated at any time, and the developer would have to know somehow that the `x` index means `y` channel.

There should be other data structures that work better than an array.

release-calendar/src/client/components/App.tsx
ajwhatson: DATA actually wont. FullCalendar adapts under the hood to changes in month. I can send it all the release events and it will organize the display for me.

release-calendar/src/client/models/release-channel.ts
ajwhatson: Oh, understood! might do a map again then :)

release-calendar/src/client/components/Calendar.tsx
```diff
@@ -23,10 +23,16 @@ import timeGridPlugin from '@fullcalendar/timegrid';
...
   events: EventSourceInput[];
+  selectedChannel: number;
 }
 
 export class Calendar extends React.Component<CalendarProps, {}> {
   render(): JSX.Element {
+    let eventsToDisplay: EventSourceInput[] = this.props.events;
+    if (this.props.selectedChannel != null) {
+      eventsToDisplay = [this.props.events[this.props.selectedChannel]];
+    }
```
estherkim: Nice! This could use a ternary operator to be cleaner and make `eventToDisplay` a const.

release-calendar/src/client/models/release-channel.ts
```diff
@@ -0,0 +1,36 @@
...
+): Map<string, string> {
+  const channels: string[] = [
+    'lts',
+    'stable',
+    'perc-beta',
+    'perc-experimental',
+    'opt-in-beta',
+    'opt-in-experimental',
+    'nightly',
+  ];
```
estherkim: Can you use the Channel enum instead?

release-calendar/src/client/components/ChannelTable.tsx
```diff
@@ -0,0 +1,132 @@
...
+  constructor(props: Readonly<ChannelTableProps>) {
+    super(props);
+    this.state = {
+      rtvs: new Map(),
+    };
+    this.handleChannelChange = this.handleChannelChange.bind(this);
+  }
+
+  componentDidMount(): void {
+    Promise.resolve(getMostRecentChannelRTVs(DATA)).then(result =>
```
estherkim: You said in another comment that DATA includes all releases and the calendar does the filtering for you. That means that DATA doesn't necessary have only the current releases, which `getMostRecentChannelRTVs` depends on. I think we need a separate API call here and not use DATA.

release-calendar/src/client/models/release-channel.ts
```diff
@@ -0,0 +1,36 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS-IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {Release} from '../../types';
+
+export function getMostRecentChannelRTVs(
```
estherkim: Why not rename to `getCurrentReleases`? Someone unfamiliar with the release process wouldn't know what `RTV` is, since the release schema has `name` and not `RTV`. We should have consistent naming.

release-calendar/src/client/components/ChannelTable.tsx
ajwhatson: I committed again. I still kept the call the same in `ChannelTable` but changed `release-channel` to expected an unordered DATA. I figured for development purposes I shouldnt make a whole new data file if the differences in the calls in the end will most likely be size. So while the data from the server differs later on, `DATA` can still get it done for both

release-calendar/src/client/components/Calendar.tsx
ajwhatson: I used it here, and also changed a function in `ChannelTable` to use the ternary operator too.

release-calendar/src/client/components/ChannelTable.tsx
estherkim: So will DATA always contain current releases? Even if the user is selecting a month from 2 years ago?

release-calendar/src/client/models/release-channel.ts
```diff
@@ -0,0 +1,45 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS-IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {Channel, Release} from '../../types';
+
+function getMostRecent(channel: Channel, releases: Release[]): string {
```
estherkim: Say there are 50 releases. This iterates through it for every channel. Is there another way to get current releases that might be more performant?

release-calendar/src/client/models/release-channel.ts
```diff
@@ -0,0 +1,45 @@
...
+  let currentRelease: Release = null;
+  releases.forEach(release => {
+    if (release.channel == channel && !release.isRollback) {
+      if (currentRelease == null || currentRelease.date < release.date) {
+        currentRelease = release;
+      }
+    }
```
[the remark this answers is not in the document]
ajwhatson: we didn't talk about what would happen if the releases we did call into `ChannelTable` didn't contain all the different release types.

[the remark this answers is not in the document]
ajwhatson: is that the kind of behavior youre talking about?

release-calendar/src/client/models/release-channel.ts
ajwhatson: Can i sort them first and then use `.find()` ? 50*log(50) = 85ish.

release-calendar/src/client/models/release-channel.ts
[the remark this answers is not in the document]
estherkim: Client side code shouldn't make assumptions on data. The cleaner way is to handle this on the server side, i.e. a separate API call for current releases. Lmk if this doesn't make sense.

release-calendar/src/client/components/ChannelTable.tsx
ajwhatson: because we separated the api calls per each component. yes. The channelTable component wont remount unless something within its own state or the state of the props passed down to it are updated.

release-calendar/src/client/models/release-channel.ts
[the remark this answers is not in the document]
ajwhatson: Maybe i used the word "assumption" too quickly. It's our query that dictates what releases the client receive. Can't we query the database to ensure we get the most recent release of each type of channel?

release-calendar/src/client/components/App.tsx
```diff
@@ -23,31 +27,69 @@ import {getEvents} from '../models/release-event';
...
+            phoneColumns={1}
+            tabletColumns={2}
+            align={'middle'}>
+            TODO: Where the SpaceBar will go
+          </Cell>
+        </Row>
+        <hr></hr>
+        <Row>
+          <Cell desktopColumns={3} phoneColumns={1} tabletColumns={2}>
+            <span> TODO: Where RTVTable will go</span>
```
estherkim: Remove this

release-calendar/src/client/components/ChannelTable.tsx
```diff
@@ -0,0 +1,133 @@
...
+    return activatedItem === this.props.selectedChannel
+      ? this.props.handleSelectChannel(null)
+      : this.props.handleSelectChannel(activatedItem);
+  };
+
+  render(): JSX.Element {
+    return (
+      <div>
+        <ListGroup>
+          <ListGroupSubheader tag='h2'>Current Releases</ListGroupSubheader>
```
estherkim: Make this header bigger than the text in the table

release-calendar/src/client/components/ChannelTable.tsx
```diff
@@ -0,0 +1,133 @@
...
+            radioList
+            singleSelection
+            selectedIndex={this.props.selectedChannel}
+            handleSelect={this.handleChannelChange}>
+            <ListItem>
+              <ListItemText
+                primaryText={'Long Term Stable'}
+                secondaryText={this.state.rtvs.get(Channel.LTS)}
+              />
+            </ListItem>
```
estherkim: The text could use some padding from the borders.
image

release-calendar/src/client/components/App.tsx
```diff
@@ -23,31 +27,69 @@ import {getEvents} from '../models/release-event';
...
+            {/* TODO: currently only allows for single channel select, implement multiple selection with checkboxes */}
+            {/* TODO: figure out how to unhighlight a row after it is unselected */}
```
estherkim: This is a bug and should be fixed; not a TODO. Example:
image

release-calendar/src/client/components/Calendar.tsx
```diff
@@ -37,7 +43,7 @@ export class Calendar extends React.Component<CalendarProps, {}> {
             right: 'dayGridMonth,timeGridWeek,listWeek',
           }}
           plugins={[dayGridPlugin, timeGridPlugin]}
-          eventSources={this.props.events}
+          eventSources={eventsToDisplay}
```
estherkim: Can we hide the time in the month view? The time could be useful in the week view.

release-calendar/src/client/stylesheets/calendar.scss
```diff
@@ -6,6 +6,6 @@
   font-family: Arial, Helvetica Neue, Helvetica, sans-serif;
   font-size: 14px;
   margin: 0 auto;
-  max-width: 90%;
+  max-width: 100%;
```
estherkim: Can we set a max-height as well, or something to remove the y scrollbar?

release-calendar/src/client/components/App.tsx
```diff
@@ -23,31 +27,69 @@ import {getEvents} from '../models/release-event';
...
+          <Cell desktopColumns={9} phoneColumns={3} tabletColumns={6}>
+            <Header title='AMP Release Calendar' />
+          </Cell>
+          <Cell
+            desktopColumns={2}
+            phoneColumns={1}
+            tabletColumns={2}
+            align={'middle'}>
+            TODO: Where the SpaceBar will go
+          </Cell>
```
estherkim: Let's remove this until it is ready to be implemented. Also, SpaceBar?

release-calendar/src/client/components/App.tsx
```diff
@@ -23,31 +27,69 @@ import {getEvents} from '../models/release-event';
...
-      <div>
-        <Header title='AMP Release Calendar' />
-        <div>
-          <Calendar events={this.state.events} />
-        </div>
-      </div>
+      <Grid>
+        <Row>
+          <Cell desktopColumns={9} phoneColumns={3} tabletColumns={6}>
+            <Header title='AMP Release Calendar' />
```
estherkim: Can this have the same font as the rest of the page?

release-calendar/src/client/stylesheets/calendar.scss
```diff
@@ -6,6 +6,6 @@
   font-family: Arial, Helvetica Neue, Helvetica, sans-serif;
   font-size: 14px;
   margin: 0 auto;
```
estherkim: The left side of the page has more whitespace than the right side. Can that be more balanced?

release-calendar/src/client/components/ChannelTable.tsx
```diff
@@ -0,0 +1,133 @@
...
+    );
+  }
+
+  handleChannelChange = (activatedItem: number): void => {
+    return activatedItem === this.props.selectedChannel
+      ? this.props.handleSelectChannel(null)
+      : this.props.handleSelectChannel(activatedItem);
+  };
+
+  render(): JSX.Element {
```
estherkim: It's unclear what the colors on the calendar mean. Can this list also serve as a legend, like in your UX mockups?

ajwhatson:
This PR has become too large and complicated. It has now been split into several smaller prs in order to allow for an easier time to review.

PR #766, opened 2020-03-24
[owners] Only suggest reviewers from the allowed reviewer set

rcebulko:
Closes https://github.com/ampproject/amp-github-apps/issues/499

### DIFF
--- a/owners/src/reviewer_selection.js
+++ b/owners/src/reviewer_selection.js
@@ -92,7 +92,17 @@ class ReviewerSelection {
    */
   static _findPotentialReviewers(fileTreeMap) {
     const deepestRules = this._deepestOwnersRules(fileTreeMap);
-    return this._reviewersForRules(deepestRules);
+
+    let root = Object.values(fileTreeMap)[0];
+    while (!root.isRoot) {
+      root = root.parent;
+    }
+    const reviewerSet = root.reviewerSetRule.owners;
+    const isAllowedReviewer = username => reviewerSet.some(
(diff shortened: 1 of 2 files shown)


owners/src/reviewer_selection.js
```diff
@@ -92,7 +92,17 @@ class ReviewerSelection {
...
+    let root = Object.values(fileTreeMap)[0];
+    while (!root.isRoot) {
+      root = root.parent;
+    }
+    const reviewerSet = root.reviewerSetRule.owners;
+    const isAllowedReviewer = username => reviewerSet.some(
+      owner => owner.includes(username)
+    );
+
+    return this._reviewersForRules(deepestRules).filter(isAllowedReviewer);
```
rsimha: How will this work in the rare case where not a single owner is an allowed reviewer? I'd imagine this function will return an empty array.

Would it make sense to request a review from a random owner in such a case?

rcebulko: Two answers:

1) If an error is encountered during reviewer selection or the algorithm selects 0 reviewers, the bot will comment with `"Encountered an error during reviewer selection: \n${error}\nSkipping reviewer assignment"`
2) That particular edge case is impossible as long as there is an owner in some parent directory. In other words, in the worst case scenario, there will always be `cramforce`, `jridgewell`, and `dvoytenko` (top-level owners) available.

rsimha: Fantastic. Thanks for clarifying.

PR #767, opened 2020-03-31
Truncate text that is over the GitHub API limit for CheckRun output

rcebulko:
Truncates output text of check-runs to the API limit of 65535 characters. Result is pretty abrupt "…[TRUNCATED]" and it's possible it could screw up formatting, such as by truncating within an italic block or something, but since this case is so rare I think that's good enough.

### DIFF
--- a/owners/src/ownership/owners_check.js
+++ b/owners/src/ownership/owners_check.js
@@ -50,12 +50,20 @@ class CheckRun {
    * @return {!object} JSON object describing check-run.
    */
   get json() {
+    let text = `${this.text}\n\n${this.helpText}`;
+    if (text.length > this.textLimit) {
+      const info = '…[TRUNCATED]';
+      const over = text.length + info.length - this.textLimit;
+      const partialText = this.text.substr(0, this.text.length - over);
+      text = `${partialText}${info}\n\n${this.helpText}`;
+    }
+
     const checkRun = {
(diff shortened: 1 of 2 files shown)


rcebulko:
The effect on that 1300 file PR:

image

(You can see the truncation at the end of the text but before the help text)

owners/src/ownership/owners_check.js
```diff
@@ -72,6 +80,7 @@ class CheckRun {
     return checkRun;
   }
 }
+CheckRun.prototype.textLimit = 65535;
```
rsimha: I suggest hoisting `65535` to a const at the top of the file and adding a comment that links to the Github API docs (even though they don't document the upper limit).

rcebulko: It's done this way so it can be readily stubbed out in testing code. I'm not aware of a clean way to do so in the case of a module constant used within the module. I'm going add the comment and submit for now, but if you have a different preference here later I will address in a follow-up.

rcebulko: I've established a module constant, and assign it to the prototype field, so that it can still be stubbed. And I've added a comment. Thanks! :+1:

rsimha: > I've established a module constant, and assign it to the prototype field, so that it can still be stubbed.

That's exactly what I meant 👍

PR #768, opened 2020-04-01
Update packages for all GitHub apps

rcebulko:
https://github.com/ampproject/amp-github-apps/pull/640 minus project-metrics Python package updates


.prettierrc
```diff
@@ -2,5 +2,6 @@
   "bracketSpacing": false,
   "singleQuote": true,
   "trailingComma": "es5",
-  "quoteProps": "preserve"
+  "quoteProps": "preserve",
+  "arrowParens": "avoid"
```
rcebulko, line 6: Without this, prettier forces parens around all arrow functions, which is a new rule. Disabling this for now (can have discussion in separate thread if this is controversial) so this can just update packages.

danielrozenberg: 👍

PR #772, opened 2020-04-08
[release-calendar] Adds more test data

### LINKED ISSUE #771: [release-calendar] add more test data
Because of the addition of the promotion table, the previous test data was outdated.

ajwhatson:
Adds test data that range across the different channels. also changes minor parentheses. Closes #771 
Adds releases
-1234567890123 to lts
-2234567890123 to stable
-3234567890123 to perc-beta and perc-experimental
-4234567890123 to opt-in-beta and opt-in-experimental
-5234567890123 to nightly

### DIFF
--- a/release-calendar/test/repository-service.ts
+++ b/release-calendar/test/repository-service.ts
@@ -41,12 +41,7 @@ export class RepositoryService {
 
   async createRelease(release: Release): Promise<Release> {
     const entity = await this.releaseRepository.save(release);
-    //TODO(ajwhatson):clarify with estherkim about what channel path of releases through nightly
-    const promotion = new Promotion(
-      entity,
-      Channel.CREATED,
-      Channel.PERCENT_NIGHTLY,
-    );
+    const promotion = new Promotion(entity, Channel.CREATED, Channel.NIGHTLY);
     await this.savePromotion(promotion);
     return this.getRelease(entity.name);
   }
--- a/release-calendar/test/repository-service.ts
+++ b/release-calendar/test/repository-service.ts
@@ -41,14 +41,9 @@ export class RepositoryService {
 
   async createRelease(release: Release, date?: Date): Promise<Release> {
     const entity = await this.releaseRepository.save(release);
-    const promotion = date? new Promotion(
-      entity,
-      Channel.CREATED,
-      Channel.NIGHTLY,
-      date,
-    ): new Promotion(entity,
-      Channel.CREATED,
-      Channel.NIGHTLY,);
+    const promotion = date
+      ? new Promotion(entity, Channel.CREATED, Channel.NIGHTLY, date)
+      : new Promotion(entity, Channel.CREATED, Channel.NIGHTLY);
     await this.savePromotion(promotion);
     return this.getRelease(entity.name);
   }
--- a/release-calendar/test/repository-service.ts
+++ b/release-calendar/test/repository-service.ts
@@ -41,7 +41,12 @@ export class RepositoryService {
 
   async createRelease(release: Release): Promise<Release> {
     const entity = await this.releaseRepository.save(release);
-    const promotion = new Promotion(entity, Channel.CREATED, Channel.NIGHTLY);
+    //TODO(ajwhatson):clarify with estherkim about what channel path of releases through nightly
+    const promotion = new Promotion(
+      entity,
+      Channel.CREATED,
+      Channel.PERCENT_NIGHTLY,
+    );
     await this.savePromotion(promotion);
     return this.getRelease(entity.name);
   }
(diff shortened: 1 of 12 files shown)


release-calendar/test/test-data.ts
```diff
@@ -32,13 +32,12 @@ export default async function addTestData(
...
-      new Release('4234567890123'), // TODO(estherkim): promote to opt_in_*
-      new Release('5234567890123'), // nightly
-      new Release('6234567890123'),
-    ].map(async (release) => {
+      new Release('1234567890123'), // nightly
+      new Release('2234567890123'), // lts
+      new Release('3234567890123'), // perc-beta and perc-experimental
+      new Release('4234567890123'), // stable
+      new Release('5234567890123'), // opt-in-beta and opt-in-experimental
+    ].map(async release => {
```
estherkim, line 40: Keep the same order as the comments had originally, because `1...` is older than `2...` and reflects our release process.

release-calendar/test/setup.ts
```diff
@@ -31,7 +31,7 @@ async function main(): Promise<void> {
     synchronize: true,
     dropSchema: true,
     logging: false,
-  }).catch((error) => {
+  }).catch(error => {
```
estherkim, line 34: This undos a prettier fix, which we updated recently. To fix, close and reopen your VSCode.

release-calendar/test/test-data.ts
```diff
@@ -32,13 +32,12 @@ export default async function addTestData(
...
       return repositoryService.createRelease(release);
```
estherkim, line 41: Can we stagger the create date so that all releases aren't created on the same day?

release-calendar/test/test-data.ts
ajwhatson: Originally releases were being saved to the repository and then called back using .getReleases. getReleases does not preserve the order that releases were added. In order to avoid having to resort and to not call releases one at a time with .getRelease(name), I pushed the creations and promotions to the repository all at the same time.

release-calendar/test/test-data.ts
ajwhatson: Can do! Also this meant adding a optional date to the parameter of .createRelease which is pretty good to have, so yay for having to do this :)

release-calendar/test/test-data.ts
```diff
@@ -30,59 +30,65 @@ function promoteRelease(
...
+        );
+        promotePromises.push(repositoryService.savePromotions(experimentalPromotions));
+      }
+    }
   }
 
-  await Promise.all(promises);
+  const rollbackDate = new Date();
+  rollbackDate.setDate(today.getDate() + 5 * 5 + 1);
+  promotePromises.push(repositoryService.savePromotions(promoteRelease(releases[5], Channel.NIGHTLY, Channel.ROLLBACK, rollbackDate)));
```
estherkim, line 86: Technically a nightly would never be rolled back (or cherry picked), but a new nightly would be promoted. Feel free to remove this.

danielrozenberg: Hmm... an _old_ nightly might be promoted - e.g. if nightly-wednesday has a P0 in it, we'll promote nightly-monday to take its place. Does that change anything?

release-calendar/test/test-data.ts
```diff
@@ -30,59 +30,59 @@ function promoteRelease(
...
+    new Release('3234567890123'), // perc-beta and perc-experimental
+    new Release('4234567890123'), // opt-in-beta and opt-in-experimental
+    new Release('5234567890123'), // nightly
+  ];
 
-  const releases = await repositoryService.getReleases();
-  const promises = [];
-  const startDate = new Date(Date.now());
+  const today = new Date();
+  const startDate = new Date(today.setMonth(today.getMonth() - 1));
```
estherkim, line 42: Looks like the releases span from `Mar 9 to Mar 29`. So when you pull up the calendar now, it will show no releases? Will it bother you instead of adding releases from `x to today`?

release-calendar/test/test-data.ts
estherkim: Database schema supports it 👍

release-calendar/test/repository-service.ts
```diff
@@ -39,9 +39,11 @@ export class RepositoryService {
     return this.releaseRepository.find({relations: ['promotions']});
   }
 
-  async createRelease(release: Release): Promise<Release> {
+  async createRelease(release: Release, date?: Date): Promise<Release> {
     const entity = await this.releaseRepository.save(release);
-    const promotion = new Promotion(entity, Channel.CREATED, Channel.NIGHTLY);
+    const promotion = date
+      ? new Promotion(entity, Channel.CREATED, Channel.NIGHTLY, date)
```
danielrozenberg, line 45: What happens if you pass `undefined`/`null` as the 4th argument? If it behaves the same as not passing it, you can skip the `?:` and just create the new Promotion instance passing in `date`

ajwhatson: You're right :) i cleaned it up

PR #774, opened 2020-04-10
[release-calendar] Add api call getCurrentReleases

ajwhatson:
Why we need a `.getCurrentReleases()` : `getCurrentReleases `will be used by the client side to query for the the most current release in each Channel. 
Screen Shot 2020-04-20 at 10 15 19 PM

Included:
- Adds ability to query the `Promotion` db table, previously only `Release`s could be returned
- Adds `getCurrentReleases()` to the set of db queries by querying the `Promotion` table directly
- Gives access to `releaseName` which was already the foreign key in table `Promotion` by clarifying the typeorm join column
- Adds `getCurrentReleases()` to `api-service` to be accessible at `http://localhost:3000/current-releases/` 
- Isolates the use of `currentReleases: new Map<Channel, string>()` by saving the data in `channelTable` state
- Adds `promotion-current.ts` to the `view-models` to transform promotions into the `Map<Channel, string>`
- Connects collected current releases to `ChannelTable` `release-button`


estherkim:
> all the releases which have ever passed through that channel will be displayed- this is the main utility of ChannelTable

Is it? I thought ChannelTable was only displaying the current release. I can imagine this table being extremely long if you are displaying every promotion ever.

ajwhatson:
> > all the releases which have ever passed through that channel will be displayed- this is the main utility of ChannelTable
> 
> Is it? I thought ChannelTable was only displaying the current release. I can imagine this table being extremely long if you are displaying every promotion ever.

In the table row, only the most current releases is labeled
Screen Shot 2020-04-10 at 12 01 48 PM
But when the row/channel is selected, then all the history of the releases that have passed through will be shown on the calendar.

ajwhatson:
image
Actually this is a better illustration.

estherkim:
I see. So you're loading the entire promotion table on app start. Then on select, you are passing the channel's promotion data to the calendar component, right? Do you see any pros or cons to that?

ajwhatson:
I assume you're pointing out how much data this is to be querying. 
Possible solutions:
- adding a limit query. 
- only sending the request when someone clicks on a row of the ChannelTable for the first time
- moving all the data off the app component and straight into Calendar

release-calendar/src/client/api-service.ts
```diff
@@ -14,8 +14,8 @@
  * limitations under the License.
  */
 
+import {Channel, Release as ReleaseEntity} from '../types';
```
estherkim, line 17: Also add `ApiService` interface to list of imports

release-calendar/src/types.ts
```diff
@@ -51,4 +51,5 @@ export class Promotion {
 export interface ApiService {
   getRelease: (name: string) => Promise<Release>;
   getReleases: () => Promise<Release[]>;
+  getCurrentReleases: () => Promise<Promotion[]>;
```
estherkim, line 54: The return type is different than the method in api-service.ts?

release-calendar/src/server/repository-service.ts
```diff
@@ -57,4 +57,58 @@ export class RepositoryService {
...
+        'release.promotions',
+        'promotion',
+        'promotion.channel = :channel',
+        {channel: 'rollback'},
+      )
+      .select('release.name')
+      .addSelect('promotion.channel')
+      .getMany();
+
+    const channelQueries = [
```
estherkim, line 83: Instead of querying releases and joining on promotions, why not query the promotions table directly? This way I think you can avoid a join because the promotions table will have `release.name` as a foreign key.

release-calendar/src/server/repository-service.ts
```diff
@@ -57,4 +57,58 @@ export class RepositoryService {
...
+    for (let i = 0; i < arr.length; i++) {
+      if (arr[i].name === rel.name) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  async getCurrentReleases(): Promise<Release[]> {
+    const rollbackQuery = this.releaseRepository
```
estherkim, line 71: See comment below on querying the promotions table directly.

release-calendar/src/server/repository-service.ts
```diff
@@ -57,4 +57,58 @@ export class RepositoryService {
...
+      .createQueryBuilder('release')
+      .innerJoin(
+        'release.promotions',
+        'promotion',
+        'promotion.channel = :channel',
+        {channel: 'rollback'},
```
estherkim, line 77: Use the channel enum.

release-calendar/src/server/repository-service.ts
```diff
@@ -57,4 +57,58 @@ export class RepositoryService {
 
     return releaseQuery;
   }
+
+  include(arr: Release[], rel: Release): boolean {
```
estherkim, line 61: You could use this instead - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter

release-calendar/src/server/repository-service.ts
```diff
@@ -57,4 +57,58 @@ export class RepositoryService {
...
+        .addSelect('promotion.channel')
+        .orderBy('promotion.date', 'DESC')
+        .getMany(),
+    );
+
+    const results: Release[] = [];
+    const rollback = await rollbackQuery;
+    const channels = await Promise.all(channelQueries);
+    channels.forEach((channel) =>
+      results.push(channel.find((first) => !this.include(rollback, first))),
```
estherkim, line 110: How is `channel.find()` working? Isn't `channel` in this case, type `Release`?

ajwhatson: `channel.find()` is scanning the releases in that channel that have never also been rolled back. It's doing a scan from the first to the last, which because I queried the releases in each channel to be in chronologically descending order will start at the most recent. Because of rollbacks happen infrequently, find should only be checking the first few releases that have came out in each channel.

release-calendar/src/server/repository-service.ts
ajwhatson: But I thought `release.name` isn't the foreign key, `id` is. We should talk about this tomorrow morning. I'm confused.

release-calendar/src/server/repository-service.ts
ajwhatson: always :)      sorry

release-calendar/src/server/repository-service.ts
ajwhatson: for example this is all the information I have accessible to me without joining in a promotion `{id: 22, channel: 'nightly', date: 2020-04-17T01:48:43.000Z}`

release-calendar/src/client/api-service.ts
ajwhatson: Ohh, haha, i was wondering about that. Added!

ajwhatson: But also changed to ApiServiceInterface. Otherwise it isn't getting recognized.

release-calendar/src/types.ts
ajwhatson: no, that was just my mistake. I kept switching between returning promotions or releases :( I will try to be more thorough in my personal reviews.

release-calendar/src/server/repository-service.ts
ajwhatson: The function was supposes to check if the `rel` releases was included in `arr` of releases. I should have named the function `isIncluded`. Anyway, I did poke around in javascript build in functions and switched to `.indexOf()` because if `.indexOf()` returns `-1` then the element was not found in the array.

release-calendar/src/server/repository-service.ts
ajwhatson: I did go back and try to clarify the variables. 
```const currentReleases = releasesInEachChannel.map((releasesInOneChannel) =>
      releasesInOneChannel.find(
        (latestRelease) => rollback.indexOf(latestRelease) == -1,
      ),
    );'''

File: release-calendar/src/server/repository-service.ts (hunk 3 above)
Comment (estherkim) [reply release-calendar/src/server/repository-service.ts:73]: Does this work now with the joinColumnOptions you added to the promotion entity? 🎉

File: release-calendar/src/server/repository-service.ts [hunk 8]
```diff
@@ -57,4 +60,43 @@ export class RepositoryService {
 
     return releaseQuery;
   }
+
+  async getCurrentReleases(): Promise<Promotion[]> {
+    const rollbackQuery = this.promotionRepository
+      .createQueryBuilder('promotion')
+      .where('promotion.channel = :channel', {channel: Channel.ROLLBACK})
+      .select('promotion.releaseName')
+      .groupBy('promotion.releaseName')
```
estherkim, line 69: You won't need to group by release name to get rollbacked releases.

release-calendar/src/server/repository-service.ts
```diff
@@ -57,4 +60,43 @@ export class RepositoryService {
...
+        .createQueryBuilder('promotion')
+        .where('promotion.channel = :channel', {channel: eachChannel})
+        .select('promotion.releaseName')
+        .groupBy('promotion.releaseName')
+        .addSelect('promotion.channel')
+        .orderBy('promotion.date', 'DESC')
+        .getMany(),
+    );
+
+    const rollback = await rollbackQuery;
```
estherkim, line 92: nit: rollbacks

release-calendar/src/client/components/ChannelTable.tsx
```diff
@@ -16,19 +16,45 @@
 
 import '../stylesheets/channelTable.scss';
 import * as React from 'react';
+import {ApiService} from '../api-service';
 import {Channel} from '../../types';
+import {getCurrentReleases} from '../models/promotion-current';
```
estherkim, line 21: nit: rename `promotion-current` to something like `current-releases` to stay consistent

release-calendar/src/client/api-service.ts
```diff
@@ -14,20 +14,37 @@
...
       return new Release(release);
     });
   }
+
+  async getCurrentReleases(): Promise<Promotion[]> {
+    const currentReleases = await this.getPromotionRequest(
+      SERVER_URL + '/current-releases/',
+    );
+    return currentReleases.map((promotion: PromotionEntity) => {
+      return new Promotion(promotion);
```
estherkim, line 47: What if you converted `PromotionEntity` directly to `Map<Channel, string>`? That way you don't have to convert it twice.

release-calendar/src/server/repository-service.ts
ajwhatson: I couldn't find good documentation on this, but it was the only way I could figure out how to get distinct releaseNames since we're using mySql. https://typeorm.io/#/select-query-builder/adding-distinct-on-expression-postgres-only

estherkim: It should always be distinct because `release.name` is a PK

release-calendar/src/client/api-service.ts
ajwhatson: That's what I was doing originally but then i realized it wasn't following the view model system we had kind of set up. I can switch back. But just to make sure that would mean removing the `current-releases`/`promotion-current` file in `models`?

release-calendar/src/server/repository-service.ts
ajwhatson: YES! :)

release-calendar/src/client/api-service.ts
estherkim: Yeah. And you can keep the view model system by adding a class to the view model, not necessarily a `Promotion` but something like `CurrentReleases`.

You don't have to have a 1:1 match with view models and entities. View models can be whatever the client needs.

release-calendar/src/server/repository-service.ts
estherkim: Nice! thanks

release-calendar/src/client/components/ChannelTable.tsx
```diff
@@ -16,19 +16,45 @@
...
+  currentReleases: Map<Channel, string>;
+}
 
 interface ChannelTableProps {
   channels: Channel[];
   handleSelectedChannel: (channel: Channel, checked: boolean) => void;
 }
 
-export class ChannelTable extends React.Component<ChannelTableProps, {}> {
+// eslint-disable-next-line prettier/prettier
```
danielrozenberg, line 32: Why?

PR #782, opened 2020-04-16
[release-calendar] Updating db schema

estherkim:
- removes fromChannel 
- renames toChannel to channel


estherkim:
Good point! Next time, can you add the comment directly on the file?

PR #784, opened 2020-04-20
[release-calendar] Event Handling

ajwhatson:
This PR focuses on handling when rows in the `ChannelTable` are checked and sending those selected channels up to the root `App` and back down to the `Calendar`. 

Included:
-  `channels`, an array of selected `Channel`s, which is a state structure in `App` and a prop in `ChannelTable` and `Calendar`.
- the function `handleSelectedChannel` in `App` to handle adding a new channel to the `channels` array or removing a channel that is getting deselected.
- the transfer of `ApiService` to `Calendar` from `App` so that `events`, the `Map<Channel, EventSourceInput>`, can be isolated to only the component that needs it, `Calendar`. 
- the restructuring of `getEvents` to return a `Map<Channel, EventSourceInput>`instead of the more rigid array of `EventSourceInput[]`

### DIFF
--- a/release-calendar/src/client/components/App.tsx
+++ b/release-calendar/src/client/components/App.tsx
@@ -16,33 +16,30 @@
 
 import '../stylesheets/app.scss';
 import * as React from 'react';
-import {ApiService} from '../api-service';
 import {Calendar} from './Calendar';
+import {Channel} from '../../types';
 import {ChannelTable} from './ChannelTable';
-import {EventSourceInput} from '@fullcalendar/core/structs/event-source';
 import {Header} from './Header';
-import {getEvents} from '../models/release-event';
 
 interface AppState {
-  events: EventSourceInput[];
+  channels: Channel[];
 }
 
 export class App extends React.Component<{}, AppState> {
-  private apiService: ApiService;
-
   constructor(props: unknown) {
     super(props);
     this.state = {
-      events: [],
+      channels: [],
     };
-    this.apiService = new ApiService();
   }
 
-  async componentDidMount(): Promise<void> {
-    const releases = await this.apiService.getReleases();
-    const events = getEvents(releases);
-    this.setState({events});
-  }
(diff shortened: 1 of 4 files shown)


release-calendar/src/client/components/App.tsx
```diff
@@ -16,44 +16,44 @@
...
-  async componentDidMount(): Promise<void> {
-    const releases = await this.apiService.getReleases();
-    const events = getEvents(releases);
-    this.setState({events});
-  }
+  handleSelectedChannel = (channel: Channel, toChecked: boolean): void => {
+    this.setState({
+      channels: toChecked
+        ? this.state.channels.concat(channel)
+        : this.state.channels.filter((item) => channel !== item),
```
danielrozenberg, line 40: neat! 😊

release-calendar/src/client/components/Calendar.tsx
```diff
@@ -25,11 +28,36 @@ const CALENDAR_CONTENT_HEIGHT = 480;
...
+  async componentDidMount(): Promise<void> {
+    const releases = await this.apiService.getReleases();
+    this.setState({events: getEvents(releases)});
+  }
+
   render(): JSX.Element {
+    const displayedEvents: EventSourceInput[] = [];
+    this.props.channels.forEach((channel) => {
+      if (this.state.events.has(channel)) {
+        displayedEvents.push(this.state.events.get(channel));
```
danielrozenberg, line 58: Another way of doing this:
```ts
    const displayEvents: EventSourceInput[] = this.props.channels
        .filter((channel) => this.state.events.has(channel))
        .map((channel) => this.state.events.get(channel));
```

(your way is fine too! I'm just personally a fan of functional programming)

ajwhatson: Oh, wait, yes, I was wondering how to make this better! filter and then map, smart.

PR #794, opened 2020-04-23
[error-issue] Identify possible assignees for issues

rcebulko:
Part of https://github.com/ampproject/amphtml/issues/27675

### DIFF
--- a/error-issue/src/issue_builder.ts
+++ b/error-issue/src/issue_builder.ts
@@ -83,7 +83,7 @@ export class IssueBuilder {
     );
   }
 
-  possibleAssignees(): Array<string> {
+  possibleAssignees(limit: number = 3): Array<string> {
     const timeSinceError = Date.now() - this.firstSeen.valueOf();
     if (timeSinceError > ONE_YEAR_MS) {
       // Don't try to guess at assignees for old errors.
@@ -98,7 +98,8 @@ export class IssueBuilder {
       // Ignore lines in the stacktrace from error throwing/logging.
       .filter(({path}) => !ERROR_HANDLING_FILES.includes(path))
       .sort((a, b) => a.committedDate < b.committedDate ? 1 : -1)
-      .map(({author}) => author);
+      .map(({author}) => author)
+      .slice(0, limit);
   }
 
   get bodyNotes(): string {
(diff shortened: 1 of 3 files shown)


error-issue/src/issue_builder.ts
```diff
@@ -79,13 +83,34 @@ export class IssueBuilder {
...
+    );
   }
 
   get bodyNotes(): string {
-    const possibleAssignees = this.possibleAssignees()
+    if (!this.blames.length) {
+      return '';
+    }
+
+    const possibleAssignees = this.possibleAssignees(2)
```
danielrozenberg, line 113: Why is the method's default value 3, and here you're calling it with 2? Would this realistically be changed by the caller? If not I think making it a const/env variable is a better approach

error-issue/test/issue_builder.test.ts
```diff
@@ -83,6 +85,99 @@ describe('IssueBuilder', () => {
     });
   });
 
+  describe('possibleAssignees', () => {
+    const blames = [
```
danielrozenberg, line 89: Consider making this a .json fixture file

error-issue/test/issue_builder.test.ts
```diff
@@ -116,6 +211,14 @@ describe('IssueBuilder', () => {
       expect(builder.bodyNotes).toContain(
         '**Possible assignees:** `@someone`, `@someoneelse`'
       );
+
+      builder.possibleAssignees = jest.fn().mockReturnValue([]);
```
danielrozenberg, line 215: Don't mock methods in the file-under-test - change the conditions of the test so that it ends up generating this result "naturally"

error-issue/test/issue_builder.test.ts
rcebulko: I considered it, but I leaned toward including it here so that someone looking at the test can quickly look at the underlying values and see ie. author, dates, changed files, etc. Instead I added a helper that takes only the relevant fields so it's a bit easier to look at/parse.

error-issue/test/issue_builder.test.ts
rcebulko: Done.

error-issue/src/issue_builder.ts
rcebulko: Made it a const and made the const the default

PR #795, opened 2020-04-23
[release-calendar] Add api call getReleases for all releases

ajwhatson:
Why we need a .getReleases() : getReleases will be used by the client side to query for the all the releases to be displayed on in the `Calendar`.

Included: 
- Removal of the `Release` type on the Client side. Before the data to be displayed on the `Calender` was being transformed from what the server side version of the `Release`type into the client side version of the `Release` type and then to `EventInput`. Now data will go straight from client side `Release` to `EventInput.`
- Addition of an end date to EventInput so that calendar events have a duration
- Addition of a server side `getReleases()` query that groups promotions by channel and then orders by date so to allow for the end date to be calculated later on the server side.
- Change of of a client side `getAllReleasesEvents()` (formally just `getEvents` to preemptively remove confusion for a future `getSingleReleaseEvents()`) so that `getAllReleasesEvents()` is ready to receive an array of `EventInputs`

Note: `Channel.ROLLBACK` will be removed in a future pr.

### DIFF
--- a/release-calendar/src/client/api-service.ts
+++ b/release-calendar/src/client/api-service.ts
@@ -14,23 +14,23 @@
  * limitations under the License.
  */
 
-import {ApiServiceInterface, Channel, Promotion} from '../types';
-import {CurrentReleases, EventInput} from './models/view-models';
+import {Channel, Promotion} from '../types';
+import {CurrentReleases, MyEventInput} from './models/view-models';
 import fetch from 'node-fetch';
 const SERVER_URL = `http://localhost:3000`;
 
-export class ApiService implements ApiServiceInterface {
+export class ApiService {
   private getPromotionRequest(url: string): Promise<Promotion[]> {
     return fetch(url).then((result) => result.json());
   }
 
-  async getReleases(): Promise<EventInput[]> {
+  async getReleases(): Promise<MyEventInput[]> {
     const allPromotions = await this.getPromotionRequest(SERVER_URL);
     const map = new Map<Channel, Date>();
     return allPromotions.map((promotion: Promotion) => {
       const date = map.get(promotion.channel) || new Date();
       map.set(promotion.channel, promotion.date);
-      return new EventInput(promotion, date);
+      return new MyEventInput(promotion, date);
     });
   }
 
--- a/release-calendar/src/client/api-service.ts
+++ b/release-calendar/src/client/api-service.ts
@@ -15,7 +15,7 @@
  */
 
 import {Channel, Promotion} from '../types';
-import {CurrentReleases, MyEventInput} from './models/view-models';
+import {CurrentReleases, ReleaseEventInput} from './models/view-models';
 import fetch from 'node-fetch';
 const SERVER_URL = `http://localhost:3000`;
 
@@ -24,13 +24,13 @@ export class ApiService {
     return fetch(url).then((result) => result.json());
   }
 
-  async getReleases(): Promise<MyEventInput[]> {
+  async getReleases(): Promise<ReleaseEventInput[]> {
     const allPromotions = await this.getPromotionRequest(SERVER_URL);
     const map = new Map<Channel, Date>();
     return allPromotions.map((promotion: Promotion) => {
       const date = map.get(promotion.channel) || new Date();
       map.set(promotion.channel, promotion.date);
-      return new MyEventInput(promotion, date);
+      return new ReleaseEventInput(promotion, date);
     });
   }
 
--- a/release-calendar/src/client/api-service.ts
+++ b/release-calendar/src/client/api-service.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import {ApiServiceInterface, Promotion} from '../types';
+import {ApiServiceInterface, Channel, Promotion} from '../types';
 import {CurrentReleases, EventInput} from './models/view-models';
 import fetch from 'node-fetch';
 const SERVER_URL = `http://localhost:3000`;
@@ -26,24 +26,12 @@ export class ApiService implements ApiServiceInterface {
 
   async getReleases(): Promise<EventInput[]> {
     const allPromotions = await this.getPromotionRequest(SERVER_URL);
-    const map = new Map();
-    allPromotions.forEach((promotion) => {
-      const channelPromotions = map.get(promotion.channel) || [];
-      map.set(promotion.channel, [...channelPromotions, promotion]);
+    const map = new Map<Channel, Date>();
+    return allPromotions.map((promotion: Promotion) => {
+      const date = map.get(promotion.channel) || new Date();
+      map.set(promotion.channel, promotion.date);
+      return new EventInput(promotion, date);
     });
-    const collect: EventInput[] = [];
-    map.forEach((promotionsInChannel: Promotion[]) => {
-      collect.push(new EventInput(promotionsInChannel[0], new Date()));
-      collect.push(
-        ...promotionsInChannel
-          .slice(1)
-          .map(
-            (promotion, i) =>
-              new EventInput(promotion, promotionsInChannel[i].date),
-          ),
-      );
-    });
-    return collect;
   }
 
   async getCurrentReleases(): Promise<CurrentReleases> {
--- a/release-calendar/src/client/api-service.ts
+++ b/release-calendar/src/client/api-service.ts
@@ -14,28 +14,20 @@
  * limitations under the License.
  */
 
-import {
-  ApiServiceInterface,
-  Promotion as PromotionEntity,
-  Release as ReleaseEntity,
-} from '../types';
-import {CurrentReleases, Release} from './models/view-models';
+import {ApiServiceInterface, Promotion as PromotionEntity} from '../types';
+import {CurrentReleases, EventInput} from './models/view-models';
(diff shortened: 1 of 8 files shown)


release-calendar/src/client/api-service.ts
```diff
@@ -14,29 +14,35 @@
...
 
-  async getReleases(): Promise<Release[]> {
-    const releases = await this.getReleaseRequest(SERVER_URL);
-    return releases.map((release: ReleaseEntity) => {
-      return new Release(release);
+  async getReleases(): Promise<EventInput[]> {
+    const promotions = await this.getPromotionRequests(SERVER_URL);
+    const collect: EventInput[] = [];
+    promotions.forEach((promotionsInChannel: Promotion[]) => {
+      const endDate: Date = new Date();
```
danielrozenberg, line 34: You can move this line outside of the `forEach` callback function - the way it is now it calls `new Date()` for every promotion

release-calendar/src/client/api-service.ts
```diff
@@ -14,29 +14,35 @@
...
+      collect.push(new EventInput(promotionsInChannel[0], endDate));
+      for (let i = 1; i < promotionsInChannel.length; i++) {
```
danielrozenberg, line 36: I'm confused about this... it looks like you're iterating `promotionsInChannel` inside `promotionsInChannel`, so you'll have a lof of duplicates, wouldn't you?

release-calendar/src/server/repository-service.ts
```diff
@@ -44,21 +54,16 @@ export class RepositoryService {
...
-    releaseQuery.then((releases) => {
-      releases.forEach((release) =>
-        release.promotions.sort(
-          (a, b): number => b.date.getTime() - a.date.getTime(),
-        ),
-      );
-    });
-
-    return releaseQuery;
+  getReleases(): Promise<Promotion[][]> {
```
danielrozenberg, line 57: What would be a sample output for this method?

release-calendar/src/client/models/release-event.ts
```diff
@@ -17,28 +17,20 @@
...
-    const event = convertReleaseToEvent(release);
-    const channelEvents = eventInputs.get(release.channel);
+  events.forEach((event) => {
+    const channelEvents = eventInputs.get(event.channel);
     if (!channelEvents) {
-      eventInputs.set(release.channel, [event]);
+      eventInputs.set(event.channel, [event]);
     } else {
-      eventInputs.set(release.channel, [...channelEvents, event]);
+      eventInputs.set(event.channel, [...channelEvents, event]);
```
danielrozenberg, line 30: Shorter version:
```typescript
const channelEvents = eventInputs.get(event.channel) || [];
eventInputs.set(event.channel, [...channelEvents, event]);
```

release-calendar/src/client/api-service.ts
ajwhatson: Hm, `promotions` is originally a 2d array which I'm making a 1d array out of. So `promotionsInChannel` is a row of promotions. Once I transform a single promotion that's from whatever row into a EventInput, I push it onto `collect`, the 1d array i'm making.

release-calendar/src/server/repository-service.ts
ajwhatson: [[LTS promotion, LTS promotion,..], [Stable promotion, Stable promotion,..], [Opt-in Bets promotion, Opt-in Beta promotion,..],...]

release-calendar/src/client/models/release-event.ts
ajwhatson: whoa

release-calendar/src/client/api-service.ts
ajwhatson: Just to clarify, endDate needs to be set to a new Date (new Date is automatically set to the time its called - think current) for each row because the promotion that ends up getting set first is the most recent (I queried the promotions in descending order by date). That promotion remains the release in that channel at the time the viewer is seeing the calendar.

danielrozenberg: I figure out that bit - but you can instantiate the constant `endDate` once, before the `forEach` line, and then use it inside the `forEach` block

release-calendar/src/client/api-service.ts
danielrozenberg: Ah! Yes, the new variable names definitely help!

release-calendar/src/client/api-service.ts
```diff
@@ -28,17 +28,15 @@ export class ApiService implements ApiServiceInterface {
...
-      collect.push(new EventInput(promotionsInChannel[0], endDate));
-      for (let i = 1; i < promotionsInChannel.length; i++) {
+      collect.push(new EventInput(rowOfPromotions[0], endDate));
+      for (let i = 1; i < rowOfPromotions.length; i++) {
         collect.push(
-          new EventInput(
-            promotionsInChannel[i],
-            promotionsInChannel[i - 1].date,
-          ),
+          new EventInput(rowOfPromotions[i], rowOfPromotions[i - 1].date),
```
danielrozenberg: More cool code for your consideration:
```typescript
collect.push(...rowOfPromotions.slice(1).map(
    (promotion, i) => new EventInput(promotion, rowOfPromotion[i - 1].date))
);

// or

rowOfPromotions.slice(1).forEach((promotion, i) => {
  collect.push(new EventInput(promotion, rowOfPromotion[i - 1].date));
});
```

release-calendar/src/server/repository-service.ts
danielrozenberg: Hmm... can you make this a `Map<Channel, Promotion[]>` instead?
```typescript
{
  Channel.LTS: [LTS Promotion, LTS Promotion, ...],
  Channel.Stable: [Stable promotion, Stable promotion, ...],
  Channel.OptInBeta: [Opt-in Beta promotion, Opt-in Beta promotion, ...],
  ...
}
```

ajwhatson:
@danielrozenberg I couldn't come up with a solution I liked very much with a `Map<Channel, Promotion[]>` because of the way node-fetch wants the fetch to return a single promise. But in having to rework things, I came up with a solution using a much smaller map of `Map<Channel, Date>` in which the date that marks the switch between releases is saved while iterating through the whole table of promotions. Take a look and see what you think, but I like it much better. If you dont, there is also a previous attempt in the prior commit, `63b7c4b` (`create map in api-service`) that also runs.

release-calendar/src/client/api-service.ts
```diff
@@ -23,24 +23,15 @@ export class ApiService implements ApiServiceInterface {
...
-        collect.push(
-          new EventInput(rowOfPromotions[i], rowOfPromotions[i - 1].date),
-        );
-      }
+    const allPromotions = await this.getPromotionRequest(SERVER_URL);
+    const map = new Map<Channel, Date>();
+    return allPromotions.map((promotion: Promotion) => {
+      const date = map.get(promotion.channel) || new Date();
+      map.set(promotion.channel, promotion.date);
+      return new EventInput(promotion, date);
```
danielrozenberg, line 33: I'm not sure I understand the two branches here - what happens in each case?

ajwhatson: Either the date is taken from the map, which contains the saved date from the previous promotion in that channel, or the date will be set to the current time because the promotion is the first promotion iterated over in its channel.

danielrozenberg: Ah, I get it now. Makes sense :)

release-calendar/src/types.ts
```diff
@@ -55,6 +53,6 @@ export class Promotion {
 }
 
 export interface ApiServiceInterface {
-  getReleases: () => Promise<ReleaseViewModel[]>;
+  getReleases: () => Promise<EventInput[]>;
```
estherkim, line 56: The api service uses `EventInput` from `view-models` but the interface uses fullcalendar's? Not sure what's going on.

estherkim: On that note why don't we get rid of this interface. It's not really helping!

release-calendar/src/client/models/view-models.ts
```diff
@@ -14,21 +14,28 @@
...
+import {Channel, Promotion} from '../../types';
 
-export class Release {
-  constructor(entity: ReleaseEntity) {
-    const currentPromotion = entity.promotions[0];
-    this.name = entity.name;
-    this.channel = currentPromotion.channel;
-    this.date = currentPromotion.date;
-    this.isRollback = this.channel == Channel.ROLLBACK;
+export class EventInput {
```
estherkim, line 19: Since FullCalendar has an `EventInput` type already, can we rename this to something else?

release-calendar/src/client/models/view-models.ts
```diff
@@ -15,8 +15,9 @@
  */
 
 import {Channel, Promotion} from '../../types';
+import {EventInput} from '@fullcalendar/core';
 
-export class EventInput {
+export class MyEventInput implements EventInput {
```
estherkim, line 20: Haha how about `ReleaseEventInput`?  "My" reminds me of demo or example code.

ajwhatson: I knew you'd come up with a better name :)

PR #796, opened 2020-04-23
[owners] Fix team @ mentions

rcebulko:
App-authenticated comments appear to come from the app user (ie. `amp-owners-bot`) but can't tag teams. The workaround is doing a fake update to the comment with a user-authenticated client, which enables the @ mention but leaves the comment author as the app name.

image

### DIFF
--- a/owners/src/notifier.js
+++ b/owners/src/notifier.js
@@ -81,7 +81,7 @@ class OwnersNotifier {
    * @param {!GitHub} github GitHub API interface.
    */
   async createNotificationComment(github) {
-    const [botComment] = await github.getBotComments(this.pr.number);
+    let [botComment] = await github.getBotComments(this.pr.number);
     const notifies = this.getOwnersToNotify();
     delete notifies[this.pr.author];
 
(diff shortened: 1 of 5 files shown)


danielrozenberg:
Unless you're in a rush to get this running, have you tried contacting GitHub support and asking about this? I feel like this should work from the app, instead of going through a hack

rcebulko:
@danielrozenberg I definitely wanted it fixed because it was preventing the bot from notifying wg-caching about validator changes. I haven't reached out to support directly, but this is a well-established challenge that has a lot of discussions on the web and the end result is always using a personal access token. I gave the app every possible permission and nothing made the tag work when using app auth. I can reach out to support directly, but in the interim I'd like to make sure the team mentions are working

owners/src/notifier.js
```diff
@@ -100,11 +100,24 @@ class OwnersNotifier {
...
+
+    // App-authenticated comments appear to come from the app user (ie.
+    // `amp-owners-bot`) but can't tag teams. The workaround is doing a fake
+    // update to the comment with a user-authenticated client, which enables the
+    // @ mention but leaves the comment author as the app name.
+    const teamNotifies = Object.keys(notifies).filter(n => n.includes('/'));
+    if (teamNotifies.length) {
+      await github.user.updateComment(
+        botComment.id,
+        `${body}\n<!-- Edited to fix team @mention -->`
```
danielrozenberg, line 119: `@mention` → `@ mentions` just in case the parser decides to tag the account called `mention` one day (this account doesn't exist, but we should be nice)

rcebulko: Good catch! I usually do that when typing "@ mention", missed it here!

rcebulko: (Though this _is_ a comment, so it wouldn't tag anyway)

PR #797, opened 2020-04-24
[release-calendar] Remove Rollbacks from Channel enum

ajwhatson:
This will remove `Channel.ROLLBACK` from the `Channel` enum. When a rollback occurs, instead of marking a release as a rollback by adding a rollback promotion, the older releases that are being reinstated will receive new promotions.

For example, a release could have these promotions, ordered by date. Oldest to latest.
`[nightly]`, `[opt-in beta]`, `[opt-in experimental]`, `[% beta]`, `[% experimental]`, `[stable]`, `[long term stable]`, *the previous release in stable was rolled back*, `[stable]`

Included:
- removal of `Channel.ROLLBACK` from `Channel` enum
- adaption of `getCurrentReleases()` query in server side
- removal of `isRollback` boolean in `view-model`

### DIFF
--- a/release-calendar/src/server/repository-service.ts
+++ b/release-calendar/src/server/repository-service.ts
@@ -62,25 +62,17 @@ export class RepositoryService {
   }
 
   async getCurrentReleases(): Promise<Promotion[]> {
-    const channelQueries = [
-      Channel.LTS,
-      Channel.NIGHTLY,
-      Channel.OPT_IN_BETA,
-      Channel.OPT_IN_EXPERIMENTAL,
-      Channel.PERCENT_BETA,
-      Channel.PERCENT_EXPERIMENTAL,
-      Channel.STABLE,
-    ].map((eachChannel) =>
-      this.promotionRepository
-        .createQueryBuilder('promotion')
-        .where('promotion.channel = :channel', {channel: eachChannel})
-        .select('promotion.releaseName')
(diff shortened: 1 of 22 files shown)


release-calendar/src/server/repository-service.ts
```diff
@@ -62,14 +62,6 @@ export class RepositoryService {
...
   async getCurrentReleases(): Promise<Promotion[]> {
-    const rollbackQuery = this.promotionRepository
-      .createQueryBuilder('promotion')
-      .where('promotion.channel = :channel', {channel: Channel.ROLLBACK})
-      .select('promotion.releaseName')
-      .groupBy('promotion.releaseName')
-      .addSelect('promotion.channel')
-      .getMany();
-
     const channelQueries = [
```
danielrozenberg, line 65: Now that all the Channel values are included here, you can do:
```typescript
const channelQueries = Object.keys(Channel).map((eachChannel) => {
    ...
```

Or even
```typescript
return Promise.all(Object.keys(Channel).map((eachChannel) => {
    ...
```

danielrozenberg: Supernit: if you rename the anonymous function's argument from `eachChannel` to `channel`, you can change the `.where` line to `.where('promotion.channel = :channel', {channel})`

PR #798, opened 2020-04-24
[test-status] update build-cop/report endpoint to use new data format

danielrozenberg:


test-status/api.js
```diff
@@ -373,13 +373,17 @@ exports.installApiRouter = (app, db) => {
   });
 
   buildCop.post('/update', async (request, response) => {
-    if (!('username' in request.body) || !request.body.username) {
+    if (
+      !('build-cop' in request.body) ||
+      !('primary' in request.body['build-cop']) ||
+      !request.body['build-cop'].primary
```
rcebulko, line 379: I can't wait for optional chaining...

danielrozenberg: It's already on the recently released Node 14, if you feel like taking the plunge........ (and if GAE supports it, which I haven't checked yet)

PR #805, opened 2020-04-27
[release-calendar] Add getRelease()

ajwhatson:
Why we need a ` getRelease(requestedRelease: string)`: getRelease will be used by the client side to query for all the promotions a single release has underwent so that they can be displayed on the `Calender.`

This is what one example release looks like as it flows from Nightly to LTS.
Screen Shot 2020-04-28 at 12 08 03 PM

Included:

- Addition of a server side getRelease() query that joins the promotions and release tables and selects the requested release and all of its promotions (promotions in descending order)
- Addition of the `Release` string that will live in the `App` state but also in the props of `Calendar`
- Event handling associated with clicking on a `rtv-button`
- Addition of `getSingleReleaseEvents()` which converts the `ReleaseEventInputs` into `EventSourceInput[]` to be read by the calendar.

TODO: Toggle between a `rtv-button` and `channel-row`

### DIFF
--- a/release-calendar/src/client/api-service.ts
+++ b/release-calendar/src/client/api-service.ts
@@ -30,7 +30,7 @@ export class ApiService {
 
   async getRelease(requestedRelease: string): Promise<ReleaseEventInput[]> {
     const release = await this.getReleaseRequest(
-      SERVER_URL + '/releases/' + requestedRelease,
+      `${SERVER_URL}/releases/${requestedRelease}`,
     );
     return [
       new ReleaseEventInput(release.promotions[0], new Date()),
@@ -55,7 +55,7 @@ export class ApiService {
 
   async getCurrentReleases(): Promise<CurrentReleases> {
     const currentReleases = await this.getPromotionRequest(
-      SERVER_URL + '/current-releases/',
+      `${SERVER_URL}/current-releases/`,
     );
     return new CurrentReleases(currentReleases);
   }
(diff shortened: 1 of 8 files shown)


release-calendar/src/server/repository-service.ts
```diff
@@ -43,14 +43,9 @@ export class RepositoryService {
       .createQueryBuilder('release')
       .leftJoinAndSelect('release.promotions', 'promotion')
       .where('release.name = :name', {name})
+      .orderBy('promotion.date', 'DESC')
```
estherkim, line 46: Oh this works?

ajwhatson: Yep!

estherkim: Nice :)

estherkim: We need to start writing unit tests soon...

release-calendar/src/client/api-service.ts
```diff
@@ -24,6 +24,25 @@ export class ApiService {
     return fetch(url).then((result) => result.json());
   }
 
+  private getReleaseRequest(url: string): Promise<Release> {
+    return fetch(url).then((result) => result.json());
+  }
+
+  async getRelease(requestedRelease: string): Promise<ReleaseEventInput[]> {
+    const release = await this.getReleaseRequest(
+      SERVER_URL + '/releases/' + requestedRelease,
```
danielrozenberg, line 33: Prefer ` syntax over string concatenation:
```
`${SERVER_URL}/releases/${requestedRelease}`
```

Issue #809, opened 2020-04-29
Figure out how to share eslint config across repo


rcebulko:
As this repo expands, the eslint configs have been growing and diverging. It would be really cool to have things like the `notice` rule in a standard config, but things like react and other project quirks make using a single share config infeasible.

WDYT of having a root-level `typescript.eslintrc.json` and `javascript.eslintrc.json` that each app can import/ready and then deep-merge where changes are needed? This would allow for standard rules across the repo, while still allowing individual apps to add their own plugins and settings.

/cc @ampproject/wg-infra

rcebulko:
Proof-of-concept for this approach: https://github.com/ampproject/amp-github-apps/pull/810

rsimha:
I just noticed that some of the project-specific lint rule files in `amp-github-apps` are using `root: true`, which prevents them from using `eslint`'s [built-in config merging capabilities](https://eslint.org/docs/user-guide/configuring#configuration-cascading-and-hierarchy). For example:

https://github.com/ampproject/amp-github-apps/blob/1a508b39c855c5ee6bbed8c9db89335edbdc3f64/error-issue/.eslintrc.js#L4

https://github.com/ampproject/amp-github-apps/blob/1a508b39c855c5ee6bbed8c9db89335edbdc3f64/pr-deploy/.eslintrc.js#L4

Perhaps the approach in #810 can be simplified as follows?

- Make sure only the project-root `.eslintrc.js` has `root: true`
- Lint rules that are common to all projects live in that file
- Project-specific rules live in `.eslintrc.js` in the project folder (without `root: true`)
- For `.ts` specific rules, create an `overrides` section like this (root-level, project-level, or both)

```json
  "overrides": [
    {
      "files": ["*.ts", "*.tsx"],
      "parser": "@typescript-eslint/parser",
      "plugins": ["@typescript-eslint"],
      "rules": {
        "@typescript-eslint/no-unused-vars": 2,
        "@typescript-eslint/class-name-casing": 2,
        "@typescript-eslint/member-delimiter-style": 2
      }
    }
  ]
```

With this, configs are auto-merged, as described in the [`eslint` docs](https://eslint.org/docs/user-guide/configuring#configuration-cascading-and-hierarchy).

rcebulko:
I wish I'd known what that did beforehand :disappointed: I wonder why we have `root: true` everywhere then :thinking:

rcebulko:
[...]

PR #810, opened 2020-04-29
[DO NOT SUBMIT] Standardize eslint config for JS/TS apps

rcebulko:
This PR contains the actual config changes, but will fail in Travis because of lint errors. This is a proof-of-concept for what the idea suggested in #809 would look like in practice for both JS and TS apps.


danielrozenberg:
I am very much in favour of this

Issue #778, opened 2020-04-14
[onduty] Add macros to assign release-on-duty and build-cop


danielrozenberg:
Maybe `/assign-release` and `/assign-build`? (I have absolutely no strong opinion here, other than not liking the term "build-cop" itself, but not enough to call any vetoes)

rcebulko:
I wasn't thinking of appending this to the invite bot, I was thinking of a separate uber-simple bot that listens for like `/release-cop` and `/build-cop` and comments @mentioning them. Past that point I'd leave it to the thread to determine if they should assign or not

danielrozenberg:
That also works :) I figured it's within the scope of the invite bot because it already has a /tryassign action

rcebulko:
My inclination would be for now to just spin up another small Cloud Function for this, and if we find a third case where we want to support macros, maybe it becomes worthwhile to merge it all somehow

rcebulko:
The more I've thought about this, the more I'm leaning toward an approach of having two dedicated teams like `@ampproject/release-onduty` and `@ampproject/build-cop`, and let a cron job with access to the rotations system update those teams automatically. This would fit more easily into the workflow of most developers who are used to tagging people, and would allow for assigning issues and PRs to those aliases

danielrozenberg:
Fine by me 🤷‍♂

mrjoro:
SGTM; thanks!

rcebulko:
This is done; closing.

PR #812, opened 2020-04-29
[onduty] Add Onduty Bot to handle rotation updates and update teams

rcebulko:
Part of #778

### DIFF
--- a/onduty/src/bot.ts
+++ b/onduty/src/bot.ts
@@ -35,10 +35,6 @@ export class OndutyBot {
 
   async updateRotation(type: RotationType, rotation: Rotation): Promise<void> {
     const teamName = this.rotationTeams[type];
-    if (!teamName) {
-      this.logger.error(`[updateRotation] Missing ${type} rotation team`);
-      return;
-    }
     this.logger.info('[updateRotation] Updating ${type} rotation');
 
     const members = await this.github.getTeamMembers(teamName);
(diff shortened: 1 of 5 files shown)


onduty/src/bot.ts
```diff
@@ -0,0 +1,60 @@
...
+    const teamName = this.rotationTeams[type];
+    if (!teamName) {
+      this.logger.error(`[updateRotation] Missing ${type} rotation team`);
+      return;
+    }
+    this.logger.info('[updateRotation] Updating ${type} rotation');
+
+    const members = await this.github.getTeamMembers(teamName);
+    const currentRotation = Object.values(rotation)
+      .filter(Boolean)
```
estherkim, line 42: Curious, why is `filter(Boolean)` needed?

onduty/types/index.d.ts
```diff
@@ -22,4 +22,13 @@ declare module 'onduty' {
     error(message: string, ...extraInfo: unknown[]): void;
     info(message: string, ...extraInfo: unknown[]): void;
   }
+
+  export interface Rotation {
+    primary: string;
+    secondary: null | string;
+  }
```
estherkim, line 29: GitHub teams don't have a concept for which members are primary or secondary, right? Could this be an array of usernames instead?

rcebulko: This is the format of the payload from the rotation reporter tool :)

onduty/src/bot.ts
rcebulko: `secondary` may not always be present in the payload from the rotation reporting tool

PR #813, opened 2020-04-30
[release-calendar] Change tooltip to use hooks 🎣 and portals

ajwhatson:
This creates a tooltip that utilizes react portals in order to live outside of the `calendar` component in which is was made. 

This is an example of a tooltip living outside of its parent component.
Screen Shot 2020-04-30 at 2 29 01 PM

Included:
- installation of @popperjs/core (for the tooltip), moment (to assist in the format of time in tooltip), react-popper (so the tooltip can be rendered in the correct position)
- creation of `Tooltip` and `Hook` which together manage the positioning of the tooltip. `Tooltip` replaces the event element which would otherwise be automatically rendered FullCalendar. It replaces it with the hook. `Hook` makes the connection between the replacement event element and the new `EventCard`. `Hook` also conceals the `EventCard` when someone clicks away from the element and does so with `eventListeners`.
- creation of `EventCard` which renders rows of information about the release. Each row has a related icon to the left and information to the right. The release cards match the corresponding channels by displaying a thin top bar and transparent background of the channel color.
- styling for `eventCard`
- styling adjustments to `calendar` in order to overwrite the default button colors of FullCalendar. FullCalendar.io does not allow for customized styling other than full Bootstrap themes- that seemed like an overkill.
- small adjustments to the calendar so that the tooltip will not be cut off when it's positioned very low on the screen
- change of testData from the first release being named 1234567890123 to 2004252135000 which instead is a real release tag so as to test the link to the github release body. Link will create a new tab.
- emojis for channel schedule
- add `getReleaseDates(requestedRelease: string)` to client side that uses the already made `getCurrentReleases()` call on the server side. This will carry `ReleaseDate[]` which contains promotion channels and start dates so to populate the emoji schedule.

- TODO: Add cherrypick attribute to `releaseDates` currently the code on `EventCard` is commented out
- TODO: Place emojis in rtv square on `ChannelTable`


release-calendar/src/client/components/EventCard.tsx
```diff
@@ -0,0 +1,94 @@
...
+import moment from 'moment';
+
+export interface EventCardProps {
+  eventApi: EventApi;
+}
+
+export class EventCard extends React.Component<EventCardProps, {}> {
+  //TODO: use readable channel titles in eventCard
+
+  content = [
```
estherkim, line 29: Instead of entered and exit, what if it showed the whole history? Something like
```
Created as Nightly on xx
Promoted to Opt-In Experimental, Opt-In Beta on xx
Promoted to 1% Experimental, 1% Beta on xx
Promoted to Stable on xx
Promoted to LTS on xx
```

release-calendar/src/client/components/EventCard.tsx
```diff
@@ -0,0 +1,94 @@
...
+              </div>
+              <div className={'text'}>
+                <a
+                  href={
+                    'https://github.com/ampproject/amphtml/releases/tag/' +
+                    this.props.eventApi.title
+                  }
+                  target='_blank'
+                  rel='noopener noreferrer'>
+                  {'to the github release body..'}
```
estherkim, line 85: This can just say "GitHub release" or "View details"

release-calendar/src/client/components/EventCard.tsx
```diff
@@ -0,0 +1,94 @@
...
+        moment(this.props.eventApi.end).format('MMMM Do, hA'),
+    },
+  ];
+
+  render(): JSX.Element {
+    return (
+      <div className={'event-card'}>
+        <div
+          className={this.props.eventApi.classNames[0]}
+          style={{
```
estherkim, line 50: Move all these inline styles to an SCSS file

release-calendar/src/client/components/EventCard.tsx
```diff
@@ -0,0 +1,94 @@
...
+            height: 'inherit',
+          }}>
+          <div className={'event-top'}></div>
+          <div className={'event-content'}>
+            {this.content.map((row) => {
+              return (
+                <div className={'content-row'} key={row.icon}>
+                  <div className={'icon'}>
+                    <i
+                      className='material-icons'
```
estherkim, line 62: Is it possible to use emojis instead of icons? We usually use 🚄 for release

ajwhatson: Also do you have a better idea for a Github emoji? 🖥is just sad

ajwhatson: And! for LTS, what's better 🏙or 🏰? or other?

estherkim: Hmm I think 🏙

estherkim: For the GitHub release link, I'd put 🔗

release-calendar/src/client/components/EventCard.tsx
```diff
@@ -0,0 +1,160 @@
...
+                  return (
+                    match && (
+                      <div className={'content-row'} key={match.emoji}>
+                        <div className={'emoji'}>
+                          <span role='img' aria-label={'row.emojiName'}>
+                            {match.emoji}
+                          </span>
+                        </div>
+                        <div className={'text'}>
+                          {match.text + moment(row.date).format('MMMM Do, hA')}
```
estherkim, line 148: Add timezone

release-calendar/src/client/components/EventCard.tsx
```diff
@@ -0,0 +1,160 @@
...
+                  <span role='img' aria-label={'laptop'}>
+                    {'🖥'}
+                  </span>
+                </div>
+                <div className={'text'}>
+                  <a
+                    href={`https://github.com/ampproject/amphtml/releases/tag/${this.props.eventApi.title}`}
+                    target='_blank'
+                    rel='noopener noreferrer'>
+                    {'Github Release'}
```
estherkim, line 129: nit: GitHub

release-calendar/src/client/components/EventCard.tsx
estherkim, line 148: Instead of having a `match.text` field, this can serve as a template to remove duplication.

release-calendar/src/client/components/EventCard.tsx
```diff
@@ -0,0 +1,160 @@
...
+      this.props.eventApi.title,
+    );
+    this.setState({releaseDates});
+  }
+
+  schedule = [
+    {
+      channel: Channel.NIGHTLY,
+      text: 'Created as Nightly on ',
+      title: 'Nightly',
```
estherkim, line 54: Can we reuse the titles from the ChannelTable component?

release-calendar/src/client/components/TooltipHook.tsx
```diff
@@ -0,0 +1,104 @@
...
+        className={'event-button'}
+        ref={setReferenceElement}
+        onClick={(): void => setClick(!isClicked)}>
+        {children.event.title}
+      </button>
+      {ReactDOM.createPortal(
+        <div ref={node}>
+          <div
+            ref={setPopperElement}
+            style={{...{zIndex: 999}, ...styles.popper}}
```
estherkim, line 95: Can this be moved to an SCSS file? Not sure what `popper` suggests

release-calendar/src/client/index.html
```diff
@@ -0,0 +1,15 @@
+<!DOCTYPE html>
```
estherkim, line 1: This file was recently deleted to use `src/client/index.ejs` as a template instead.

release-calendar/src/client/components/EventCard.tsx
ajwhatson: I'm not sure what you mean. Hm, if you're thinking that there could be something like
Promoted to Opt-in Experimental, Opt-In Beta on April 20th, 11 am
Promoted to Opt-in Experimental, Opt-In Beta on April 20th, 11 am
one line which represents `Channel.OPT_IN_EXPERIMENTAL` and the other `Channel.OPT_IN_BETA,` that wont happen because `Channel.OPT_IN_BETA` doesn't exist in the schedule template. There is the possibility of duplication if there had been a rollback and a release is getting reinstated to a channel it was already in. Is that what you're referring to? If so, I can brainstorm about adding a rollback line to divide the two promotions?

release-calendar/src/client/components/TooltipHook.tsx
ajwhatson: There's a couple of things I can do. 
1. Change the CSSProperty in styles.popper directly. 
`styles.popper['zIndex'] = 999;`
2. Keep what I have above
`style={{...{zIndex: 999}, ...styles.popper}}`
3. Or declare `zIndex: 999` in a different variable
`const bringForward: React.CSSProperties = {zIndex: 999};`
`style={{...bringForward,...styles.popper}}`

Popper doesn't have a suggestion for smaller changes to style like this. But when doing a full customized tooltip, they suggest the third (when I wouldn't use styles.popper at all).

release-calendar/src/client/components/EventCard.tsx
estherkim: By template I mean removing the `match.text` field and putting here:
`{match.title} was promoted ...`

You don't have to have a single line for experimental/beta, if it makes it easier. 1 channel per line is okay.

release-calendar/package.json
```diff
@@ -24,8 +24,6 @@
     "setup": "npm run setup-build && npm run setup-serve",
     "setup-build": "webpack --config ./webpack.setup.config.js",
     "setup-serve": "node -r dotenv/config ./dist/bundle-setup.js",
-    "setup-build": "webpack --config webpack.setup.config.js",
-    "setup-serve": "node -r dotenv/config dist/bundle-setup.js",
```
estherkim, line 28: Don't remove this please

release-calendar/src/client/components/EventCard.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+export interface EventCardProps {
+  eventApi: EventApi;
+}
+
+interface EventCardState {
+  releaseDates: ReleaseDates;
+}
+
+export class EventCard extends React.Component<EventCardProps, EventCardState> {
```
estherkim, line 33: Can this component be renamed to something specific to Releases? All the components are named that way, but `EventCard` is more FullCalendar-esque.

release-calendar/src/client/components/EventCard.tsx
```diff
@@ -0,0 +1,158 @@
...
+    this.apiService = new ApiService();
+  }
+  async componentDidMount(): Promise<void> {
+    const releaseDates: ReleaseDates = await this.apiService.getReleaseDates(
+      this.props.eventApi.title,
+    );
+    this.setState({releaseDates});
+  }
+
+  schedule = [
```
estherkim, line 50: nit: rename to `history` or `promotionHistory`

release-calendar/src/client/components/TooltipHook.tsx
```diff
@@ -0,0 +1,104 @@
...
+  view: View;
+}): void => {
+  const Content = (): JSX.Element => {
+    return <Hook event={arg.event}></Hook>;
+  };
+
+  ReactDOM.render(<Content />, arg.el);
+};
+
+export const Hook = (children: {event: EventApi}): JSX.Element => {
```
estherkim, line 40: Keep name of the component consistent with the name of the file.

release-calendar/src/client/components/TooltipHook.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+import {EventApi, View} from '@fullcalendar/core';
+import {EventCard} from './EventCard';
+import {usePopper} from 'react-popper';
+import React from 'react';
+import ReactDOM from 'react-dom';
+
+const TOOLTIP_HEIGHT = 300;
+
+export const Tooltip = (arg: {
```
estherkim, line 25: Move this to its own file. 1 component per file.

release-calendar/webpack.client.dev.config.js
```diff
@@ -17,6 +17,8 @@
 const path = require('path');
 const HtmlWebPackPlugin = require('html-webpack-plugin');
 const Dotenv = require('dotenv-webpack');
+const MomentTimezoneDataPlugin = require('moment-timezone-data-webpack-plugin');
+const webpack = require('webpack');
```
estherkim, line 21: Adjusting for timezone should not happen during bundle compilation, because it depends on who's doing the compiling. I don't think it this belongs in webpack.

release-calendar/src/client/components/TooltipHook.tsx
estherkim: Hmm, okay. We can leave this as is.

release-calendar/src/client/components/EventCard.tsx
ajwhatson: ah, ReleaseCard?

release-calendar/package.json
ajwhatson: 👍I thought I had accidentally duplicated. oop

estherkim: Oh weird, it is duplicated. Maybe merge master?

release-calendar/webpack.client.dev.config.js
ajwhatson: Note for later: we talked offline. We decided to keep them. Cite this for reasons for the empty module https://github.com/moment/moment/issues/2979

release-calendar/src/client/components/EventCard.tsx
estherkim: that works

release-calendar/package.json
```diff
@@ -68,13 +72,15 @@
     "webpack-node-externals": "1.7.2"
   },
   "devDependencies": {
+    "empty-module": "0.0.2",
```
estherkim, line 75: Move this and moment-timezone-data-webpack-plugin to dependencies. It's needed there along with other webpack packages for deployment.

release-calendar/src/client/models/view-models.ts
```diff
@@ -56,3 +56,13 @@ export class CurrentReleases {
...
+
+export class ReleaseDates {
+  constructor(release: Release) {
+    this.dates = release.promotions.map((promotion) => {
+      return {channel: promotion.channel, date: promotion.date};
+    });
+  }
+
+  dates: {channel: Channel; date: Date}[];
+}
```
estherkim, line 68: Why do we need a new class when it's just paring down `Promotion`?

Another note, `dates` to be of type `{Channel, Date}[]` is misleading; you would think it would be an array of `Dates`. A more clear class would have Channel and Date members, then you could have a `const releaseDates = ReleaseDate[]`.

But to my point above, this is just a pared down version of `Promotion`

release-calendar/src/client/components/Tooltip.tsx
```diff
@@ -0,0 +1,35 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {EventApi, View} from '@fullcalendar/core';
+import {Hook} from './Hook';
+import React from 'react';
+import ReactDOM from 'react-dom';
+
+export const Tooltip = (arg: {
```
estherkim, line 22: The way this component is created looks different than other components - i.e. why is this a const instead of a class of type `React.Component`?

release-calendar/src/client/components/ReleaseCard.tsx
```diff
@@ -0,0 +1,184 @@
...
+      emojiName: 'house',
+    },
+    {
+      channel: Channel.LTS,
+      emoji: '🏙️',
+      emojiName: 'city',
+    },
+  ];
+
+  getTitle(search: Channel): string {
```
estherkim, line 82: What if `history` was a dict with Channel as the key? Could be a lot cleaner here.

release-calendar/src/client/components/ReleaseCard.tsx
```diff
@@ -0,0 +1,184 @@
...
+                  );
+                  return (
+                    match && (
+                      <div className={'content-row'} key={match.emoji}>
+                        <div className={'emoji'}>
+                          <span role='img' aria-label={'row.emojiName'}>
+                            {match.emoji}
+                          </span>
+                        </div>
+                        <div className={'text'}>
```
estherkim, line 165: Do you need the curly brackets or can it be `className='text'`? Same question for the rest of this PR.

release-calendar/src/client/components/Tooltip.tsx
```diff
@@ -0,0 +1,35 @@
...
+  isMirror: boolean;
+  isStart: boolean;
+  isEnd: boolean;
+  event: EventApi;
+  el: HTMLElement;
+  view: View;
```
estherkim, line 28: Where are the parameters besides `el` and `event` being used? Are they needed?

release-calendar/src/client/stylesheets/_base.scss
```diff
@@ -1,16 +1,20 @@
...
+$perc-beta: rgb(134, 62, 216);
+$perc-experimental: rgb(100, 204, 135);
+$opt-in-beta: rgb(200, 125, 207);
 $opt-in-experimental: SteelBlue;
-$nightly: Tomato;
+$nightly: rgb(209, 36, 36);
+$amp-blue: rgb(33, 150, 243);
+$amp-off-blue: rgba(23, 93, 151, 0.596);
+$amp-yellow: #fff198d0;
+$text-color: #24292e;
```
estherkim, line 12: Make these values consistent as much as you can; either use variables, rgb, rbga, or hex

release-calendar/src/client/components/Tooltip.tsx
ajwhatson: Because it's not really a component in itself. It's replacing the existing component events from the calendar with the component I build in the hook. https://reactjs.org/docs/react-dom.html

release-calendar/src/client/components/ReleaseCard.tsx
ajwhatson: No, I don't. I'll get clean it up

release-calendar/src/client/components/Tooltip.tsx
ajwhatson: They're needed because that's what the method eventRender is expecting in FullCalendar. It wont accept any other.

release-calendar/src/client/stylesheets/_base.scss
ajwhatson: This will be automatically fixed when the scss linter is added, but I manually did it for now.

ajwhatson:
getReleaseDates() will change to getRelease() once the current getRelease is changed in a different pr

release-calendar/src/client/components/ReleaseCard.tsx
ajwhatson: The best solution I can think to debug the channels with "-" not being accessible within the dictionary is to create an interface where the key is typed as a string. This will push the enum into a string under the hood, but it allows for normal dictionary sets and gets. If you are okay with here, then I could also make the channelsTitles a dictionary, which is currently a {channel: Channel, title: string}[]. We had bad it this because of the same weird "-" enum behavior. If I switched it, then I'd again have direct access to the titles I want instead of doing the find() loop I do now.

release-calendar/package.json
```diff
@@ -65,7 +69,9 @@
     "webpack": "4.43.0",
     "webpack-cli": "3.3.11",
     "webpack-dev-server": "3.10.3",
-    "webpack-node-externals": "1.7.2"
+    "webpack-node-externals": "1.7.2",
+    "moment-timezone-data-webpack-plugin": "1.3.0",
+    "empty-module": "0.0.2"
```
estherkim, line 74: Alphabetize

release-calendar/src/client/api-service.ts
```diff
@@ -61,4 +61,10 @@ export class ApiService {
     );
     return new CurrentReleases(currentReleases);
   }
+
+  async getReleaseDates(requestedRelease: string): Promise<Release> {
```
estherkim, line 65: rename `getReleaseDates` because it's not getting `releaseDates` anymore.

release-calendar/src/client/components/ReleaseCard.tsx
```diff
@@ -0,0 +1,174 @@
...
+      emojiName: 'experiment',
+    },
+    [Channel.STABLE]: {
+      title: channelTitles.find((channel) => channel.channel == Channel.NIGHTLY)
+        .title,
+      emoji: '🏠',
+      emojiName: 'house',
+    },
+    [Channel.LTS]: {
+      title: channelTitles.find((channel) => channel.channel == Channel.NIGHTLY)
```
estherkim, line 94: Switch out Channel.NIGHTLY

release-calendar/src/client/components/ReleaseCard.tsx
```diff
@@ -0,0 +1,174 @@
...
+        ).title
+      }, ${
+        channelTitles.find((channel) => channel.channel == Channel.PERCENT_BETA)
+          .title
+      }`,
+      emoji: '🧪',
+      emojiName: 'experiment',
+    },
+    [Channel.STABLE]: {
+      title: channelTitles.find((channel) => channel.channel == Channel.NIGHTLY)
```
estherkim, line 88: Switch out Channel.NIGHTLY

release-calendar/src/client/components/ReleaseCard.tsx
```diff
@@ -0,0 +1,174 @@
...
+        channelTitles.find(
+          (channel) => channel.channel == Channel.PERCENT_EXPERIMENTAL,
+        ).title
+      }, ${
+        channelTitles.find((channel) => channel.channel == Channel.PERCENT_BETA)
+          .title
+      }`,
+      emoji: '🧪',
+      emojiName: 'experiment',
+    },
```
estherkim, line 86: This assumes that a release is promoted to perc-beta at the same time as perc-experimental, but let's treat this as separate channels and have separate lines.

Maybe in the future we'll consolidate these channels into one. But for now let's stay consistent throughout the app.

release-calendar/src/client/components/ReleaseCard.tsx
```diff
@@ -0,0 +1,174 @@
...
+    [Channel.OPT_IN_EXPERIMENTAL]: {
+      title: `${
+        channelTitles.find(
+          (channel) => channel.channel == Channel.OPT_IN_EXPERIMENTAL,
+        ).title
+      }, ${
+        channelTitles.find((channel) => channel.channel == Channel.OPT_IN_BETA)
+          .title
+      }`,
```
estherkim, line 71: Let's keep these separate, more explanation in another comment.

release-calendar/src/client/components/ReleaseCard.tsx
estherkim: Can we revisit the dict idea? Hyphens should not be a problem. For example
```
const channelMap = {
    'lts': 'lts',
    'stable': 'stable',
    'opt-in-beta': 'opt in beta',
    'perc-beta': '1% beta',
};

console.log(channels['opt-in-beta']);
// opt in beta

console.log(channels['perc-beta']);
// 1% beta
```

release-calendar/src/client/components/Tooltip.tsx
estherkim: I see. The docs say "Most of your components should not need to use this module." hmmm... we can punt this for now.

release-calendar/src/client/components/Hook.tsx
```diff
@@ -0,0 +1,95 @@
...
+ * distributed under the License is distributed on an "AS-IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {ReleaseCard} from './ReleaseCard';
+import {usePopper} from 'react-popper';
+import React from 'react';
+import ReactDOM from 'react-dom';
```
estherkim, line 20: Since React is loaded from a script tag, we won't need to import this, according to https://reactjs.org/docs/react-dom.html

release-calendar/src/client/stylesheets/_base.scss
estherkim: It's still inconsistent?

release-calendar/src/client/api-service.ts
estherkim: Oh, I just saw your comment. Okay.

release-calendar/src/client/components/ReleaseCard.tsx
estherkim: Making channelsTitles a dictionary is a great idea :)

estherkim: Also I totally misread your comment! My "revisit the dict idea" was not helpful haha. Nice workaround.

release-calendar/src/client/api-service.ts
ajwhatson: :)

release-calendar/src/client/components/ReleaseCard.tsx
ajwhatson: oop

release-calendar/src/client/components/Hook.tsx
ajwhatson: I'm confused. What should i do instead then?

release-calendar/src/client/stylesheets/_base.scss
ajwhatson: no, it should be good!

release-calendar/src/client/components/Hook.tsx
ajwhatson: I get the warning "'ReactDOM' refers to a UMD global, but the current file is a module. Consider adding an import instead." when I just delete the import line. So yes, it is a global but doesn't that still mean we have to import it?

estherkim: Hmm okay not sure what the warning means. Not a big deal though, resolving.

PR #814, opened 2020-04-30
Define root ES6 and TS eslint configs

rcebulko:
Thanks to Raghu pointing out what `root: true` was doing in the config, combined with finding out that eslint configs can explicitly set `extends` to include other configs, this got a lot cleaner than #810 .

This PR:
- defines root eslint configs for ES6 and TS
- makes the root default config ES6
- updates the eslint config for `onduty`

Once this goes in, I'll follow up asynchronously with PRs to convert other apps.

Part of https://github.com/ampproject/amp-github-apps/issues/809


rcebulko:
> Honestly I thought that `root: true` just means "use the root configuration"

Yeah with a lot of configs it's so hard to keep track of all the pieces eventually it becomes "trust what's in the other directory" :laughing:

rsimha:
👏👏👏

PR #821, opened 2020-05-01
[release-calendar] Calendar view date

ajwhatson:
This will navigate the calendar to a relevant date when a release is selected.

Undesirable behavior: Let's say a user went to select a release, but was viewing the calendar in a different month than one where the requested release had been promoted in. Then, when they select to view the release, the calendar could appear empty.

Fix: Change the calendar to display the month of the latest release promotion when a release is selected. Also, when a release is unselected, return the viewer to the current month.

Included: 
- reference to the calendar element, `referenceCalendar`, in order to access the `FullCalendar` method `gotoDate()`, https://fullcalendar.io/docs/Calendar-gotoDate.
- creation of my own `gotoDate(date: Date)` which uses the `referenceCalendar` to change the calendar view date
- use of `gotoDate()` when the `Calendar` component `singleRelease` state is changed
- very unrelated to anything else, but the binding of `handleReleaseClick()` in `ChannelTable` that I should have done long ago

### DIFF
--- a/release-calendar/src/client/components/Calendar.tsx
+++ b/release-calendar/src/client/components/Calendar.tsx
@@ -56,6 +56,8 @@ export class Calendar extends React.Component<CalendarProps, CalendarState> {
     this.apiService = new ApiService();
   }
 
+  calendarReference = React.createRef<FullCalendar>();
+
   async componentDidMount(): Promise<void> {
     const releases = await this.apiService.getReleases();
     this.setState({allEvents: getAllReleasesEvents(releases)});
@@ -68,11 +70,18 @@ export class Calendar extends React.Component<CalendarProps, CalendarState> {
           this.props.singleRelease,
         );
         this.setState({singleEvents: getSingleReleaseEvents(release)});
+        this.gotoDate(release[0].start);
       } else {
         this.setState({singleEvents: []});
+        this.gotoDate(new Date());
       }
     }
   }
+
+  gotoDate = (date: Date): void => {
+    const calendarApi = this.calendarReference.current.getApi();
+    calendarApi.gotoDate(date);
+  };
 
   tooltip = (arg: {
     isMirror: boolean;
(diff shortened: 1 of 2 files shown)


release-calendar/src/client/components/Calendar.tsx
```diff
@@ -68,12 +70,19 @@ export class Calendar extends React.Component<CalendarProps, CalendarState> {
           this.props.singleRelease,
         );
         this.setState({singleEvents: getSingleReleaseEvents(release)});
+        this.gotoDate(release[0].start);
       } else {
         this.setState({singleEvents: []});
+        this.gotoDate(new Date());
```
estherkim, line 76: Curious, does Full Calendar have a "go to Today" button option? Going back to the current date when you remove filters might be unexpected, so I'm not sure we should do this. But having a "go to Today" button would help!

release-calendar/src/client/components/Calendar.tsx
```diff
@@ -68,12 +70,18 @@ export class Calendar extends React.Component<CalendarProps, CalendarState> {
...
         );
         this.setState({singleEvents: getSingleReleaseEvents(release)});
+        this.gotoDate(release[0].start);
       } else {
         this.setState({singleEvents: []});
       }
     }
   }
 
+  gotoDate = (date: Date): void => {
```
danielrozenberg, line 80: Is there a reason why you use an anonymous function here and not a regular function syntax? (`gotoDate(date: Date): void {`)

Anonymous functions are more useful when you use them as, say, an argument to another function (e.g., `arr.sort((firstEl, secondEl) => { ... })`)

ajwhatson: No there wasn't any reason. I just wasn't aware one was the "regular", and I've gotten used to the anonymous way since I use functions as arguments more often. I'll switch this one to the regular function syntax!

PR #826, opened 2020-05-01
Simplify build targets for apps

rcebulko:


build-system/build-targets.js
```diff
@@ -23,59 +23,25 @@
...
-  'BUNDLE_SIZE',
-  'CHECKLIST',
-  'ERROR_ISSUE',
-  'INVITE',
-  'ONDUTY',
-  'OWNERS',
-  'PR_DEPLOY',
-  'RELEASE_CALENDAR',
-  'TEST_STATUS',
+const APPS_TO_TEST = [
```
rcebulko, line 26: When I first added built targets, we discussed keeping them functionally separate from the apps. In practice, having 3 places to update lead to apps being added without the tests running in CI, and we don't have any non-app build targets. If we do in the future that can be added easily, but until then, this is a much cleaner solution with less room for error.

rsimha: Right, we discussed this in https://github.com/ampproject/amp-github-apps/pull/606#commitcomment-35817975. Clearly, the concern I raised there hasn't come true over all these months, so +1 to your new plan 😃

PR #828, opened 2020-05-02
[release-calendar] 🐛Single release event bug 🐛

ajwhatson:
This fixes the bug in how events are displayed when called from a single requested release. Before there the `end` date was incorrectly assumed to be when that same release was promoted to a new channel. This is incorrect because of a number of reasons. For example, a release can live in a more than one channel at a time.

In addition, it optimizes `componentDidUpdate` so that `singleEvents` that lives within the `calendar` state does not update when a user requests to view a single release, clicks away (either unselects or switches to channel view), and then re requests that same single release.

In addition, this changes the way getReleases() was assigning the current date. Instead of new Date() being passed into the `ReleaseEventInput` constructor. The data is optional. Whenever the date is not added within the constructor the current date is assigned. 

Included: (Fixes only occurred on the client side)

- `getSinglePromotions` to replace the old `getRelease`. was renamed to match the change in the return type from `ReleaseEventInput[]` to `Promotion[]`
- Adaption of `ReleaseEventInput` to allow for an optional date to be passed to it. If no date, then the end date will be set to the current time.
- rename of `getReleaseDates` to `getRelease` as mentioned in #813 
- Optimization of `componentDidUpdate` as mentioned above


release-calendar/src/client/models/release-event.ts
```diff
@@ -15,19 +15,20 @@
  */
 
 import {Channel} from '../../types';
-import {EventSourceInput} from '@fullcalendar/core/structs/event-source';
-import {ReleaseEventInput} from './view-models';
+import {EventInput} from '@fullcalendar/core/structs/event';
+import {ExtendedEventSourceInput} from '@fullcalendar/core/structs/event-source';
```
estherkim, line 19: If `ExtendedEventSourceInput` isn't required, I would opt for using the smaller, more simple `EventInput`.

release-calendar/src/client/models/view-models.ts
```diff
@@ -56,3 +56,13 @@ export class CurrentReleases {
 
   map: Map<Channel, string>;
 }
+
+export class SingleReleasePromotion {
```
estherkim, line 60: What is the point of this class? Can we use `Promotion` since we have it already?

release-calendar/src/client/models/release-event.ts
```diff
@@ -38,11 +39,19 @@ export function getAllReleasesEvents(
...
+      events: [
+        (allReleases.get(promotion.channel).events as EventInput[]).find(
+          (event) => {
+            return event.start == promotion.date;
+          },
+        ),
+      ],
+      textColor: 'white',
+    };
+  });
```
estherkim, line 56: It's hard to follow what's going on here. What is this doing?

release-calendar/src/client/api-service.ts
```diff
@@ -28,7 +32,10 @@ export class ApiService {
     return fetch(url).then((result) => result.json());
   }
 
-  async getRelease(requestedRelease: string): Promise<ReleaseEventInput[]> {
+  //TODO: specialize this for use in EventCard
+  async getEventSchedule(
```
estherkim, line 36: Remove this until it is being used.

release-calendar/src/client/components/Calendar.tsx
```diff
@@ -62,15 +62,20 @@ export class Calendar extends React.Component<CalendarProps, CalendarState> {
...
+    ) {
+      const release = await this.apiService.getRelease(
+        this.props.singleRelease,
+      );
+      this.setState({
+        singleEvents: {
+          releaseName: this.props.singleRelease,
+          events: getSingleReleaseEvents(release, this.state.allEvents),
+        },
+      });
```
estherkim, line 78: I don't think we should be saving `singleEvents` when a user clicks on another filter because this muddles the separation of `calendarProps`, which has filter criteria, and `state`, which has events. We should keep these separate.

release-calendar/src/client/models/release-event.ts
ajwhatson: In order to use the get methods, I have to have it as ExtendedEventSourceInput.
```
allReleases.get(promotion.channel) as ExtendedEventSourceInput)
          .events
```
This wouldn't work in `getSingelReleaseEvents`

release-calendar/src/client/components/Calendar.tsx
ajwhatson: I'm happy to change it back but can you explain this more? what you mean when you say filter criteria

release-calendar/src/client/models/release-event.ts
estherkim: Ooh I see! Instead of filtering twice, this can be simplified to 
```
return event.start == promotion.date &&
  event.channel == promotion.channel &&
  event.releaseName == promotion.releaseName
```

Unsure if those are the right property names but you get the gist!

release-calendar/src/client/components/Calendar.tsx
estherkim: By filter criteria I mean - what channels are selected, if any? which release is selected, if any?

Because `state` in this PR includes a selected release, it's possible that it does not match `calendarProps`, which could lead to bugs in the future. (A developer could easily assume that the calendar should show whatever is in `state`, but in this case it could be stale). We should let `calendarProps` drive the content of `state`, not the other way around.

ajwhatson: oh, I see, that makes sense.

release-calendar/src/client/models/release-event.ts
ajwhatson: Update: adjustment was made to getSingleReleaseEvents that removed the need for the events on the EventSourceInput to be accessed.

release-calendar/src/client/components/Calendar.tsx
```diff
@@ -88,7 +98,7 @@ export class Calendar extends React.Component<CalendarProps, CalendarState> {
           right: 'dayGridMonth,timeGridWeek,listWeek',
         }}
         plugins={[dayGridPlugin, timeGridPlugin]}
-        eventSources={displayEvents}
+        eventSources={[{events: displayEvents, textColor: 'white'}]}
```
ajwhatson, line 101: Because of this, where I put the `ReleaseEventInput[]` into a `EventSourceInput`, there isn't a need for `EventSourceInput` to be imported in any of my file.

release-calendar/src/client/components/Calendar.tsx
```diff
@@ -47,37 +44,50 @@ export class Calendar extends React.Component<CalendarProps, CalendarState> {
...
+      displayEvents = this.state.singleEvents;
+    } else {
+      this.props.channels
+        .filter((channel) => this.state.allEvents.has(channel))
+        .forEach(
+          (channel) =>
+            (displayEvents = [
+              ...displayEvents,
+              ...this.state.allEvents.get(channel),
+            ]),
```
danielrozenberg, line 88: When creating an arrow with a single expression not inside a block (e.g., `() => expr` instead of `() => { expr; }`) it's equivalent to having a block  that returns that single expression. That is...
```ts
// this:
(channel) => someFoo(channel);

// is basically this:
(channel) => {
  return someFoo(channel);
}
```

This means that the arrow function here is equivalent to:
```ts
(channel) => {
  return (displayEvents = [
    ...displayEvents,
    ...this.state.allEvents.get(channel),
  ]);
}
```

and that's a little confusing, because it makes the reader wonder why you're returning an assignment expression ("`return (x = y)`") to a function that doesn't take arguments (`forEach` doesn't act on return values from its callback function)

In this case, it makes better sense to do:
```ts
(channel) => {
  displayEvents = [
    ...displayEvents,
    ...this.state.allEvents.get(channel),
  ];
}
```

Without a return statement. An even more concise version would be:
```ts
(channel) => {
  displayEvents.push(...this.state.allEvents.get(channel));
}
```

release-calendar/src/client/models/release-event.ts
```diff
@@ -14,35 +14,29 @@
...
+export function getAllEvents(
   events: ReleaseEventInput[],
-): Map<Channel, EventSourceInput> {
+): Map<Channel, ReleaseEventInput[]> {
   const eventInputs = new Map<Channel, ReleaseEventInput[]>();
   events.forEach((event) => {
     const channelEvents = eventInputs.get(event.channel) || [];
     eventInputs.set(event.channel, [...channelEvents, event]);
   });
+  console.log(eventInputs);
```
danielrozenberg, line 28: Debugging leftover?

ajwhatson: Drats, a mistake I make until the very end. :(

danielrozenberg: Haha don't worry, literally everyone does that all the time no matter how much experience we have

release-calendar/src/client/components/Calendar.tsx
estherkim: Great comment!

release-calendar/src/client/components/Calendar.tsx
estherkim: SG

release-calendar/src/client/components/Calendar.tsx
ajwhatson: Yes! Thanks, for explaining :)

PR #829, opened 2020-05-04
[release-calendar] Adding SearchBar

ajwhatson:
This is the addition of an input protected search bar. The field will only allow for input to be submitted when the string is 13 characters long and only consists of digits. Once a 13 digit number is entered it will be checked if it exists in the database, and then be passed back up the `App`

This is what the search bar looks like when the input is valid but the release is not found within the database.
Screen Shot 2020-05-03 at 10 47 01 PM

This is what the search bar looks like when the input is not valid
Screen Shot 2020-05-03 at 11 08 11 PM

Included: 
- Installation of @material-ui/core (for the TextField), @material-ui/icons (for the search icon that is with the TextFeild), and react-hook-form (which helps with checking for errors and minimizes rerenders)
- Protection of invalid input including non digits strings, empty inputs, and strings not 13 characters long
- Addition of `countRelease` on the client side which will be used on the server side in partnership with `isRelease` to check is the requested release name is really among the past releases.
- Adjustments to the styling in other components to provide space for the SearchBar
- Creation of styling for SearchBar
- Addition of hopefully helpful error messaging when input is invalid


release-calendar/src/server/repository-service.ts
```diff
@@ -28,6 +28,13 @@ export class RepositoryService {
     this.promotionRepository = connection.getRepository(PromotionEntity);
   }
 
+  countRelease(name: string): Promise<number> {
```
estherkim, line 31: Since `release.name` is a primary key (must be unique). This will only return a 1 or 0.

release-calendar/src/client/api-service.ts
```diff
@@ -43,6 +47,13 @@ export class ApiService {
     ];
   }
 
+  async isRelease(requestedRelease: string): Promise<boolean> {
```
estherkim, line 50: Why not reuse `getRelease` and see if it turns up empty?

release-calendar/src/client/components/SearchBar.tsx
```diff
@@ -0,0 +1,112 @@
...
+      validateCriteriaMode: 'all',
+    });
+    const onSubmit = handleSubmit(async ({release}) => {
+      if (await this.apiService.isRelease(release)) {
+        this.props.handleSelectedRelease(release);
+        this.setState({
+          badSearch: null,
+        });
+      } else {
+        this.setState({badSearch: `Release ${release} is not found 🤭`});
```
estherkim, line 61: nit: `was` not found

release-calendar/src/client/api-service.ts
ajwhatson: I had thought that counting is better for perfomance. Is that not correct?

estherkim: It's arguable, because you're adding `getCount()` then seeing if `count > 0`.

release-calendar/src/server/repository-service.ts
```diff
@@ -28,6 +28,10 @@ export class RepositoryService {
     this.promotionRepository = connection.getRepository(PromotionEntity);
   }
 
+  getReleaseNames(): Promise<Release[]> {
+    return this.releaseRepository.createQueryBuilder('release').getMany();
+  }
```
estherkim, line 33: Rename this to `getReleases()` and the existing `getReleases()` to `getPromotions()`. This includes changes to the api route and the api service.

We should probably rename `getCurrentReleases()` to `getCurrentPromotions()` too.

release-calendar/src/client/components/SearchBar.tsx
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  isErrorDisplayed(input: string): boolean {
+    if (input?.length > 13) {
+      return true;
+    }
+    return !(input == null || /^[0-9]*$/.test(input));
+  }
+
+  render(): JSX.Element {
```
estherkim, line 109: What happened to the magnifying glass button?

release-calendar/src/client/components/SearchBar.tsx
```diff
@@ -0,0 +1,138 @@
...
+    return (
+      <React.Fragment>
+        <h1 className='title-bar'>Release Search</h1>
```
estherkim, line 112: I think it will look cleaner without this title

release-calendar/src/client/components/SearchBar.tsx
```diff
@@ -0,0 +1,138 @@
...
+        <form onSubmit={this.onSubmit}>
+          <Autocomplete
+            freeSolo
+            id='release-autocomplete'
+            onClose={this.onClose}
+            onChange={this.onChange}
+            onInputChange={this.onInputChange}
+            options={this.state.releaseNames}
```
estherkim, line 120: Is there a limit of releases that go in this drop down? What happens if you have 500 releases?

This concern applies to the drop down whether you have nothing in the search bar, or if it shows results. The dropdown shouldn't be huge.

release-calendar/src/client/components/SearchBar.tsx
```diff
@@ -0,0 +1,138 @@
...
+      if (input != null) {
+        return `${input.length}/13`;
+      } else {
+        return 'Search for a release...';
+      }
+    } else {
+      if (this.state.validSearch) {
+        return 'Found! 👀';
+      } else {
+        return 'Error! 🙈';
```
estherkim, line 97: Instead of error, say "Not found"

release-calendar/src/client/components/SearchBar.tsx
```diff
@@ -0,0 +1,138 @@
...
+  async componentDidMount(): Promise<void> {
+    const releaseNames = await this.apiService.getReleaseNames();
+    this.setState({releaseNames});
+  }
+
+  onSubmit(event: React.FormEvent<HTMLFormElement>): void {
+    event.preventDefault();
+  }
+
+  onClose(event: React.ChangeEvent<{}>, reason: string): void {
```
estherkim, line 58: When the search bar is cleared, the calendar still filters on the previous release.

release-calendar/src/client/components/SearchBar.tsx
```diff
@@ -0,0 +1,138 @@
...
+  labelText(input: string): string {
+    if (this.state.validSearch == null) {
+      if (input != null) {
+        return `${input.length}/13`;
+      } else {
+        return 'Search for a release...';
+      }
+    } else {
+      if (this.state.validSearch) {
+        return 'Found! 👀';
```
estherkim, line 95: How about "Selected ✅" so that it's clear that you're filtering it on the calendar? I think the search bar makes it clear whether a release was found.

release-calendar/src/client/components/SearchBar.tsx
```diff
@@ -0,0 +1,138 @@
...
+      this.setState({input});
+    }
+    if (this.state.validSearch != null) {
+      this.setState({validSearch: null});
+    }
+  }
+
+  labelText(input: string): string {
+    if (this.state.validSearch == null) {
+      if (input != null) {
```
estherkim, line 88: When the 'x' button is clicked, this says "0/13" which is fine. But when you click away from the search bar, the placeholder says "0/13" instead of "Search for a release..." which looks a little weird.

release-calendar/src/client/components/SearchBar.tsx
ajwhatson: It limits it to 9. I checked.

release-calendar/src/client/components/SearchBar.tsx
ajwhatson: The easiest way to change this behavior is to remove the option for "0/13" otherwise I would have to add eventlisteners to the the rest of the app on mouse down. I could but that as a future todo on the todo list?

release-calendar/src/client/components/SearchBar.tsx
ajwhatson: I'll put it back. 👍

estherkim:
Can you delete your package-lock.json and regenerate it? Somehow it's throwing an error.

release-calendar/src/client/components/SearchBar.tsx
estherkim: This problem still exists.

release-calendar/src/client/stylesheets/channelTable.scss
```diff
@@ -4,7 +4,7 @@
     display: flex;
     justify-content: space-between;
     align-items: center;
-    padding-top: 0px;
+    margin-top: 0px;
```
danielrozenberg, line 7: Super-nit: zero-values don't need a unit (`0` instead of `0px`)

release-calendar/src/client/components/SearchBar.tsx
```diff
@@ -0,0 +1,151 @@
...
+        return 'Not Found! 🙈';
+      }
+    }
+  }
+
+  isErrorDisplayed(input: string): boolean {
+    if (input?.length > 13) {
+      return true;
+    }
+    return !(input == null || /^[0-9]*$/.test(input));
```
danielrozenberg, line 114: Super-nit: `/^\d*$/`

PR #830, opened 2020-05-04
[release-calendar] Adding stylelint

ajwhatson:
This is supposed to add stylelint to format scss and css files.

Hi @estherkim, I'm going to ask you about this tomorrow morning.


estherkim:
Hiya, instead of having a separate script, let's use the stylelint prettier plugin so we can "prettify" all files in the same process.

https://github.com/prettier/stylelint-prettier

PR #835, opened 2020-05-04
[error-issue] Report errors in private repo and provide utility to transfer them

rcebulko:


rcebulko:
Closing since it seems like the GitHub permissions model won't allow a repo to be private to org members

PR #837, opened 2020-05-05
[error-issue] Include list of versions an error has been seen in for issue

rcebulko:


error-issue/index.ts
```diff
@@ -61,21 +62,18 @@ export async function errorIssue(
   const shouldLinkIssue = linkIssue === '1';
 
   let parsedReport: ErrorReport;
-  if (firstSeen && dailyOccurrences && stacktrace) {
+  if (firstSeen && dailyOccurrences && stacktrace && seenInVersions) {
```
danielrozenberg, line 65: I feel like with so many properties here to validate, how about making a static method on ErrorReport that takes the unstructured data and returns a Promise that resolves to a built ErrorReport, or rejects if it fails

This way the code would look something like:
```ts
const parsedReport = await ErrorReport.fromJSON(errorReport).catch(() => {
  // what's inside the else clause
  return monitor.reportFromGroupStatus(groupStats);
});
```

rcebulko: Quick note, `ErrorReport` is a type interface rather than a class, so this would instead be a method in `index.ts`. I split this out into a helper function that's not async and doesn't need to be, so I'm just catching the error and handling it in the catch (which works out better with the early-return redirects)

danielrozenberg: Also good :)

PR #840, opened 2020-05-06
[error-issue] Provide error listing for each service group

rcebulko:
Hopefully the last major update to this backend before building out a simple UI to consume it.

This PR:
- adds a Stackdriver API method to fetch errors from service groups (production, development, nightly, and experiments)
- defines a subclass of `ErrorMonitor` which filters to a service, scaling the frequency threshold based on the diversion % and throttling settings for errors
- adds an optional `serviceType` query param
- adds related service type info to the JSON response


rcebulko:
[removed Daniel per his request due to PR overload :laughing: ]

error-issue/src/service_error_monitor.ts
```diff
@@ -0,0 +1,89 @@
...
+ * limitations under the License.
+ */
+
+import {ErrorMonitor} from './error_monitor';
+import {ServiceGroup, Stackdriver} from 'error-issue-bot';
+import {StackdriverApi} from './stackdriver_api';
+
+export enum ServiceName {
+  PRODUCTION = 'CDN Production',
+  DEVELOPMENT = 'CDN Development',
```
estherkim, line 23: Can these be renamed to the channel names? i.e. STABLE, BETA

error-issue/src/service_error_monitor.ts
```diff
@@ -0,0 +1,89 @@
...
+  protected async newErrors(): Promise<Array<Stackdriver.ErrorGroupStats>> {
+    return this.client.listServiceGroups(this.serviceName, this.pageLimit);
+  }
+}
+
+declare module './error_monitor' {
+  interface ErrorMonitor {
+    service(serviceName: ServiceName): ServiceErrorMonitor;
+  }
+}
```
estherkim, line 75: I don't understand what's going on here. You have a file called `error_monitor` but you also declare a module called `./error_monitor`?

error-issue/src/service_error_monitor.ts
```diff
@@ -0,0 +1,89 @@
...
+
+/**
+ * Set of service groups to monitor and details for scaling frequency.
+ * Note: These values do not need to be exact; the order-of-magnitude is what's
+ * important here.
+ */
+const SERVICE_GROUPS: Record<ServiceName, ServiceGroup> = {
+  'CDN Production': {diversionPercent: 0.98, throttleRate: 0.1},
+  'CDN Development': {diversionPercent: 0.01, throttleRate: 1},
+  'CDN Experiments': {diversionPercent: 0.01, throttleRate: 1},
```
estherkim, line 36: Just so I understand, do both the 1% and opt-in channels fall in these groups?

error-issue/index.ts
```diff
@@ -27,6 +28,7 @@ const GITHUB_REPO = process.env.GITHUB_REPO || 'ampproject/amphtml';
 const [GITHUB_REPO_OWNER, GITHUB_REPO_NAME] = GITHUB_REPO.split('/');
 const GITHUB_ACCESS_TOKEN = process.env.GITHUB_ACCESS_TOKEN;
 const PROJECT_ID = process.env.PROJECT_ID || 'amp-error-reporting';
+const MIN_FREQUENCY = Number(process.env.MIN_FREQUENCY || 2500);
```
estherkim, line 31: Default is 2500 here but `ErrorMonitor` defaults to 5000 in the constructor. I would either remove the default value in the constructor, or make it the same value.

error-issue/src/service_error_monitor.ts
rcebulko: These are larger groupings than just the channels. "CDN Production" includes LTS, Stable, and Control channels. "CDN Development" includes Beta and Experimental. "CDN Experiments" includes Experiment-[ABC]. The idea here is that, when looking at and measuring frequency of errors, we can group all versions that are running the same commit of the code. In other words, they correspond to the Service bucket (first dropdown), not the actual Version (second dropdown)

error-issue/src/service_error_monitor.ts
rcebulko: Yes; for the purposes of error-reporting, opt-in is (and was previously) treated the same as 1%. A useful reference here (and I can include a link in the comment, if you think it wise) is the [error-tracker code](https://github.com/ampproject/error-tracker/blob/master/utils/rtv/release-channels.js) that maps from RTV prefixes to a service group and version name. Replicating below for convenience:

```javascript
module.exports = {
  '00': { group: 'Development', name: 'Experimental' },
  '01': { group: 'Production', name: 'Stable' },
  '02': { group: 'Production', name: 'Control' },
  '03': { group: 'Development', name: 'Beta' },
  '04': { group: 'Nightly', name: 'Nightly' },
  '05': { group: 'Nightly', name: 'Nightly-Control' },
  '10': { group: 'Experiments', name: 'Experiment-A' },
  '11': { group: 'Experiments', name: 'Experiment-B' },
  '12': { group: 'Experiments', name: 'Experiment-C' },
  // Ads error reporting will get all of the below channels, so the service
  // bucket names can be more verbose.
  '20': { group: 'Inabox-Control-A', name: 'Inabox-Control-A' },
  '21': { group: 'Inabox-Experiment-A', name: 'Inabox-Experiment-A' },
  '22': { group: 'Inabox-Control-B', name: 'Inabox-Control-B' },
  '23': { group: 'Inabox-Experiment-B', name: 'Inabox-Experiment-B' },
  '24': { group: 'Inabox-Control-C', name: 'Inabox-Control-C' },
  '25': { group: 'Inabox-Experiment-C', name: 'Inabox-Experiment-C' },
};
```

error-issue/src/service_error_monitor.ts
rcebulko: When `ErrorMonitor` is being defined, I can't add a method with a function signature returning the `ServiceErrorMonitor` subclass because of circular dependency issues. This was the recommended workaround I found to handle adding a method to a class after the fact. The alternative would be merging both class definitions into one file, so there is no circular `import` issues. WDYT?

error-issue/index.ts
rcebulko: Will remove.

error-issue/src/service_error_monitor.ts
rcebulko: I shifted the `ServiceErrorMonitor` class into the same file as `ErrorMonitor`, so this workaround was no longer needed.

error-issue/src/service_error_monitor.ts
estherkim: Ahh I see, thanks!

error-issue/src/service_error_monitor.ts
estherkim: Nice!

PR #851, opened 2020-05-08
Linkify stacktraces in created issues

rcebulko:


rcebulko:
Just realized #850 is a PR not an FR :man_facepalming:

PR #858, opened 2020-05-13
[release-calendar] Add endpoint to insert release data into database

estherkim:
It will be called by the versioning subscriber.


release-calendar/src/server/repository-service.ts
```diff
@@ -73,4 +73,16 @@ export class RepositoryService {
       }),
     );
   }
+
+  async createReleases(releases: Release[]): Promise<void> {
+    await this.releaseRepository.save(releases).catch((error) => {
+      throw error;
```
rcebulko, line 79: If an `await`-ed Promise rejects, I believe it will throw synchronously. Unless I'm misinformed, this `catch` won't do anything

release-calendar/cloud_build.yaml
```diff
@@ -49,3 +49,4 @@ secrets:
     secretEnv:
       DB_USERNAME: CiQAvF4DHWDrMTBZvbCxwRpkA1AT9k+fRDNvKVFKFmvb9WsBtRQSLwDYFKiUQ+96PiBOialv1MC1VhgctciMC5T5xkOJtURbURyN/O0SgGpRlHm8SGB+
       DB_PASSWORD: CiQAvF4DHVS/vkFxZQxF9AwHS0ELlfGYVSxPONmtfmVaK4PZx2ISLwDYFKiUJgUhlqEZRGNoPxcr27zigWXiZwibVrpLEcaklecTI48xCGvnmmkjwZkG
+      BASIC: CiQAvF4DHV54zziFx9vYRkcyjAl5gCLTTGHazTnKcrBD7n50OCwSRADYFKiUCZMdSYBgjTKHey0eMiZ3ou7npFPvsNI8+BzfDThhPhOSF0j0a3ZZRbgH/HqGu4L7Tl6of3Ltx0ZVgfBAEup0
```
rcebulko, line 52: Nit: `BASIC_AUTH_TOKEN` or something more informative could be helpful here

release-calendar/src/server/app.ts
```diff
@@ -68,6 +74,40 @@ async function main(): Promise<void> {
...
+
+      // parse ctime string into date
+      const {releases, promotions} = req.body;
+      promotions.forEach((p: {time: string; date: Date}) => {
+        p.date = moment(p.time, _CTIME_FORMAT).toDate();
+      });
+
+      const errors = {
+        createReleases: {},
+        createPromotions: {},
```
rcebulko, line 94: Does this have to be initialized, or could the `catch` handlers below add properties to an otherwise empty object? `errors.createReleases` and `errors.createPromotions` below will always be truthy (`Boolean({}) === true`)

release-calendar/src/server/app.ts
```diff
@@ -18,10 +18,14 @@ import {Connection, createConnection} from 'typeorm';
 import {RepositoryService} from './repository-service';
 import PromotionEntity from './entities/promotion';
 import ReleaseEntity from './entities/release';
+import bodyParser from 'body-parser';
 import express from 'express';
+import moment from 'moment';
 import path from 'path';
 import rateLimit from 'express-rate-limit';
 
+const _CTIME_FORMAT = 'ddd MMM D HH:mm:ss YYYY';
```
rcebulko, line 27: Is it possible to use a more standard format here, like a Unix timestamp or the UTC format? I'm guessing this is related to the format the versioning subscriber deals with, but it seems like that formatting could be done on the request side rather than adjusting the API to that particular format. WDYT?

release-calendar/src/server/repository-service.ts
estherkim: It's a TypeORM quirk - https://github.com/typeorm/typeorm/issues/5057

rcebulko: Gotcha, thanks for the reference! Maybe include a comment with this link?

release-calendar/src/server/app.ts
estherkim: Good catch. I turned errors into an array.

release-calendar/src/server/app.ts
estherkim: Good point about not adjusting the API side. Ok, going to use ISO format then.

rcebulko: SGTM

release-calendar/src/server/repository-service.ts
estherkim: Okiedokes

release-calendar/src/server/app.ts
```diff
@@ -68,6 +72,39 @@ async function main(): Promise<void> {
     app.get('/', async (req, res) => {
       res.sendFile(path.join(DIST_DIR, 'index.html'));
     });
+
+    app.post('/insert/', async (req, res) => {
+      // authorization
+      const authorization = req.header('authorization');
+      const auth = authorization.split('Basic ')[1];
```
rcebulko, line 79: Something I thought about after seeing `bearer xxxxx` in the other auth refactor: this should probably be case insensitive for resiliency

```suggestion
      const auth = authorization.split(/Basic /i)[1];
```

estherkim: Nice!

PR #864, opened 2020-05-18
[release-calendar] Cloud build bugs

estherkim:


release-calendar/cloud_build.yaml
```diff
@@ -36,11 +37,12 @@ steps:
...
   # Deploy the app
-- name: gcr.io/cloud-builders/gcloud
-  dir: release-calendar
-  args:
+  - name: gcr.io/cloud-builders/gcloud
+    dir: release-calendar
+    args:
       - app
       - deploy
+    timeout: 1600s
```
rcebulko, line 45: What's this for? I think Cloud Build already times out way before this

estherkim: I need to set a longer one for app deploy because it times out at around 8 min.

rcebulko: Out of curiosity, is that due to build time or initialization/fetching RTV info?

estherkim: Build is done in previous cloud build steps. Seems like it awhile to update the service. Log says `Updating service [default] (this may take several minutes)...`

rcebulko: Ah gotcha, yeah that takes forever on Owners Bot too, not sure why. Thanks for explaining!

PR #880, opened 2020-06-09
[bundle-size] do not add bundle-size reviewers to draft/wip PRs

### LINKED ISSUE #618: [bundle-size] Disable reviewer assignment on draft, WIP, and DNS PRs
Any PR created as a draft, or with "WIP" or "DO NOT SUBMIT" in the title should not have reviewers automatically added.

Mirrors #587

danielrozenberg:
Fixes #618

### DIFF
--- a/bundle-size/test/api.test.js
+++ b/bundle-size/test/api.test.js
@@ -164,6 +164,15 @@ describe('bundle-size api', () => {
           'dist/v0/amp-ad-0.1.js': 4.56,
         },
       };
+
+      this.pullRequestFixture = getFixture('pulls.get.19603');
+      nock('https://api.github.com')
+        .persist()
+        .get('/repos/ampproject/amphtml/pulls/19603')
+        // Make the reply a callback function, because nock stringifies the
+        // body when calling .reply() on the nock scope, and we want tests to be
+        // able to modify this.pullRequestFixture.
+        .reply(200, () => this.pullRequestFixture);
     });
 
     test.each([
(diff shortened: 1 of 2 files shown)


bundle-size/test/api.test.js
```diff
@@ -164,6 +164,15 @@ describe('bundle-size api', () => {
           'dist/v0/amp-ad-0.1.js': 4.56,
         },
       };
+
+      this.pullRequestFixture = getFixture('pulls.get.19603');
+      nock('https://api.github.com')
+        .persist()
+        .get('/repos/ampproject/amphtml/pulls/19603')
+        // Make the reply a callback function, because nock stringifies the
```
rcebulko, line 172: Does it have to be a member property? Could we `let pullRequestFixture` at the top of the describe block, then assign to it in beforeEach and access it as a local variable instead?

bundle-size/test/api.test.js
```diff
@@ -371,6 +380,83 @@ describe('bundle-size api', () => {
...
+        },
+      ],
+      [
+        'title has "WIP" in it',
+        pullRequest => {
+          pullRequest.title = `[WIP] ${pullRequest.title}`;
+        },
+      ],
+    ])(
+      'update a check on bundle-size report with no approvers (report/base = 12.34KB/12.00KB) for PR with %s',
```
rcebulko, line 397: I'm having trouble following what this is testing; it seems like a lot given what's actually changing, and the test description doesn't mention that that this is testing whether or not reviewers are added based on the title/status of the PR. Maybe add a standalone function `shouldAddReviewers` or something like that which can be more easily unit tested?

bundle-size/test/api.test.js
danielrozenberg: It does not have to be a member, you are correct. I still want to maintain the callback trick though because by default almost all the tests will just use the pre-existing fixture. Done.

bundle-size/test/api.test.js
danielrozenberg: These tests honestly need a serious overhaul. I changed the name of the test to be more clear as to what it checks for, but I'll defer the major overhaul to [probably never 🙃]

rcebulko: I didn't notice the %s in the test name, so that makes it a bit clearer. Thanks!

danielrozenberg:
Verified working

PR #900, opened 2020-06-12
Add typescript types for main models

### LINKED ISSUE #882: [test-case-reporting] Define JS models & DB tables
/cc @rcebulko 
Part of #881

Rafer45:
/cc @rcebulko
Together with #896, works toward #882

### DIFF
--- a/test-case-reporting/types/test-case-reporting.d.ts
+++ b/test-case-reporting/types/test-case-reporting.d.ts
@@ -22,9 +22,7 @@ declare module 'test-case-reporting' {
 
   export type TestStatus = 'pass' | 'fail' | 'skip' | 'error';
 
-  /**
-   * A build on Travis.
-   */
+  /** A build on Travis. */
   export interface Build {
     username: string,
     commitHash: string,
@@ -34,13 +32,11 @@ declare module 'test-case-reporting' {
     // The list of jobs we know are contained in the build.
     // When we create a build, we fill this with its jobs, but when
     // we get a build from a server, the jobs list may not be populated
-    // if it is more performant to omit them.
+    // if the jobs are not needed.
     jobs: Array<Job>,
   }
 
-  /**
-   * A job within a Travis build
-   */
+  /** A job within a Travis build. */
   export interface Job {
     jobNumber: string,
     type: JobType,
@@ -51,17 +47,13 @@ declare module 'test-case-reporting' {
     testRuns: Array<TestRun>,
   }
 
-  /**
-   * A single kind of test case, one `it` or `test` block
-   */
+  /** A single kind of test case, one `it` or `test` block. */
   export interface TestCase {
     name: string,
     createdAt: Date,
   }
 
-  /**
-   * An instance of a test being run, with results
-   */
+  /** An instance of a test being run, with results. */
   export interface TestRun {
     testCase: TestCase,
     status: TestStatus,
--- a/test-case-reporting/types/test-case-reporting.d.ts
+++ b/test-case-reporting/types/test-case-reporting.d.ts
@@ -31,10 +31,10 @@ declare module 'test-case-reporting' {
     prNumber: number,
     startedAt: Date,
 
-    // The list of jobs we know are contained in the Build.
-    // When we create a Build, we fill this with its Jobs, but when
-    //   we get a build from a server, the jobs list may not be populated
-    //   if it is more performant to omit them.
+    // The list of jobs we know are contained in the build.
+    // When we create a build, we fill this with its jobs, but when
+    // we get a build from a server, the jobs list may not be populated
+    // if it is more performant to omit them.
     jobs: Array<Job>,
   }
 
--- a/test-case-reporting/types/test-case-reporting.d.ts
+++ b/test-case-reporting/types/test-case-reporting.d.ts
@@ -16,27 +16,14 @@
 
 declare module 'test-case-reporting' {
   /**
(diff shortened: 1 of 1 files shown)


test-case-reporting/types/test-case-reporting.d.ts
```diff
@@ -0,0 +1,76 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+declare module 'test-case-reporting' {
+  /**
+   * A standard logging interface.
+   */
+  export interface Logger {
```
rcebulko, line 21: This was added for Probot (GitHub) apps, because Probot provides app- and request-level loggers with this interface, which matches that of `console.XXX`, so this type definition allowed us to use a single uniform `Logger` interface everywhere. Since our app will just use regular `console.` logging, this isn't necessary here

test-case-reporting/types/test-case-reporting.d.ts
```diff
@@ -0,0 +1,76 @@
...
+    debug(message: string, ...extraInfo: unknown[]): void;
+    warn(message: string, ...extraInfo: unknown[]): void;
+    error(message: string, ...extraInfo: unknown[]): void;
+    info(message: string, ...extraInfo: unknown[]): void;
+  }
+
+  /**
+   * Possible job types.
+   */
+  export type JobType = 'unit' | 'integration';
```
rcebulko, line 31: When possible, try to avoid comments that don't add useful information. Here, something like:
`/** Travis job types for which test results may be reported. */`

The test status type below likewise could be slightly more descriptive, a la
`/** Possible statuses for a given test run. */`

Additional style nit: When the JSDoc is just one line, we can collapse it as shown above

test-case-reporting/types/test-case-reporting.d.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  /**
+   * Possible test statuses
+   */
+  export type TestStatus = 'pass' | 'fail' | 'skip' | 'error';
+
+  /**
+   * One concrete Travis build
```
rcebulko, line 39: I would just say `A build on Travis.` and below `A job within a Travis build.`

On another note, I see the models here are connected the way the relational database connects them. That is, a test run has a job, and a job has a build. When we think about these models, we're more likely to think of a `Build` as containing a list of `Job`s, while a `Job` includes a list of `TestRuns`. We'll chat about this in our next 1:1

test-case-reporting/types/test-case-reporting.d.ts
```diff
@@ -0,0 +1,76 @@
...
+   */
+  export interface Job {
+    build: Build,
+    jobNumber: string,
+    type: JobType,
+    startedAt: Date,
+  }
+
+  /**
+   * One abstract kind of test case
```
rcebulko, line 59: Maybe something more like `A single test case, ie. one `it` or `test` block`

test-case-reporting/types/test-case-reporting.d.ts
```diff
@@ -16,58 +16,53 @@
...
   export type JobType = 'unit' | 'integration';
 
-  /**
-   * Possible test statuses
-   */
   export type TestStatus = 'pass' | 'fail' | 'skip' | 'error';
 
   /**
-   * One concrete Travis build
+   * A build on Travis.
```
rcebulko, line 26: nit: `/** One line doc comment. */`

test-case-reporting/types/test-case-reporting.d.ts
```diff
@@ -16,58 +16,53 @@
...
   export interface Build {
     username: string,
     commitHash: string,
     prNumber: number,
     startedAt: Date,
+
+    // The list of jobs we know are contained in the Build.
+    // When we create a Build, we fill this with its Jobs, but when
+    //   we get a build from a server, the jobs list may not be populated
+    //   if it is more performant to omit them.
```
rcebulko:
```suggestion
    //   if they are not needed.
```

test-case-reporting/types/test-case-reporting.d.ts
rcebulko: (same above)

test-case-reporting/types/test-case-reporting.d.ts
```diff
@@ -16,58 +16,53 @@
...
+   * A single kind of test case, one `it` or `test` block
    */
   export interface TestCase {
     name: string,
     createdAt: Date,
   }
 
   /**
-   * One concrete run, in a job, of some testcase
+   * An instance of a test being run, with results
```
rcebulko, line 63: hyper-nit: period at the end of these :)

test-case-reporting/types/test-case-reporting.d.ts

test-case-reporting/types/test-case-reporting.d.ts

test-case-reporting/types/test-case-reporting.d.ts

test-case-reporting/types/test-case-reporting.d.ts
Rafer45: Done.

test-case-reporting/types/test-case-reporting.d.ts

test-case-reporting/types/test-case-reporting.d.ts

test-case-reporting/types/test-case-reporting.d.ts

Issue #884, opened 2020-06-10
[test-case-reporting] Define API endpoint stub


Rafer45:
/cc @rcebulko
Part of #881

PR #906, opened 2020-06-16
[test-case-reporting] Add types for Travis test reports

Rafer45:
/cc @rcebulko 
Works toward #884.

Sample of the test result report JSON generated when we run `gulp unit`
```json
    "summary": {
        "success": 10879,
        "failed": 21,
        "skipped": 195,
        "error": false,
        "disconnected": false,
        "exitCode": 1
    },
    "browsers": [
        {
            "browser": {
                "id": "52060",
                "fullName": "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/83.0.4103.97 Safari/537.36",
                "name": "Chrome 83.0.4103.97 (Linux x86_64)",
                "state": "DISCONNECTED",
                "lastResult": {
                    "startTime": 1592249357208,
                    "total": 11095,
                    "success": 10879,
                    "failed": 21,
                    "skipped": 195,
                    "totalTime": 813363,
                    "netTime": 275545,
                    "error": false,
                    "disconnected": false
                },
                "disconnectsCount": 0,
                "noActivityTimeout": 120000,
                "disconnectDelay": 120000
            },
            "errors": [],
            "results": [
                {
                    "id": "",
                    "description": "generates extensions metadata",
                    "suite": [],
                    "success": true,
                    "skipped": false,
                    "pending": false,
                    "time": 4,
                    "log": [],
                    "assertionErrors": [],
                    "startTime": 1592249357248,
                    "endTime": 1592249357258
                },
                {
                    "id": "",
                    "description": "generates runtime offsets",
                    "suite": [],
                    "success": true,
                    "skipped": false,
                    "pending": false,
                    "time": 2,
                    "log": [],
                    "assertionErrors": [],
                    "startTime": 1592249357264,
                    "endTime": 1592249357271
                },
                {
                    "id": "",
                    "description": "generates json offsets",
                    "suite": [],
                    "success": true,
                    "skipped": false,
                    "pending": false,
                    "time": 1,
                    "log": [],
                    "assertionErrors": [],
                    "startTime": 1592249357271,
                    "endTime": 1592249357275
                },
                ...
            ]
        }
    ]
}
```

### DIFF
--- a/test-case-reporting/types/test-case-reporting.d.ts
+++ b/test-case-reporting/types/test-case-reporting.d.ts
@@ -76,7 +76,7 @@ declare module 'test-case-reporting' {
     results: Array<TrrTestResult>
   }
 
-  export interface TrrTestResult {
+  export interface TestResult {
     description: string,
     suite: Array<string>,
     success: boolean,
--- a/test-case-reporting/types/test-case-reporting.d.ts
+++ b/test-case-reporting/types/test-case-reporting.d.ts
@@ -61,14 +61,14 @@ declare module 'test-case-reporting' {
     durationMs: number,
   }
 
-  export interface TravisReport {
+  export interface StructuredTestReport {
     successCount: number,
     failedCount: number,
     skippedCount: number,
-    results: Array<Result>,
+    results: Array<TestResultReport>,
   }
 
-  export interface Result {
+  export interface TestResultReport {
     name: string,
     status: TestStatus,
     timeMs: number,
--- a/test-case-reporting/types/test-case-reporting.d.ts
+++ b/test-case-reporting/types/test-case-reporting.d.ts
@@ -61,16 +61,27 @@ declare module 'test-case-reporting' {
     durationMs: number,
   }
 
-  export interface StructuredTestReport {
-    successCount: number,
-    failedCount: number,
-    skippedCount: number,
-    results: Array<TestResultReport>,
+  export interface TestResultReport {
+    summary: TrrSummary,
+    browsers: TrrBrowsers,
   }
 
-  export interface TestResultReport {
-    name: string,
-    status: TestStatus,
-    timeMs: number,
+  export interface TrrSummary {
+    success: number,
+    failed: number,
+    skipped: number,
+  }
+
+  export interface TrrBrowsers {
+    results: Array<TrrTestResult>
+  }
+
+  export interface TrrTestResult {
+    description: string,
+    suite: Array<string>,
+    success: boolean,
+    skipped: boolean,
+    pending: boolean,
+    time: number, // in milliseconds
   }
 }
--- a/test-case-reporting/types/test-case-reporting.d.ts
+++ b/test-case-reporting/types/test-case-reporting.d.ts
@@ -61,27 +61,22 @@ declare module 'test-case-reporting' {
     durationMs: number,
   }
 
-  export interface TestResultReport {
-    summary: TrrSummary,
-    browsers: TrrBrowsers,
-  }
+  namespace KarmaReporter {
+    export interface TestResultReport {
+      browsers: BrowserResultSet,
+    }
 
-  export interface TrrSummary {
-    success: number,
-    failed: number,
-    skipped: number,
-  }
+    export interface BrowserResultSet {
+      results: Array<TestResult>
+    }
 
-  export interface BrowserResultSet {
-    results: Array<TrrTestResult>
-  }
-
-  export interface TestResult {
-    description: string,
-    suite: Array<string>,
-    success: boolean,
-    skipped: boolean,
-    pending: boolean,
-    time: number, // in milliseconds
+    export interface TestResult {
+      description: string,
+      suite: Array<string>,
+      success: boolean,
+      skipped: boolean,
+      pending: boolean,
+      time: number, // in milliseconds
+    }
   }
 }
--- a/test-case-reporting/types/test-case-reporting.d.ts
+++ b/test-case-reporting/types/test-case-reporting.d.ts
@@ -72,7 +72,7 @@ declare module 'test-case-reporting' {
     skipped: number,
   }
 
-  export interface TrrBrowsers {
+  export interface BrowserResultSet {
     results: Array<TrrTestResult>
   }
 
--- a/test-case-reporting/types/test-case-reporting.d.ts
+++ b/test-case-reporting/types/test-case-reporting.d.ts
@@ -60,4 +60,17 @@ declare module 'test-case-reporting' {
     timestamp: Date,
     durationMs: number,
   }
+
+  export interface TravisReport {
+    successCount: number,
+    failedCount: number,
+    skippedCount: number,
+    results: Array<Result>,
+  }
+
+  export interface Result {
+    name: string,
+    status: TestStatus,
+    timeMs: number,
+  }
 }


test-case-reporting/types/test-case-reporting.d.ts
```diff
@@ -60,4 +60,17 @@ declare module 'test-case-reporting' {
...
+
+  export interface TravisReport {
+    successCount: number,
+    failedCount: number,
+    skippedCount: number,
+    results: Array<Result>,
+  }
+
+  export interface Result {
+    name: string,
```
Rafer45, line 72: `name` is not a good label for this, since tests don't have a name, they have a description and a suite. The same is true for the label on line 52.

I'm thinking that, for example, the following block:
```ts
describe('Array', function () {
  describe('#indexOf()', function () {
    it('should return -1 when the value is not present', function () {
      assert.equal([1, 2, 3].indexOf(4), -1);
    });
  });
});
```
should be identified by the following string:
`'Array/#indexOf()/should return -1 when the value is not present'`
or maybe
`'Array/#indexOf():should return -1 when the value is not present'`
We could call this string a description, an identifier, a suite_and_description...?

test-case-reporting/types/test-case-reporting.d.ts
```diff
@@ -60,4 +60,17 @@ declare module 'test-case-reporting' {
     timestamp: Date,
     durationMs: number,
   }
+
+  export interface TravisReport {
```
Rafer45, line 64: This name is brittle -- it'll become inaccurate if we stop using Travis. Any suggestions? Maybe `RemoteTestReport` or `TestReport`?

rcebulko: `TestResultReport`? `¯\_(ツ)_/¯`

test-case-reporting/types/test-case-reporting.d.ts
rcebulko: I would be in favor of either `' / '` or `' | '` as the delimiter (with spaces on either side). I'd lean toward naming the interface `TestResult`. Internally, mocha contexts offer a `name` field and `fullName()` method, so I think this nomenclature is fine. Besides, with regards to these interfaces, the goal here is to define what the JSON reporter actually uses, which isn't up to us. We'll take that data and massage it into our preferred format in the report handler.

Rafer45: `> Internally, mocha contexts offer a name field and fullName() method, so I think this nomenclature is fine.`
Understood.
`> the goal here is to define what the JSON reporter actually uses, which isn't up to us.`
I've added a sample of the JSON reporter output to the PR description. Are the types good, or should they stick more closely to the JSON structure? (e.g. by have a string array for the suites).

test-case-reporting/types/test-case-reporting.d.ts
Rafer45: SGTM, I'll change it to that.

Rafer45: Oh, one more thought -- the name `Result` isn't very descriptive, but all the alternatives off the top of my head are clunky: `ResultReport`, `SingleResultReport`, `ReportResult`...

Any ideas?

rcebulko: See above suggestion

test-case-reporting/types/test-case-reporting.d.ts
rcebulko: The point here is that these interfaces will describe the JSON reported by Travis, so that you have a type definition for the data incoming to the API endpoint. So yes, any fields defined in the interface must match the JSON format exactly. It looks like a lot of these fields need to change to match the data here

Rafer45: Okay, I'll get to work on that.

test-case-reporting/types/test-case-reporting.d.ts
Rafer45: By `above suggestion` do you mean `TestResultReport shrug` or `The point here is that these interfaces will describe the JSON...`?

If you mean the former, I thought that was a suggestion for the type on line 64, I'm asking about the type on line 71 (my bad for putting that comment here).

rcebulko: Here, I mean the former. In the thread on line 72, I suggested `TestResult`

Rafer45: Understood.

test-case-reporting/types/test-case-reporting.d.ts

test-case-reporting/types/test-case-reporting.d.ts
Rafer45: Done, kind of (I called it `TrrTestResult`, let me know if you disagree).

test-case-reporting/types/test-case-reporting.d.ts
```diff
@@ -61,16 +61,27 @@ declare module 'test-case-reporting' {
     durationMs: number,
   }
 
-  export interface TravisReport {
-    successCount: number,
-    failedCount: number,
-    skippedCount: number,
-    results: Array<Result>,
+  export interface TestResultReport {
```
rcebulko, line 64: Lets wrap all of these interfaces in `namespace Travis`

test-case-reporting/types/test-case-reporting.d.ts
```diff
@@ -61,16 +61,27 @@ declare module 'test-case-reporting' {
...
+    success: number,
+    failed: number,
+    skipped: number,
+  }
+
+  export interface TrrBrowsers {
+    results: Array<TrrTestResult>
+  }
+
+  export interface TrrTestResult {
```
rcebulko, line 79:
```suggestion
  export interface TestResult {
```

test-case-reporting/types/test-case-reporting.d.ts
```diff
@@ -61,16 +61,27 @@ declare module 'test-case-reporting' {
...
+  export interface TestResultReport {
+    summary: TrrSummary,
+    browsers: TrrBrowsers,
   }
 
-  export interface Result {
-    name: string,
-    status: TestStatus,
-    timeMs: number,
+  export interface TrrSummary {
```
rcebulko, line 69: How will this data be used? Does it match up with any of our DB columns?

test-case-reporting/types/test-case-reporting.d.ts
```diff
@@ -61,16 +61,27 @@ declare module 'test-case-reporting' {
...
+    success: number,
+    failed: number,
+    skipped: number,
+  }
+
+  export interface TrrBrowsers {
```
rcebulko, line 75:
```suggestion
  export interface BrowserResultSet {
```

test-case-reporting/types/test-case-reporting.d.ts
Rafer45: A namespace is a great idea.
Nit: Do we want to use the name `Travis`, or something that is build-tool-independent (`RemoteReports` or something).

rcebulko: You're right, `KarmaReporter` makes more sense here for the namespace

test-case-reporting/types/test-case-reporting.d.ts
Rafer45: Oh, you're right -- I added that thinking it might be useful for the web view, but the web view will be fetching from the database, so this not used.

I'll remove it and let you know when done.

test-case-reporting/types/test-case-reporting.d.ts

test-case-reporting/types/test-case-reporting.d.ts
Rafer45: Fix applied.

test-case-reporting/types/test-case-reporting.d.ts
Rafer45: Fix applied.

test-case-reporting/types/test-case-reporting.d.ts

Rafer45:
@rcebulko, I think googlebot is complaining about our co-authored commits. What should I do?

PR #907, opened 2020-06-16
[test-case-reporting] Create types for interfacing with database

### LINKED ISSUE #882: [test-case-reporting] Define JS models & DB tables

Rafer45:
/cc @rcebulko 
Closes #882.

### DIFF
--- a/test-case-reporting/types/test-case-reporting.d.ts
+++ b/test-case-reporting/types/test-case-reporting.d.ts
@@ -18,7 +18,7 @@ declare module 'test-case-reporting' {
   /**
    * Travis job types for which test results may be reported.
    */
-  export type JobType = 'unit' | 'integration';
+  export type TestSuiteType = 'unit' | 'integration';
 
   export type TestStatus = 'PASS' | 'FAIL' | 'SKIP' | 'ERROR';
 
@@ -39,7 +39,7 @@ declare module 'test-case-reporting' {
   /** A job within a Travis build. */
   export interface Job {
     jobNumber: string,
-    type: JobType,
+    testSuiteType: TestSuiteType, // e.g. unit, integration, etc.
     startedAt: Date,
 
     // This list is treated similarly to the `jobs` array of the `Build` type.
@@ -77,7 +77,7 @@ declare module 'test-case-reporting' {
       id?: number,
       buld_id: number,
       job_number: string,
-      test_suite_type: string,
+      test_suite_type: string, // e.g. unit, integration, etc.
       started_at: number,
     }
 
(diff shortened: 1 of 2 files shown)


test-case-reporting/types/test-case-reporting.d.ts
```diff
@@ -60,4 +60,42 @@ declare module 'test-case-reporting' {
...
+      pull_request_number: number,
+      started_at: number,
+    }
+
+    export interface Job {
+      // See comment under `DB.Build.id`
+      id?: number,
+      buld_id: number,
+      job_number: string,
+      test_suite_type: string,
```
rcebulko, line 80: comment something like `// ex.  unit, integration, etc.`

test-case-reporting/types/test-case-reporting.d.ts
```diff
@@ -39,7 +39,7 @@ declare module 'test-case-reporting' {
   /** A job within a Travis build. */
   export interface Job {
     jobNumber: string,
-    type: JobType,
+    testSuiteType: TestSuiteType, // e.g. unit, integration, etc.
```
rcebulko, line 42: Here you're referencing the type, so it's obvious. Below it's just a string, which is why I suggested the comment

Rafer45: Removed.

test-case-reporting/types/test-case-reporting.d.ts

test-case-reporting/types/test-case-reporting.d.ts
```diff
@@ -18,65 +18,106 @@ declare module 'test-case-reporting' {
...
+      pull_request_number: number;
+      started_at: number;
+    }
+
+    export interface Job {
+      // See comment under `DB.Build.id`
+      id?: number;
+      buld_id: number;
+      job_number: string;
+      test_suite_type: string; // e.g. unit, integration, etc.
```
estherkim, line 83: Should this be type `TestSuiteType`?

Rafer45: Oh, good point.

Instead of turning this into `TestSuiteType`, I think I'll turn the other field into a string and remove the enum to follow  https://github.com/ampproject/amp-github-apps/pull/896#discussion_r441030655. Not sure, though -- what do you think?

rcebulko: The comment was about removing the _database_ enum type. In typescript, however, we still want the type-checking/validation. I agree with making this `TestSuiteType`

PR #704, opened 2020-02-12
[invite] Add (currently failing) end-to-end tests for main user stories

rcebulko:
```
 FAIL  test/e2e.test.ts
  end-to-end
    when a comment includes "/invite @someone"
      when @someone is a member of the org
        ✕ comments, doesn't record (5ms)
      when @someone is not a member of the org
        ✕ invites, records, comments (3ms)
        once the invite is accepted
          ✕ comments, archives (2ms)
    when a comment includes "/tryassign @someone"
      when @someone is a member of the org
        ✕ assigns, comments, doesn't record (2ms)
      when @someone is not a member of the org
        ✕ invites, records, comments (2ms)
        once the invite is accepted
          ✕ assigns, comments, archives (1ms)
    when a comment includes no macros
      ✕ ignores it (1ms)
    when someone joins without a recorded invitation
      ✕ ignores it (2ms)
```

PR #908, opened 2020-06-16
[test-case-reporting] Add API endpoint stub

### LINKED ISSUE #884: [test-case-reporting] Define API endpoint stub

Rafer45:
/cc @estherkim @rcebulko 
Closes #884.


test-case-reporting/index.ts
```diff
@@ -34,6 +34,10 @@ app.get('/test-results/history/:testCaseId', (req, res) => {
   res.send(`Test history for test with name/ID ${testCaseId}`);
 });
 
+app.post('/report', function (req, res) {
+  res.send(`Travis report! The request body is below.\n\n${req.body}`);
```
rcebulko, line 38: Something like `const report: TravisReport = request.body.report` would set this up for later progress using your type defs

Rafer45: Do I rebase this onto the tip of #906 so I have the types?

Rafer45: Nevermind, #906 just got merged; I'll pull master then rebase.

Rafer45: Done.

Rafer45:
@googlebot is rejecting the PR because of a co-authored commit ([`f890df5`](f890df5)), so I need someone to override it (like in #906). If someone has a cleaner solution, let me know -- I considered rebasing to remove that commit from the branch (since it's redundant with the one on main), but I think I would then have to force push, and I think that would erase the comment history (don't know, though).

test-case-reporting/types/test-case-reporting.d.ts
```diff
@@ -24,40 +24,59 @@ declare module 'test-case-reporting' {
...
+    export interface BrowserResultSet {
+      results: Array<TestResult>;
+    }
+
+    export interface TestResult {
+      description: string;
+      suite: Array<string>;
+      success: boolean;
+      skipped: boolean;
+      pending: boolean;
```
rcebulko, line 78: How do you plan to use `pending`?

Rafer45: From what I saw, every test with `skipped: true` has `pending: true`, so I don't know exactly what `pending` means here, but it's not always false for tests that have run.

I'll remove it since we don't currently have a use for it.

Rafer45: Also, I'll fix this in the db-types pull request (#907)

rcebulko: I suspect `pending` is set to `false` after a test case completes, and would mainly be used to check something during the tests running. Since we are only determining this after the whole test suite is completed, it shouldn't matter to us.

rcebulko:
See https://github.com/ampproject/amp-github-apps/pull/704/ for an example of some e2e tests written before the underlying implementation, which also use test data and a test DB

Rafer45:
> Given how small this PR is, it seems like it may be a good place to start introducing e2e tests. The tests will be failing for now, but having an example report payload and asserting that a POST to that endpoint followed by a DB query produces the desired results should be a reasonably straightforward test to write and would set us up for success moving forward

I would like to do that in a new PR so that we stop getting blocked by the CLA and because this PR already looks a bit funky with [`c523155`](https://github.com/ampproject/amp-github-apps/pull/908/commits/c52315582c3c1df6c941942d2d43b3062f2b5d9e) fixing the same typo as [`e12294f`](https://github.com/ampproject/amp-github-apps/pull/907/commits/e12294f86d5b51f91ae7a600d6c18201ffb55b46).

If we do that I would change this PR's description from "Closes #884" to "Works toward #884" so that we can close the issue with the PR that adds the tests.

Would that be okay, or would you prefer I do it on this PR?

rcebulko:
@Rafer45 If the commit is duplicated, then you probably should be rebasing this. I'm not sure why the commit from #906 is a part of this branch, but you should be able to rebase the changes from this branch on top of `master:HEAD` and any CLA issues will vanish.

Rafer45:
> @Rafer45 If the commit is duplicated, then you probably should be rebasing this. I'm not sure why the commit from #906 is a part of this branch, but you should be able to rebase the changes from this branch on top of `master:HEAD` and any CLA issues will vanish.

If I do that, I'll have to force push, right? Would that be okay?

test-case-reporting/test/e2e.test.ts
```diff
@@ -0,0 +1,90 @@
...
+describe('end-to-end', () => {
+  let db: Database;
+
+  beforeAll(async () => {
+    nock.disableNetConnect();
+    process.env = {
+      DB_UNIX_SOCKET: 'amp-test-cases/socket',
+      DB_USER: 'test_user',
+      DB_PASSWORD: 'test_password',
+      DB_NAME: 'test_results',
```
rcebulko, line 44: Are any of these ENV vars necessary, given that you're mocking out the `db` module?

test-case-reporting/test/e2e.test.ts
```diff
@@ -0,0 +1,90 @@
...
+  });
+
+  afterAll(() => {
+    nock.enableNetConnect();
+    db.destroy();
+  });
+
+  afterEach(() => {
+    jest.restoreAllMocks();
+    db('builds').truncate();
```
rcebulko, line 59: Something to be aware of: unless you have `ON CASCADE DELETE` set for your foreign keys (you don't), trying to delete a row that is referenced by a foreign key on another table will fail. For instance, if you have a row in `builds` with `id=6`, and a row in `jobs` with `build_id=6`, trying to delete the build will fail. You'll have to delete entries in reverse-dependency order, so in this case, `jobs` have to be deleted before the `builds` they reference. You should be able to extrapolate for the remaining tables.

SQL has lots of little edge cases :P

test-case-reporting/test/e2e.test.ts
```diff
@@ -0,0 +1,90 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS-IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Knex from 'knex';
+import nock from 'nock';
```
rcebulko, line 18: We don't actually mock any network requests, so you can remove this import along with all `nock` message calls, and probably the `package.json` dependency too.

`nock` would be used if our app needed to, for example, make a POST request to the Travis API. We would "mock" that network request, and could expect certain properties about outgoing requests _from_ our app.

test-case-reporting/types/test-case-reporting.d.ts
```diff
@@ -103,6 +103,7 @@ declare module 'test-case-reporting' {
   }
 
   namespace KarmaReporter {
+    // TODO(rafer45): Add job & build info to the karma report
```
rcebulko, line 106: Since the job and build are part of Travis, not Karma, they'd be added as a separate interface outside of this Karma namespace.

test-case-reporting/test/fixtures/sample-karma-report.json
```diff
@@ -0,0 +1,77 @@
...
+                    "time": 413,
+                    "log": [],
+                    "assertionErrors": [],
+                    "startTime": 1610,
+                    "endTime": 2050
+                }
+            ]
+        }
+    ]
+}
```
rcebulko, line 77: Nit: EOF newline

test-case-reporting/test/e2e.test.ts
```diff
@@ -0,0 +1,90 @@
...
+      nock.cleanAll();
+      throw new Error('Not all nock interceptors were used!');
+    }
+  });
+
+  describe('when one post request is received', async () => {
+    it('updates the database if the post request is good', async () => {
+      await request(app)
+        .post('/report')
+        .send(getFixtureAsString('sample-karma-report.json'));
```
rcebulko, line 75: `.send()` accepts JSON, and you'll need it to in order to add job/build info here. This might end up looking something like:
```javascript
.send({
  job: {...},
  build: {...},
  results: getFixture('sample-karma-report')
})
```

Note that the `getFixture` methods you copied over add the `.json` suffix for you

test-case-reporting/test/e2e.test.ts
```diff
@@ -0,0 +1,90 @@
...
+
+      // TODO(rafer45): Replace `db('table_name').select()` calls with more readable
+      // functions for getting builds, jobs, and invites.
+      // See https://github.com/ampproject/amp-github-apps/blob/master/invite/src/invitation_record.ts
+      // for an example of what I'm thinking of.
```
rcebulko, line 80: Looks good; when that happens, we should check the values of the results ie validate that the testruns being returned match the ones in our test data. Fine to leave this as it is for now.

test-case-reporting/test/fixtures/index.ts
```diff
@@ -0,0 +1,30 @@
...
+import path from 'path';
+
+export function getFixtureAsString(name: string): string {
+  return fs.readFileSync(path.join(__dirname, `${name}.json`)).toString('utf8');
+}
+
+/**
+ * Get a JSON test fixture object.
+ */
+export function getFixture(name: string): KarmaReporter.TestResultReport {
```
rcebulko, line 28:
```suggestion
export function getFixture(name: string): Record<string, unknown> {
```

Let's make this `getFixture` helper as generic as possible, since it could conceivably return any JSON. We can always cast it to be more specific at the call site.

rcebulko:
This is a great start! Lots of little comments, but all are minor tweaks. High-level this PR is on point

test-case-reporting/test/e2e.test.ts
rcebulko: Nice explanation of `ON DELETE` actions, explained through a metaphor of businesses and the bourgeoisie:
https://dba.stackexchange.com/a/213239

test-case-reporting/test/e2e.test.ts
Rafer45: I don't think so; I'll remove them.

test-case-reporting/test/e2e.test.ts
Rafer45: Done.

test-case-reporting/test/fixtures/sample-karma-report.json
Rafer45: Done.

test-case-reporting/types/test-case-reporting.d.ts

test-case-reporting/test/e2e.test.ts
Rafer45: Done.

test-case-reporting/test/e2e.test.ts
Rafer45: Done. I'll check out the link!

test-case-reporting/test/e2e.test.ts
Rafer45: Done.

test-case-reporting/test/fixtures/index.ts
Rafer45: Done.

test-case-reporting/test/e2e.test.ts
Rafer45: Understood.

test-case-reporting/tsconfig.json
```diff
@@ -21,6 +21,7 @@
     "./*",
     "src/**/*",
     "test/*",
-    "types/*"
+    "types/*",
+    "test/fixtures/*"
```
estherkim, line 25: You can include all files under `test/` with `"test/**/*"`

test-case-reporting/test/e2e.test.ts
```diff
@@ -0,0 +1,83 @@
...
+  });
+
+  describe('when one post request is received', async () => {
+    it('updates the database if the post request is good', async () => {
+      await request(app)
+        .post('/report')
+        .send({
+          build: {
+            buildNumber: '413413',
+            commitHash: 'abcdefg123gomugomu',
```
estherkim, line 61: nit: GitHub calls this a commit sha

test-case-reporting/test/e2e.test.ts
```diff
@@ -0,0 +1,83 @@
...
+
+import Knex from 'knex';
+import request from 'supertest';
+
+import {Database, dbConnect} from '../src/db';
+import {app} from '../index';
+import {getFixture} from './fixtures';
+import {setupDb} from '../src/setup_db';
+
+jest.mock('../src/db', () => ({
```
estherkim, line 25: nit: I like to name test files after the src files, i.e. `db.test.ts`, because you might want to split your e2e tests across multiple files.

test-case-reporting/test/e2e.test.ts
```diff
@@ -0,0 +1,83 @@
...
+      // functions for getting builds, jobs, and invites.
+      // See https://github.com/ampproject/amp-github-apps/blob/master/invite/src/invitation_record.ts
+      // for an example of what I'm thinking of.
+      const buildsLength = (await db('builds').select()).length;
+      const jobsLength = (await db('jobs').select()).length;
+      const testRunsLength = (await db('testRunsLength').select()).length;
+
+      expect(buildsLength).toEqual(1);
+      expect(jobsLength).toEqual(1);
+      expect(testRunsLength).toEqual(3);
```
estherkim, line 80: jest supports `.toHaveLength(number)` :) https://jestjs.io/docs/en/expect#tohavelengthnumber

test-case-reporting/test/e2e.test.ts
Rafer45: What would that look like? Would it be like
```
import {Database, dbConnect} from '../src/db.test';

jest.mock('../src/db.test', () => ({
  dbConnect: () =>
    Knex({
      client: 'sqlite3',
      connection: ':memory:',
      useNullAsDefault: true,
    }),
}));
```

test-case-reporting/tsconfig.json
Rafer45: Done.

test-case-reporting/test/e2e.test.ts
Rafer45: Done.

test-case-reporting/test/e2e.test.ts
Rafer45: Done.

test-case-reporting/test/e2e.test.ts
Rafer45: Oh, do you mean renaming the file from `e2e.test` to `db.test`?

rcebulko: @estherkim the plan is to have unit tests mirror the module names, but many of the apps in this repo use `e2e.test.ts` as the end-to-end test file.

test-case-reporting/test/e2e.test.ts
Rafer45: Also adding a fix in https://github.com/ampproject/amp-github-apps/pull/913

test-case-reporting/test/e2e.test.ts
estherkim: Yeah I meant renaming the test file itself. Ok, sg.

Rafer45: Understood, resolving conversation.

test-case-reporting/test/e2e.test.ts
```diff
@@ -0,0 +1,83 @@
...
+            commitSha: 'abcdefg123gomugomu',
+          },
+          job: {
+            jobNumber: '413413.612',
+            testSuiteType: 'unit',
+          },
+          results: getFixture('sample-karma-report'),
+        });
+
+      // TODO(rafer45): Replace `db('table_name').select()` calls with more readable
```
Rafer45, line 70: Changing this TODO so it points to https://github.com/ampproject/amp-github-apps/issues/914...

Rafer45: Done.

PR #911, opened 2020-06-23
[owners] Validate keys in owner/rule definitions

rcebulko:
Prevents situations like https://github.com/ampproject/amphtml/pull/29001

### DIFF
--- a/owners/src/ownership/parser.js
+++ b/owners/src/ownership/parser.js
@@ -30,6 +30,7 @@ const {
 } = require('./owner');
 
 const GLOB_PATTERN = '**/';
+const RULE_DEF_KEYS = new Set(['owners', 'pattern']);
 const OWNER_DEF_KEYS = new Set([
   'name',
   'notify',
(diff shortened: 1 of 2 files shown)


rcebulko:
> I would encourage you to not reinvent the wheel, and use a [JSON schema](https://json-schema.org/learn/miscellaneous-examples.html) with a well-developed validator library (e.g., [ajv](https://www.npmjs.com/package/ajv))

That looks like a good option to cut down on a lot of the existing parsing validator logic. How would you feel about me filing a separate tracking issue to migrate to a proper validation library, submit this for now to prevent such mistakes in the interim, and mark the migration as a P2 to prioritize within the next month or so?

danielrozenberg:
> That looks like a good option to cut down on a lot of the existing parsing validator logic. How would you feel about me filing a separate tracking issue to migrate to a proper validation library, submit this for now to prevent such mistakes in the interim, and mark the migration as a P2 to prioritize within the next month or so?

Sounds good :)

PR #913, opened 2020-06-24
[test-case-reporting] Type & database cleanup

Rafer45:
/cc @rcebulko @estherkim 
Adds types for the Travis POST requests, replaces pull request number in build rows with build number (because pull request number is redundant with commit hash).


test-case-reporting/types/test-case-reporting.d.ts
```diff
@@ -119,4 +119,20 @@ declare module 'test-case-reporting' {
...
+      job: Travis.Job;
+      build: Travis.Build;
+      result: KarmaReporter.TestResultReport;
+    }
+    export interface Build {
+      buildNumber: string;
+      commitSha: string;
+    }
+    export interface Job {
+      jobNumber: string;
```
estherkim, line 134: Should this be type number?

test-case-reporting/types/test-case-reporting.d.ts
```diff
@@ -119,4 +119,20 @@ declare module 'test-case-reporting' {
...
   }
+
+  namespace Travis {
+    export interface Report {
+      job: Travis.Job;
+      build: Travis.Build;
+      result: KarmaReporter.TestResultReport;
+    }
+    export interface Build {
+      buildNumber: string;
```
estherkim, line 130: Should this be type number?

test-case-reporting/types/test-case-reporting.d.ts
```diff
@@ -70,15 +69,15 @@ declare module 'test-case-reporting' {
...
+      build_number: number;
       started_at: number;
     }
 
     export interface Job {
       // See comment under `DB.Build.id`
       id?: number;
-      buld_id: number;
+      build_id: number;
       job_number: string;
```
estherkim, line 81: Should this be type number?

test-case-reporting/types/test-case-reporting.d.ts
Rafer45: I made it a string because that's what Travis returns -- https://github.com/ampproject/amphtml/blob/master/build-system/common/travis.js#L52

I see the inconsistency, though -- would you recommend making the buildNumber a string across the board, a number across the board, something else?

test-case-reporting/types/test-case-reporting.d.ts
Rafer45: I think so; it's a foreign key which is a number on the database. It doesn't correspond to the build number.

test-case-reporting/types/test-case-reporting.d.ts
Rafer45: (I should definitely add a comment to the Travis namespace explaining it's supposed to match the Travis POST requests)

test-case-reporting/types/test-case-reporting.d.ts
Rafer45: Job numbers look like `1234.567`, so it should be a string.

test-case-reporting/types/test-case-reporting.d.ts
estherkim: Ok, let's change it.

test-case-reporting/types/test-case-reporting.d.ts
estherkim: SG

estherkim: Although that should be fine with number type too...

test-case-reporting/types/test-case-reporting.d.ts
estherkim: You could cast it from string to number. It's just that it's misleading that `*Number` is a string. Up to you.

test-case-reporting/types/test-case-reporting.d.ts
Rafer45: This one's already type number

test-case-reporting/types/test-case-reporting.d.ts
Rafer45: I'll leave it as a string and add a clarifying comment

test-case-reporting/types/test-case-reporting.d.ts
Rafer45: We might get a floating point error in that case, I believe.

test-case-reporting/types/test-case-reporting.d.ts
estherkim: Sorry, I was commenting on `job_number`.

Rafer45: Oh, sorry, I thought you meant `build_id`. `job_number` is represented as a string in the database because of the `123.456` structure

test-case-reporting/types/test-case-reporting.d.ts
rcebulko: I'd stick with string here since JS floating points can drift, and since it's not actually representative of a number

PR #915, opened 2020-06-26
[invite] Exclude expired invites when checking if a user has an invite pending

### LINKED ISSUE #876: Invite bot should handle expired/no-longer existing invitations
User was invited once here, ~1.5 months ago:

https://github.com/ampproject/amphtml/issues/27762

User did not accept invite, so I tried to invite again. Bot won't create a second invite:

https://github.com/ampproject/amphtml/issues/28708

When going to `https://github.com/orgs/ampproject/invitation`, invitation is no longer there:

image

rcebulko:
Closes #876 by excluding invites older than a week when checking if a user has any pending invitations. Leaves the invitations around so that the expected actions can still take place on the threads that created the now-expired invites, such as update comments and assigning issues.

/cc @alanorozco


rcebulko:
@danielrozenberg friendly ping :upside_down_face:

invite/src/invite_bot.ts
```diff
@@ -34,6 +34,12 @@ const FULL_MACRO_REGEX = new RegExp(
   `(?<!\\S)${MACRO_PATTERN} @${USERNAME_PATTERN}`,
   'ig'
 );
+// GitHub expires invitations after one week.
+const EXPIRATION_INTERVAL_MS = 7 * 24 * 60 * 60 * 1000;
+
+function expirationDate(): Date {
+  return new Date(Number(new Date()) - EXPIRATION_INTERVAL_MS - 60 * 60 * 1000);
```
danielrozenberg, line 41: Not a fan of the back-and-forth type conversions... how about:

```typescript
const date = new Date();
date.setSeconds(date.getSeconds() - EXPIRATION_INTERVAL_SEC - 60 * 60);
```

(you'll have to change the expiration const to seconds, or use date.setMilliseconds but I don't see why that would be beneficial here)

Additional questions:
* why are you doing `- 60 * 60 * 1000` after EXPIRATION_INTERVAL?
* if it isn't _that_ important to be super-accurate, how about just using `date.setDate(date.getDate() - 7)`
* umm... shouldn't this be `+ EXPIRATION_INTERVAL` instead of minus?

invite/src/invite_bot.ts
```diff
@@ -154,9 +160,12 @@ export class InviteBot {
...
     this.logger.debug(`tryInvite: Trying to invite`, invite);
     const existingInvites = await this.record.getInvites(invite.username);
-    const pendingInvite = !!existingInvites.length;
+    const expiration = expirationDate();
+    const pendingInvites = existingInvites.filter(
+      ({created_at}) => created_at > expiration
+    );
 
-    if (pendingInvite) {
+    if (pendingInvites.length) {
```
danielrozenberg, line 168: Oh, JavaScript...

invite/test/invite_bot.test.ts
```diff
@@ -37,6 +37,10 @@ jest.mock('../src/db', () => {
   };
 });
 
+function _daysAgo(days: number): Date {
+  return new Date(Number(new Date()) - days * 24 * 60 * 60 * 1000);
```
danielrozenberg, line 41: (also here)

invite/src/invite_bot.ts
```diff
@@ -34,6 +34,14 @@ const FULL_MACRO_REGEX = new RegExp(
   `(?<!\\S)${MACRO_PATTERN} @${USERNAME_PATTERN}`,
   'ig'
 );
+// GitHub expires invitations after one week.
+const EXPIRATION_INTERVAL_MS = 7 * 24 * 60 * 60;
```
danielrozenberg, line 38: `_SEC` ;)

invite/src/invite_bot.ts
rcebulko: Made the changes you suggested. It's subtracting because anything older than 7 days before right now (hence the minus) is expired.

invite/src/invite_bot.ts
rcebulko: Fixed!

rcebulko:
Travis passed but never reported; merging

Issue #887, opened 2020-06-10
[test-case-reporting] Define standard queries


Rafer45:
/cc @rcebulko
Part of #881
Depends on #882 

- [ ] All tests for a PR
- [ ] History of runs for one test

PR #920, opened 2020-06-29
[test-case-reporting] Implement query for all test runs in a build

Rafer45:
/cc @rcebulko 
Closes part of #887

Still in progress, I need to sanity check this


test-case-reporting/src/db_interfacing.ts
```diff
@@ -0,0 +1,116 @@
...
+ * limitations under the License.
+ */
+
+import {Build, DB, Job, TestCase, TestRun} from 'test-case-reporting';
+import {Database, TIMESTAMP_PRECISION} from './db';
+
+// Types in the DB namespace interface use snake_case instead of camelCase.
+/* eslint @typescript-eslint/camelcase: "off" */
+function getDateFromTimestamp(timestamp: number): Date {
+  const msConversionConstant: number = Math.pow(10, 3 - TIMESTAMP_PRECISION);
```
rcebulko, line 23: Lets define this once up top

test-case-reporting/src/db_interfacing.ts
```diff
@@ -0,0 +1,116 @@
...
+  return new Date(timestamp * msConversionConstant);
+}
+
+export class TestResultsRecords {
+  constructor(private db: Database) {}
+
+  private async getDbBuild(buildNumber: string): Promise<DB.Build> {
```
rcebulko, line 30: Please add some JSDoc function comments for this file, and mention what this will return if the build number isnt't found in the DB

test-case-reporting/src/db_interfacing.ts
```diff
@@ -0,0 +1,116 @@
...
+  private makeTestRunFromRow(row): TestRun {
+    return {
+      testCase: this.makeTestCaseFromRow(row),
+      status: row.test_run_status,
+      timestamp: getDateFromTimestamp(row.test_run_timestamp),
+      durationMs: row.duration_ms,
+    };
+  }
+
+  // I need help typechecking this
```
rcebulko, line 62: Add comments to the PR, rather than in code where it's easy to forget to remove them 🙃

test-case-reporting/src/db_interfacing.ts
```diff
@@ -0,0 +1,116 @@
...
+        'test_runs.status as test_run_status',
+        'test_runs.timestamp as test_run_timestamp',
+        'test_runs.duration_ms as test_run_duration_ms',
+      ])
+      .where('build_number', buildNumber)
+      .join('jobs', 'jobs.build_id', 'builds.id')
+      .join('test_runs', 'test_runs.job_id', 'jobs.id')
+      .join('test_cases', 'test_cases.id', 'test_runs.test_case_id');
+
+    const jobs = joinOfEveryTable
```
rcebulko, line 101: While I tend to like functional solutions, I worry this is introducing a lot of complexity to shoehorn the DB row into the ideal type we wanted. The logic could get more complex for the history of a single test, which will also mean grouping by build in addition to job. What do you think about removing the ''jobs'' and ''testRuns'' properties entirely from the Build and Job types, keeping them very flat, and maybe letting one object whole info for the build, job, and test?

Rafer45: I like that idea; it'll simplify things.

test-case-reporting/src/db_interfacing.ts
Rafer45: Whoops -- I was kind of thinking-out-loud on comments in the code and I forgot to remove that one (proving your point).

test-case-reporting/src/db_interfacing.ts
Rafer45: What does "up top" mean? In the top of the file, or?

Rafer45: What do you mean by "up top"? In the top of the file, or...?

rcebulko: yep, let's make it a file level constant

Rafer45: I believe it is already in the top of the file; it's before the class declaration.

test-case-reporting/src/db_interfacing.ts
Rafer45: I'll do it and get back to you when done.

test-case-reporting/src/db_interfacing.ts
Rafer45: Done.

test-case-reporting/src/db_interfacing.ts
Rafer45: Done adding JSDoc comment to `getDbBuild`, will add more once I've replaced some of the overly-complex functions.

test-case-reporting/src/db_interfacing.ts
Rafer45: Oh, gotcha.
Done.

test-case-reporting/src/db_interfacing.ts
```diff
@@ -0,0 +1,121 @@
...
+function getDateFromTimestamp(timestamp: number): Date {
+  return new Date(timestamp * msConversionConstant);
+}
+
+export class TestResultsRecords {
+  constructor(private db: Database) {}
+
+  /**
+   * Gets a DB.Build from the database by its name.
+   * If there is no build, it returns undefined.
```
Rafer45, line 33: It may be a good idea to throw an error instead of returning undefined -- what do you think? @rcebulko

rcebulko: I'm inclined to use undefined since it's more like `myDict['missing key']` and simpler to handle than errors, but either is valid

test-case-reporting/src/db_interfacing.ts
```diff
@@ -0,0 +1,121 @@
+/**
```
rcebulko, line 1: Let's name this file to reflect the class it defines, ie. `test_results_records.ts`

test-case-reporting/src/db_interfacing.ts
Rafer45: Done.

test-case-reporting/src/db_interfacing.ts
Rafer45: Done.

test-case-reporting/src/db_interfacing.ts
Rafer45: Re: (https://github.com/ampproject/amp-github-apps/pull/920#discussion_r450354029)
Okay, I'll leave it as it is.

test-case-reporting/src/db_interfacing.ts
Rafer45: Done.

test-case-reporting/src/test_result_record.ts
```diff
@@ -0,0 +1,128 @@
...
+  private async getBuildFromBuildNumber(buildNumber: string): Promise<Build> {
+    const dbBuild = await this.getDbBuildFromBuildNumber(buildNumber);
+    return this.getBuildFromDbBuild(dbBuild);
+  }
+
+  private async getTestRunsOfBuild(
+    buildNumber: string
+  ): Promise<Array<TestRun>> {
+    const build: Build = await this.getBuildFromBuildNumber(buildNumber);
+    const dbJobs: Array<DB.Job> = await this.db<DB.Job>('builds')
```
rcebulko, line 71: I suspect this query can also JOIN against the other tables and include the job, test run, and test case info all in one query.

Rafer45: A benefit of the current approach is that it creates only one `Job` per job -- i.e. multiple `TestRun`s that belong to the same `Job` actually reference the same object. If we did a larger join that included job, test case, and rest run info, I don't think I could do the same thing. In your experience, is this kind of benefit worth it, or not really?

rcebulko: What benefits are gained by sharing a single Job object? How does it impact the way the results will be used?

Rafer45: It doesn't impact the way the results will be used, but it avoids duplication of `Job` objects and makes it _slightly_ simpler to check whether two test runs belong to the same job. It also feels structurally cleaner; though I'll acknowledge that's a subjective aesthetic preference and not super practical.

I'm guessing from the conversation that the benefits aren't worth the added database queries -- let me know if my guess is correct. I'll be working on a single-join approach in the meantime.

Rafer45: Done implementing single-query approach.

Rafer45:
> What is `travis_api_utils.ts`?

I had some issues with an interactive rebase; I'll remove the file.

PR #923, opened 2020-07-07
[error-monitoring] Allow for reporting errors to a separate repo + cron endpoint to do so

rcebulko:
This PR:
- allows us to use a dedicated "error-reports" repository to track issues separately from user-created issues, per request of @kristoferbaxter to reduce issue noise for the UI WG
- defines cron endpoints to automatically file issues each morning for newly detected frequent errors in the wild

### DIFF
--- a/error-monitoring/cron.yaml
+++ b/error-monitoring/cron.yaml
@@ -1,0 +1,13 @@
+cron:
+- description: "File issues for errors in production"
+  url: /_cron/monitor?serviceType=production
+  schedule: every day at 7:00am
+- description: "File issues for errors in 1%/opt-in"
+  url: /_cron/monitor?serviceType=development
+  schedule: every day at 7:02am
+- description: "File issues for errors in experiments"
+  url: /_cron/monitor?serviceType=experiments
+  schedule: every day at 7:04am
+- description: "File issues for errors in nightly"
+  url: /_cron/monitor?serviceType=nightly
+  schedule: every day at 7:06am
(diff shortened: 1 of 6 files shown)


error-monitoring/cron.yaml
```diff
@@ -0,0 +1,13 @@
+cron:
+- description: "File issues for errors in production"
+  url: /_cron/monitor?serviceType=production
+  schedule: every day at 7:00am
```
rcebulko, line 4: Note: times are staggered so they don't all run at once and trigger various API limits

error-monitoring/redacted.env
```diff
@@ -1,4 +1,5 @@
 GITHUB_REPO=ampproject/amphtml
+ISSUE_REPO_NAME=amphtml
```
rcebulko, line 2: When the new repo is created by Joey, this will be updated.

error-monitoring/src/bot.ts
```diff
@@ -32,7 +32,8 @@ export class ErrorIssueBot {
   constructor(
     token: string,
     private repoOwner: string,
-    private repoName: string
+    private repoName: string,
+    private issueRepoName: string | undefined
```
danielrozenberg, line 36: As I'm learning more and more TS, I think this should be `issueRepoName?: string`

rcebulko: Fixed! I always forget about that, since `string?` is the JSDoc form that Typescript rejects.

Issue #890, opened 2020-06-10
[test-case-reporting] Store data from API request to database


Rafer45:
/cc @rcebulko
Part of #881
Depends on #884, #882

PR #942, opened 2020-07-21
[test-case-reporting] Store reports that are POSTed to endpoint

Rafer45:
/cc @rcebulko @estherkim 
Closes https://github.com/ampproject/amp-github-apps/issues/890

### DIFF
--- a/test-case-reporting/index.ts
+++ b/test-case-reporting/index.ts
@@ -44,8 +44,12 @@ app.post('/report', async (req, res) => {
 
   const report: Travis.Report = req.body;
 
-  await testResultRecord.storeTravisReport(report);
-  res.send(`Travis report stored successfully!`);
+  try {
+    await testResultRecord.storeTravisReport(report);
+    res.send(`Travis report stored successfully!`);
+  } catch (e) {
+    res.send(`Failed to send Travis report.`);
+  }
 });
 
 app.listen(PORT, () => {
--- a/test-case-reporting/index.ts
+++ b/test-case-reporting/index.ts
@@ -16,8 +16,11 @@
 
 require('dotenv').config();
 
-import {KarmaReporter} from 'test-case-reporting';
+import {Database, dbConnect} from './src/db';
+import {TestResultRecord} from './src/test_result_record';
+import {Travis} from 'test-case-reporting';
 import express from 'express';
+
 const app = express();
 const PORT = process.env.PORT || 8080;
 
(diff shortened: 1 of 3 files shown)


test-case-reporting/index.ts
```diff
@@ -35,9 +38,18 @@ app.get('/test-results/history/:testCaseId', (req, res) => {
...
 });
 
-app.post('/report', (req, res) => {
-  const report: KarmaReporter.TestResultReport = req.body;
-  res.send(`Travis report! The request body is below.\n\n${report}`);
+app.post('/report', async (req, res) => {
+  const db: Database = dbConnect();
+  const testResultRecord = new TestResultRecord(db);
+
+  const report: Travis.Report = req.body;
```
rcebulko, line 45: A few suggestions here, but I'm rolling them all into one comment so provide a unified example.

1. Creating the Knex DB connection can be done once and used repeatedly by all handlers, so both `db` and `testResultRecord` can live up top, rather than re-creating them in each handler. 
2. Responses should try to use HTTP status codes as best as is reasonable. Adding the `npm` package `http-status-codes` helps make them readable, and Express response objects have `res.setStatus` and `res.sendStatus` helpers to do this. For instance, we'd send a `400 Bad Request` for a report missing key data, a `500 Server Error` for an error during recording the test results, and maybe `202 Created` (or just `200 OK`) for a happy-path report.
3. We should do simple validation on the incoming data and verify the main properties are present.
4. Finally, something I've learned in working with Express: if the error handler errors out, the request often hangs and times out, and Express logs the error to the server log. I've taken to the practice of using a try/catch block for request handlers and ensuring that any error encountered is logged and factored into the response. See examples below of what this might look like.

```javascript
import statusCodes from 'http-status-codes';
// ...
const db = dbConnect();
const record = new TestResultRecord(db);
// ...
app.post('/report', async (req, res) => {
  try {
    const report = validateReport(req.body);
    for (const key of ['job', 'build', 'result']) {
      if (!report[key]) {
        throw new TypeError(`Report payload must include ${key} property`);
      }
    }

    await record.storeTravisReport(report);
    res.sendStatus(statusCodes.CREATED)
  } catch (err) {
    console.warn(err);
    if (err instanceof TypeError) {
      // Missing top-level keys from the report.
      res.status(statusCodes.BAD_REQUEST);
    } else {
      res.status(statusCodes.SERVER_ERROR);
    }

    res.json({ error: err.toString() });
  }
});
```

Rafer45: 1. Done.
2. Done.
3. Done (slightly different from your code sample, PTAL).
4. Done.

In the code sample, it looks like you validate the report twice (once with `validateReport` and once with the for loop). Is this intentional?

rcebulko: Nope that was a mistake

test-case-reporting/index.ts
```diff
@@ -16,11 +16,18 @@
...
+import {Database, dbConnect} from './src/db';
+import {TestResultRecord} from './src/test_result_record';
+import {Travis} from 'test-case-reporting';
 import express from 'express';
+import statusCodes from 'http-status-codes';
+
 const app = express();
 const PORT = process.env.PORT || 8080;
 
+const db: Database = dbConnect();
```
estherkim, line 28: nit: you might not have to import `Database` because `const db = dbConnect()` should already provide the type.

Rafer45: Done.

Issue #889, opened 2020-06-10
[test-case-reporting] Define view templates


Rafer45:
/cc @rcebulko
Part of #881
Depends on #882, #883 

- [ ] Test-case
- [ ] Test-case list
- [ ] All tests for a PR
- [ ] History of runs for one test

Rafer45:
/cc @danielrozenberg @estherkim,
This issue will be unblocked soon, what do y'all think?

PR #950, opened 2020-07-27
[test-case-reporting] Create mustache view templates

Rafer45:
Closes https://github.com/ampproject/amp-github-apps/issues/889

### DIFF
--- a/test-case-reporting/app.ts
+++ b/test-case-reporting/app.ts
@@ -24,6 +24,7 @@ import fs from 'fs';
 import statusCodes from 'http-status-codes';
 
 const MAX_PAGE_SIZE = 500;
+const DEFAULT_PAGE_SIZE = 100;
 
 const app = express();
 const jsonParser = bodyParser.json({limit: '15mb'});
@@ -56,8 +57,8 @@ function handleError(error: Error, res: express.Response): void {
 function extractPageInfo(req: express.Request): PageInfo {
   const {limit, offset} = req.query;
 
-  let limitNum = parseInt(limit.toString(), 10);
-  const offsetNum = parseInt(offset.toString(), 10);
+  let limitNum = limit ? parseInt(limit.toString(), 10) : DEFAULT_PAGE_SIZE;
+  const offsetNum = offset ? parseInt(offset.toString(), 10) : 0;
 
   if (limitNum > MAX_PAGE_SIZE) {
     limitNum = MAX_PAGE_SIZE;
--- a/test-case-reporting/app.ts
+++ b/test-case-reporting/app.ts
@@ -17,8 +17,10 @@
 import {PageInfo, TestRun, Travis} from 'test-case-reporting';
 import {TestResultRecord} from './src/test_result_record';
 import {dbConnect} from './src/db';
+import Mustache from 'mustache';
 import bodyParser from 'body-parser';
 import express from 'express';
+import fs from 'fs';
 import statusCodes from 'http-status-codes';
 
 const MAX_PAGE_SIZE = 500;
@@ -28,8 +30,15 @@ const jsonParser = bodyParser.json({limit: '15mb'});
 const db = dbConnect();
 const record = new TestResultRecord(db);
 
+let testRunListTemplate = '';
 function renderTestRunList(testRuns: Array<TestRun>): string {
-  return `Rendering ${testRuns.length} test runs!`;
+  if (!testRunListTemplate) {
+    testRunListTemplate = fs
+      .readFileSync('./static/test-run-list.html')
+      .toString();
+  }
+
+  return Mustache.render(testRunListTemplate, {testRuns});
 }
 
 function handleError(error: Error, res: express.Response): void {
--- a/test-case-reporting/app.ts
+++ b/test-case-reporting/app.ts
@@ -57,15 +57,10 @@ function handleError(error: Error, res: express.Response): void {
 function extractPageInfo(req: express.Request): PageInfo {
   const {limit, offset} = req.query;
 
-  let limitNum = limit ? parseInt(limit.toString(), 10) : DEFAULT_PAGE_SIZE;
-  const offsetNum = offset ? parseInt(offset.toString(), 10) : 0;
-
-  if (limitNum > MAX_PAGE_SIZE) {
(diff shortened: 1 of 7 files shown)


test-case-reporting/app.ts
```diff
@@ -0,0 +1,128 @@
...
+
+const MAX_PAGE_SIZE = 500;
+
+const app = express();
+const jsonParser = bodyParser.json();
+const db = dbConnect();
+const record = new TestResultRecord(db);
+
+let testRunListTemplate = '';
+function renderTestRunList(testRuns: Array<TestRun>): string {
```
Rafer45, line 34: This would be a little cleaner imo if we used an IIFE to cache testRunListTemplate, but (a) it wouldn't be lazy anymore and (b) it wouldn't be as readable. Thoughts?

test-case-reporting/static/test-run-list.html
```diff
@@ -0,0 +1,17 @@
+<html ⚡>
+<head>
+  <title>Test runs</title>
+</head>
+<body>
+  <ul>
+    {{#testRuns}}
+    <li>Name is "{{testCase.name}}", job num is "{{job.jobNumber}}", status is {{status}}</li>
```
Rafer45, line 8: Let's meet to discuss design of the views (which links & kinds of pages we will have).

rcebulko: Let's start _extremely_ simple, just throw text on the page. Point of this PR is to get the infrastructure in place to render these two views, doesn't matter what they look like yet. That way we save the design and UX discussions for PR separate from the technicalities of Mustache.

test-case-reporting/app.ts
rcebulko: Daniel is strongly opposed to filesystem I/O on startup, and I'm inclined to stick with the lazy approach as well. I think this is fine as-is

test-case-reporting/static/test-run-list.html
Rafer45: Does that look simple enough? Or, do you want me to remove the links

Rafer45:
> Does this work as it's written so far? If so, throw a small screenshot into the PR description for posterity

I don't know; I'm not sure what would be the best way to test it. I can hard-code the template for a quick-and-dirty check, but I'm not sure I see the point there

test-case-reporting/app.ts
Rafer45: SGTM

test-case-reporting/static/test-run-list.html
rcebulko: LGTM

rcebulko:
> > Does this work as it's written so far? If so, throw a small screenshot into the PR description for posterity
> 
> I don't know; I'm not sure what would be the best way to test it. I can hard-code the template for a quick-and-dirty check, but I'm not sure I see the point there, because a hard-coded template working doesn't tell us much about whether it works end-to-end.

You should be able to run it locally using the Cloud SQL Proxy, would definitely be worth figuring out how to do so so that you can test changes, especially as you adjust the frontend

test-case-reporting/app.ts
```diff
@@ -47,8 +57,8 @@ function handleError(error: Error, res: express.Response): void {
 function extractPageInfo(req: express.Request): PageInfo {
   const {limit, offset} = req.query;
 
-  let limitNum = parseInt(limit.toString(), 10);
-  const offsetNum = parseInt(offset.toString(), 10);
+  let limitNum = limit ? parseInt(limit.toString(), 10) : DEFAULT_PAGE_SIZE;
```
rcebulko, line 60:
```suggestion
  const limitNum = limit 
    ? Math.min(parseInt(limit.toString()), MAX_PAGE_SIZE)
    : DEFAULT_PAGE_SIZE;
```
Nit: since the warning about the query size is only in the backend logs and not visible to the client/request, and since we probably won't care much about it on the backend, I'd just skip the logging entirely. Also, I know it used to be the case that parseInt required the second parameter to satisfy linters, but I have a vague memory of it no longer being required (maybe that's from another repo's conventions though, see what the linter says)

Rafer45: I could imagine someone getting confused at why they're only getting 500 results -- is there a reasonable way to send the warning to the client/request? Do you believe it would be worth it?

You're right about the radix; I'll remove it.

Rafer45: I'll apply the suggested changes, and if we decide to report the page size restriction later, we can make an issue.

rcebulko: Not really, since the UI is likely to have a "1-500 of 10000" or something, and any pagination buttons would link to a url with the "correct" params, so upon clicking a link it'll be set properly anyway. TL;DR I don't think it's a common scenario and I don't think anyone would be too surprised to see a page limit in place, and it's a more graceful scenario than the alternative, which is responding with an error. If we wanted to nitpick we could send a redirect to the url with the adjusted parameter value(s) or embed an error/warning message in the page indicating the results were truncated, but I'd put those wayyyyyy down on the list of priorities under the "Does not merit the developer time required" heading :laughing:

test-case-reporting/app.ts
```diff
@@ -57,15 +57,10 @@ function handleError(error: Error, res: express.Response): void {
...
-  if (limitNum > MAX_PAGE_SIZE) {
-    limitNum = MAX_PAGE_SIZE;
-    console.warn(
-      `Maximum query size exceeded. Showing only first ${MAX_PAGE_SIZE} results.`
-    );
-  }
+  const limitNum = limit
+    ? Math.min(parseInt(limit.toString()), MAX_PAGE_SIZE)
+    : DEFAULT_PAGE_SIZE;
+  const offsetNum = offset ? parseInt(offset.toString()) : 0;
```
rcebulko, line 63: Lookin good! Ready for something cool? Destructuring supports default assignment.
```javascript
const {limit = DEFAULT_PAGE_SIZE, offset = 0} = req.query;
const limitNum = Math.min(parseInt(limit.toString()), MAX_PAGE_SIZE);
const offsetNum = parseInt(offset.toString());
```

Rafer45: :O

Issue #949, opened 2020-07-24
[test-case-reporting] Add error handling to GET endpoints


Rafer45:
Part of #881

Rafer45:
Closed by https://github.com/ampproject/amp-github-apps/pull/951

PR #951, opened 2020-07-27
[test-case-reporting] Add error handling to GET endpoints

### LINKED ISSUE #948: [test-case-reporting] Add type errors for bad GET params
Part of #881

Rafer45:
Closes #948 and #949

### DIFF
--- a/test-case-reporting/app.ts
+++ b/test-case-reporting/app.ts
@@ -32,10 +32,21 @@ function renderTestRunList(testRuns: Array<TestRun>): string {
   return `Rendering ${testRuns.length} test runs!`;
 }
 
+function handleError(error: Error, res: express.Response): void {
+  console.warn(error);
+  if (error instanceof TypeError) {
+    // Request params are undefined or the wrong type
+    res.status(statusCodes.BAD_REQUEST);
+  } else {
+    res.status(statusCodes.INTERNAL_SERVER_ERROR);
+  }
+
+  res.json({error: error.toString()});
+}
+
 function extractPageInfo(req: express.Request): PageInfo {
   const {limit, offset} = req.query;
 
-  // TODO(#948): Add type errors for bad GET params
   let limitNum = parseInt(limit.toString(), 10);
   const offsetNum = parseInt(offset.toString(), 10);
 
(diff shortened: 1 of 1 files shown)


test-case-reporting/app.ts
```diff
@@ -57,36 +68,42 @@ app.get('/', (req, res) => {
...
+    handleError(error, res);
   }
 });
 
 app.get('/test-results/history/:testCaseId', async (req, res) => {
-  // TODO(#949): Add error handling to GET endpoints
-  const {testCaseId} = req.params;
-  const {json} = req.query;
+  try {
+    const {testCaseId} = req.params;
```
rcebulko, line 92: nit: these can't throw errors, so let's move them out of the try/catch and have the try encompass as small a set of actions as possbile

test-case-reporting/app.ts
```diff
@@ -32,10 +32,21 @@ function renderTestRunList(testRuns: Array<TestRun>): string {
   return `Rendering ${testRuns.length} test runs!`;
 }
 
+function handleError(error: Error, res: express.Response): void {
+  console.warn(error);
+  if (error instanceof TypeError) {
```
rcebulko, line 37: nit: if you're using this for a catch-all catch block, it's possible for `TypeError`s to bubble up unrelated to pagination. The best thing to do here is probably to extend `Error` with our own `PaginationError` and check for that. I don't really think it's worth that effort though tbh.

Rafer45: I think characterizing bad input as a `TypeError` makes sense -- this isn't only for pagination errors, it also catches bad `buildNumber`s and bad `testCaseId`s. I don't know if I'm being a bit naïve in assuming nothing else will cause `TypeError`s, though.

test-case-reporting/app.ts

test-case-reporting/app.ts
rcebulko: It is certainly possible some other part of the code could throw a typeerror, though using TS obviously makes that much less likely. Like I said, I don't think it's worth the effort to change, just wanted to acknowledge the risk

Rafer45:
Prematurely squash-merged before full reviews (I thought this was #952), please let me know if you have any comments @estherkim & we can revert to address them.

Issue #881, opened 2020-06-10
[test-case-reporting] I2I: Test case reporting database


Rafer45:
/cc @rcebulko

We want to have an app that collects and displays the results of test runs in Travis Builds, to make it easier to tell when a test is flaky (and which commits caused tests to flake/fail). The app will be built with node.js and run on Google AppEngine.

The following diagram depicts the steps in the project and how they depend on each other:
![image](https://user-images.githubusercontent.com/12011689/84295652-e05d5800-ab18-11ea-9321-523602644740.png)

The following diagram depicts the structure of the database that will store test runs:
![image](https://user-images.githubusercontent.com/12011689/84295934-434eef00-ab19-11ea-8fbe-4490395a0ce4.png)

PR #956, opened 2020-07-29
[test-case-reporting] Increase payload limit, fix typos & add missing settings

Rafer45:
Not a very atomic PR, but all the tweaks are small.
Works toward #881

### DIFF
--- a/test-case-reporting/src/db.ts
+++ b/test-case-reporting/src/db.ts
@@ -22,7 +22,7 @@ export function dbConnect(): Database {
   return Knex({
     client: 'pg',
     connection: {
-      host: process.env.DB_UNIX_SOCKET,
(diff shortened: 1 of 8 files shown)


test-case-reporting/src/db.ts
```diff
@@ -22,7 +22,7 @@ export function dbConnect(): Database {
   return Knex({
     client: 'pg',
     connection: {
-      host: process.env.DB_UNIX_SOCKET,
+      host: `${process.env.DB_SOCKET_PATH}/${process.env.DB_INSTANCE_CONNECTION_NAME}`,
```
rcebulko, line 25: `DB_SOCKET_PATH`?

Rafer45: Ah, whoops -- I updated `.env`, but didn't update `redacted.env`.

Done.

PR #957, opened 2020-07-30
[test-case-reporting] Create default timestamps for table rows

Rafer45:


Rafer45:
Under all circumstances, I believe -- I don't think we fill these fields in manually. We would use these fields to sort builds/test runs chronologically &c.

PR #959, opened 2020-07-31
[test-case-results] Add page for displaying builds

Rafer45:


test-case-reporting/types/test-case-reporting.d.ts
```diff
@@ -70,19 +70,20 @@ declare module 'test-case-reporting' {
   /* eslint @typescript-eslint/camelcase: "off" */
   namespace DB {
     export interface Build {
-      // `id` is nullable because is not set when uploading, it is only set
-      // when we get the build from the database.
-      // It is not nullable in the database.
+      // Nullable because it is not set when uploading. Not nullable in the database.
```
Rafer45, line 73: I'm not fully convinced by this rewording -- let me know if you want to change it back.

test-case-reporting/app.ts
```diff
@@ -42,6 +42,15 @@ function renderTestRunList(testRuns: Array<TestRun>): string {
...
 }
 
+let buildListTemplate = '';
+function renderBuildList(builds: Array<Build>): string {
+  if (!buildListTemplate) {
+    buildListTemplate = fs.readFileSync('./static/build-list.html').toString();
+  }
+
+  return Mustache.render(buildListTemplate, {builds});
+}
```
rcebulko, line 52:
```javascript
const templateCache = {};
function render(templateName, data) {
  if (!templateCache[templateName]) {
    templateCache[templateName] = fs.readFileSync(
      `${TEMPLATE_DIR}/${templateName}.html`
    ).toString();
  }

  return Mustache.render(templateCache[templateName], data)
}
...
render('build-list', {builds});
...
render('test-run-list', {testRuns});
```

test-case-reporting/src/test_result_record.ts
```diff
@@ -31,6 +31,18 @@ import md5 from 'md5';
 type QueryFunction = (q: QueryBuilder) => QueryBuilder;
 
 /* eslint-disable @typescript-eslint/camelcase */
+
+/**
+ * Creates a Build object from a build row of the build table.
+ * @param dbBuild A build row of the build table
+ */
+function getBuildFromDbBuild({commit_sha, build_number}: DB.Build): Build {
```
rcebulko, line 39: tbh I think the level of indirection here isn't worth splitting this out, given how simple it is

test-case-reporting/types/test-case-reporting.d.ts
rcebulko: I think it was fine before

test-case-reporting/types/test-case-reporting.d.ts
```diff
@@ -102,7 +106,7 @@ declare module 'test-case-reporting' {
     }
 
     export interface TestRun {
-      // See comment under `DB.Build.id`
+      // Nullable because it is not set when uploading. Not nullable in the database.
```
rcebulko, line 109: I think rather than adding comments in each case, one comment at the top of the `DB` namespace would be enough to explain that some fields will be optional because they are set by the DB automatically

test-case-reporting/app.ts
Rafer45: I like this approach, but how do we typecheck it? `renderBuildList` takes an array of Builds, while `renderTestRuns` takes an array of TestRuns. We _could_ make a type alias for Array<TestRuns> & Array<Builds> & whatever else we eventually want to render.

test-case-reporting/src/test_result_record.ts
Rafer45: I don't think I agree, but I don't feel strongly about this.

Done removing getBuildFromDbBuild.

test-case-reporting/types/test-case-reporting.d.ts
Rafer45: Removed.

test-case-reporting/types/test-case-reporting.d.ts
Rafer45: Good idea. Done.

test-case-reporting/app.ts
Rafer45: Done putting that approach in, PTAL at how I (kind of) did the typechecking.

test-case-reporting/app.ts
```diff
@@ -23,23 +23,26 @@ import express from 'express';
...
 const db = dbConnect();
 const record = new TestResultRecord(db);
 
-let testRunListTemplate = '';
-function renderTestRunList(testRuns: Array<TestRun>): string {
-  if (!testRunListTemplate) {
-    testRunListTemplate = fs
-      .readFileSync('./static/test-run-list.html')
+const templateCache: Record<string, string> = {};
+function render(templateName: string, data: Renderable): string {
```
rcebulko, line 38: Tbh I'd just use `Object` here. More flexible when you start adding to the page context, and the typechecking is less essential since you'll quickly see the results in a rendered template if you screwed something up. We can always declare types in the view itself, before passing to the render function.

One alternative which I don't support but will list here for completeness would be defining wrappers for renderTestRunList and renderBuildList, which call render with the proper templates and have stricter types. I don't think that's cleaner or helpful though.

Rafer45: Typescript complains when I use the type `Object`, telling me to use `Record<string, any>`, but then it complains about `any`.

How do you feel about making the type `Renderable = Record<string, (Renderable | string)>`?

rcebulko: Use unknown instead of any

rcebulko: `Record<string, unknown>`

test-case-reporting/app.ts
```diff
@@ -68,8 +71,20 @@ function extractPageInfo(req: express.Request): PageInfo {
   };
 }
 
-app.get('/', (req, res) => {
-  res.send('Hello, world, 2.0!');
+app.get('/', async (req, res) => {
```
rcebulko, line 74:
```suggestion
app.get(['/', '/builds'], async (req, res) => {
```

test-case-reporting/src/test_result_record.ts
```diff
@@ -273,4 +274,22 @@ export class TestResultRecord {
...
+  async getRecentBuilds({limit, offset}: PageInfo): Promise<Array<Build>> {
+    const dbBuilds = await this.db<DB.Build>('builds')
+      .orderBy('started_at', 'DESC')
+      .limit(limit)
+      .offset(offset);
+
+    /* eslint-disable @typescript-eslint/camelcase */
+    return dbBuilds.map(({commit_sha, build_number}) => ({
+      commitSha: commit_sha,
+      buildNumber: build_number,
```
rcebulko, line 291: Is the pass/fail status of a build available?

test-case-reporting/static/build-list.html
```diff
@@ -0,0 +1,17 @@
+<html ⚡>
+<head>
+  <title>Builds</title>
+</head>
+<body>
+  <ul>
+  {{#builds}}
+  <li><a href="/test-results/build/{{buildNumber}}">Build {{buildNumber}}</a></li>
```
rcebulko, line 8: nit: intendation, possibly break across multiple lines if past 80 characters. Also, not sure if you did this elsewhere, but since the page isn't an AMP page it shouldn't have the lightning bolt in the html tag. just want a `<!DOCTYPE html>` maybe

test-case-reporting/static/build-list.html
```diff
@@ -0,0 +1,17 @@
+<html ⚡>
+<head>
+  <title>Builds</title>
```
rcebulko, line 3:
```suggestion
  <title>Latest Builds</title>
```

test-case-reporting/static/build-list.html
Rafer45: Done.

test-case-reporting/app.ts

test-case-reporting/app.ts

test-case-reporting/src/test_result_record.ts
Rafer45: Not at the moment. To know the actual pass/fail/err status of the build, we would need to call the Travis API.

We could estimate the status by checking whether a test failure was reported for the build, which I think would be more useful in our application. What do you think?

test-case-reporting/static/build-list.html
Rafer45: Build-list is for any build list (it's meant to be modular). I'll add a title field to the passed object.

Rafer45: Done adding the title field, PTAL and let me know what you think.

test-case-reporting/src/test_result_record.ts
rcebulko: I agree that would be useful, albeit tricky to query efficiently. I'd say let's punt on build status for now to avoid that complexity, and it can be looked at later if needed.

rcebulko: I'll note that at the time a Job is being inserted, we probably have the info necessary to report passed/failed. A job list may end up being more useful than a build list, but again for now I think punting on this is fine.

test-case-reporting/app.ts
```diff
@@ -68,8 +69,20 @@ function extractPageInfo(req: express.Request): PageInfo {
...
+app.get(['/', '/builds'], async (req, res) => {
+  const {json} = req.query;
+
+  try {
+    const builds = await record.getRecentBuilds(extractPageInfo(req));
+
+    if (json) {
+      res.json({builds});
+    } else {
+      res.send(render('build-list', {title: 'Latest Builds', builds}));
```
rcebulko, line 81: This title field is a good pattern to use for the test run list pages as well, so you could set the title as "Test History - 'the thing...'" or "Test Runs - Build #12345" etc.

rcebulko: (feel free to punt this to a later PR)

test-case-reporting/app.ts
```diff
@@ -25,21 +25,22 @@ import statusCodes from 'http-status-codes';
...
 const jsonParser = bodyParser.json({limit: '15mb'});
 const db = dbConnect();
 const record = new TestResultRecord(db);
 
-let testRunListTemplate = '';
-function renderTestRunList(testRuns: Array<TestRun>): string {
-  if (!testRunListTemplate) {
-    testRunListTemplate = fs
-      .readFileSync('./static/test-run-list.html')
+const templateCache: Record<string, string> = {};
```
estherkim, line 35: Where does `Record` come from?

rcebulko: It's a native TypeScript type

Rafer45: Comes from TS -- https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkt

estherkim: Oh neat. This might be useful if the types were non-basic but they're just strings. Otherwise I think the Map type gives you nice utilities like has, get and set.

Issue #886, opened 2020-06-10
[test-case-reporting] Cron config & stubbed endpoint


Rafer45:
/cc @rcebulko
Part of #881

PR #961, opened 2020-08-03
[test-case-reporting] Stub cron config, add cron yaml

Rafer45:
Closes https://github.com/ampproject/amp-github-apps/issues/886


test-case-reporting/app.ts
```diff
@@ -130,4 +130,9 @@ app.post('/report', jsonParser, async (req, res) => {
   }
 });
 
+app.get('/_cron/compute-stats', async (req, res) => {
+  const {count} = req.query;
+  res.send(`Computing pass/fail % for past ${count} runs`);
```
rcebulko, line 135: nit: while this is a stub, also log this text so you can verify in the logs that the cron job is running as expected.

PR #962, opened 2020-08-03
[test-case-reporting] Simplify logic in extractPageInfo

Rafer45:


test-case-reporting/app.ts
```diff
@@ -55,12 +55,13 @@ function handleError(error: Error, res: express.Response): void {
...
-  const {limit, offset} = req.query;
+  const {limit = DEFAULT_PAGE_SIZE, offset = 0} = req.query;
 
-  const limitNum = limit
-    ? Math.min(parseInt(limit.toString()), MAX_PAGE_SIZE)
-    : DEFAULT_PAGE_SIZE;
-  const offsetNum = offset ? parseInt(offset.toString()) : 0;
+  // This may look odd because the default limit/offset values are uselessly
+  // turned from ints, to strings, back into ints; however, we need the
+  // toString because query parameters aren't strings.
```
rcebulko, line 62: TBH I don't think this comment is necessary. It's clear they come from the query params, which require toString, and trying to remove that call will make TS complain.

Issue #964, opened 2020-08-05
[test-case-reporting] Make test run list pages prettier and more usable


Rafer45:
Part of #881

PR #965, opened 2020-08-05
[test-case-reporting] Make test run list page prettier with CSS

Rafer45:
Works toward #964. 
Looks like this:
image


test-case-reporting/app.ts
```diff
@@ -128,7 +131,10 @@ app.get('/test-results/history/:testCaseId', async (req, res) => {
       res.send(
         render('test-run-list', {
           title: `Test Runs for test case "${testCaseName}"`,
-          testRuns,
+          testRuns: testRuns.map((testRun, index) => ({
+            testRun,
+            index,
+          })),
```
rcebulko, line 137: Nit: I suspect this could fit cleanly on one line, which in this case would be less visual clutter. If not, nbd. But add a comment explaining why this is needed, and consider a simple helper like
```javascript
function enumeratedTestRuns(testRuns) {
  return testRuns.map((testRun, index) => ({testRun, index});
}
```

You'll end up reusing this for the flaky-sorted list view as well.

test-case-reporting/static/css/surface.min.css
```diff
@@ -0,0 +1 @@
...:#4caf50}.color--light-green{color:#8bc34a}.color--lime{color:#cddc39}.color--river{color:#3498db}.color--belize{color:#2980b9}.color--asphalt{color:#34495e}.color--midnight-blue{color:#2c3e50}.color--blue{color:#2196f3}.color--light-blue{color:#03a9f4}.color--cyan{color:#00bcd4}.color--teal{color:#009688}.color--alizarin{color:#e74c3c}.color--pomegranate{color:#c0392b}.color--red{color:#f44336}.color--carrot{color:#e67e22}.color--pumpkin{color:#d35400}.color--dull-orange{color:#f39c12}.color--orange{color:#ff9800}.color--blood-orange{color:#ff5722}.color--amber{color:#ffc107}.color--sunflower{color:#f1c40f}.color--yellow{color:#ffeb3b}.color--amethyst{color:#9b59b6}.color--plum{color:#8e44ad}.color--purple{color:#9c27b0}.color--deep-purple{color:#673ab7}.color--pink{color:#e91e63}.color--indigo{color:#3f51b5}.color--brown{color:#795548}.color--grey{color:#9e9e9e}.color--gun-metal{color:#607d8b}.color--asbestos{color:#7f8c8d}.color--concrete{color:#95a5a6}.color--silver{color:#bdc3c7}.color--clouds{color:#dde4e6}.color--paper{color:#efefef}.color--white{color:#fff}.color--black{color:#212121}.color--primary{color:#1abc9c}.color--secondary{color:#28e1bd}.color--accent{color:#ffeb3b}
```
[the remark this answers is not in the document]
Rafer45: Done.

test-case-reporting/app.ts
```diff
@@ -67,6 +67,15 @@ function extractPageInfo(req: express.Request): PageInfo {
   };
 }
 
+// We need the testRun index in mustache to give each collapsable
+// its own id.
+function enumerateTestRun(
```
rcebulko, line 72: Does this method take a single testRun-index pair as input, or does it take a list of testRuns?

Rafer45: It takes a single testRun-index pair as an input, because I find `testRuns.map(enumerateTestRun)` to be clearer than `enumerateTestRuns(testRuns)`. I'm open to changing it.

rcebulko:
```typescript
const testRuns: Array<TestRun> = [/* stuff */];
function enumerateTestRun(testRun: TestRun, index: number) {/* stuff */}
testRuns.map(enumerateTestRun); // Where is the value of index coming from?
```

rcebulko:
I think one thing that would work well visually is if each row started with a small tile indicating the status, like a green/red/yellow/orange icons or something similar (or even just a yellow box with the word 'SKIP'). Something like this might work/be a good starting point: 
```css
.status-badge {
  width: 60px;
  padding: 2.5px;
  display: inline-block;
  font-family: sans-serif;
  font-variant: small-caps;
  color: white;
}
.status-pass { background-color: green; }
.status-fail { background-color: orange; }
.status-skip { background-color: gray; }
.status-error { background-color: red; }
```

```html
<span class='status-badge status-pass'>Pass</span>
<span><a href=''>this is a test case blah blah</a></span>
```

Hopefully this gets close to working out of the box, have not tested yet

Rafer45:
> Hopefully this gets close to working out of the box, have not tested yet

It mostly worked, but the badge looked a little weird, so I tried applying the color to the background of the collapsible -- let me know what you think.

test-case-reporting/app.ts
```diff
@@ -76,6 +76,13 @@ function enumerateTestRun(
   return {testRun, index};
 }
 
+function lowerCaseStatus({status, ...rest}: TestRun): any {
```
rcebulko, line 79: Are you going to be using Handlebars or Mustache?

rcebulko:
> > Hopefully this gets close to working out of the box, have not tested yet
> 
> It mostly worked, but the badge looked a little weird, so I tried applying the color to the background of the collapsible -- let me know what you think.

Screenshot? :)

test-case-reporting/app.ts
Rafer45: I'm keeping mustache for now.

Rafer45:
> > > Hopefully this gets close to working out of the box, have not tested yet
> > 
> > 
> > It mostly worked, but the badge looked a little weird, so I tried applying the color to the background of the collapsible -- let me know what you think.
> 
> Screenshot? :)

Of what looked weird, or to the collapsibles with background colors? I have screenshots of the latter at the top.

rcebulko:
The reason I prefer the badges over just coloring the collapsibles is a) the color-to-status relation is non-obvious unless you provide a legend or something and b) color-only approach is very inaccessible to color-impaired users and screen-readers

Rafer45:
Makes sense, let's go with the badge approach. It currently looks like this -- 
image

rcebulko:
Try adding the following rules to `.status-badge`

```css
margin-right: 10px;
text-align: center;
font-weight: bold;
```

And try to put it inside the collabsible label section so it lines up with the test name, then we can remove the "had status X" at the end

Rafer45:
Would you support swapping the colors for fail and error, so that error is orange and fail is yellow?

rcebulko:
> @rcebulko, would you support swapping the colors for fail and error, so that error is orange and fail is yellow?

However you have it in the screenshot in the description right now looks good to me :+1:

Rafer45:
Ok, keeping the colors like that, then.

rcebulko:
I like the collapsibles too. Right now it's pretty empty, but down the line I can see including test case stats, metadata about the build/PR, duration, and even spaces where people could leave comments with related issues, error info, or reasons why something was skipped. Seems super flexible

test-case-reporting/static/test-run-list.html
```diff
@@ -1,23 +1,36 @@
...
+        <h1>{{title}}</h1>
+        {{#testRuns}}
+          <div class="collapsible-wrap card no-pad">
+            <input type="checkbox" id="collapsible-{{index}}">
+            <label for="collapsible-{{index}}">
+              <span class='status-badge status-{{testRun.status}}'>{{testRun.status}}</span>&nbsp;
+              <a href="/test-results/history/{{testRun.testCase.id}}">{{testRun.testCase.name}}</a>
+            </label>
+            <div class="collapsible-{{index}}-area">
+              <p>Job num is {{testRun.job.jobNumber}}</p>
```
rcebulko, line 23: `Job: {{...}}`, this could link to the Travis job log
`Build: {{...}}`

test-case-reporting/app.ts
Rafer45: Not sure how I missed this.

I don't understand the suggestion, are you saying `enumerateTestRun` should be declared inside the arrow function of the app.get? It is used by two different blocks so it needs to be declared outside.

test-case-reporting/static/test-run-list.html
Rafer45: To link to the Travis job log, we need the Travis API, which would take a bit of setting up (getting the API key and so on). Seems like it would be worth it, but I'm not sure I want to start that at this moment of the internship.

We could also add links to the job & build objects and get those through the wire (they're easy to get on the Travis end), that would be simpler but less flexible than using the API if we need it for other things.

Rafer45: Regardless, I would want to move that to a separate PR.

test-case-reporting/app.ts
rcebulko: Cleared up offline, resolving

test-case-reporting/static/test-run-list.html
rcebulko: Let's at least update the text/formatting, but leave the links as-is then

Rafer45: Done.

test-case-reporting/app.ts
```diff
@@ -67,6 +67,24 @@ function extractPageInfo(req: express.Request): PageInfo {
...
+// We need the testRun index in mustache to give each collapsable
+// its own id.
+function enumerateTestRun(
+  testRun: TestRun,
+  index: number
+): {testRun: TestRun; index: number} {
+  return {testRun, index};
+}
+
+function lowerCaseStatus({status, ...rest}: TestRun): any {
```
rcebulko, line 79: At some point I think I'd prefer to see a single helper with something like `testRunsForView` that takes a list of test runs and does everything, including indexes, capitalization, and possible eventually finding build IDs or constructing other URLs. Fine with punting on it for now

test-case-reporting/types/test-case-reporting.d.ts
```diff
@@ -110,7 +111,9 @@ declare module 'test-case-reporting' {
       duration_ms: number;
     }
 
-    export interface BigJoin extends Build, Job, TestCase, TestRun {}
+    export interface BigJoin extends Build, Job, TestCase, TestRun {
```
danielrozenberg, line 114: I know this wasn't done in this PR but I'm not a big fan of this name??? 😅
What is it used for exactly? Give it a descriptive name

rcebulko: Lmao I missed this name. `TestRunWithJob` or something maybe

Rafer45: `JoinOfAllTables`? Because it's actually a `TestRunWithJobAndBuildAndTestCase`...

rcebulko: The type doesn't need to explicitly mention it's based on DB tables or a join. TestRunWithMeta, TestRunWithJobAndBuild would both be more appropriate imo

test-case-reporting/app.ts
Rafer45: I like that; agreed on punting it, though.

test-case-reporting/types/test-case-reporting.d.ts
Rafer45: Well, we only need the type because we're using a DB table join -- if we weren't doing that, we wouldn't need this type -- so I don't think that coupling would be too bad.

I think I like `TestRunWithJobAndBuild` or `TestRunWithFullInfo`. Thoughts?

Rafer45: (We can do that in a separate PR, pushing for now...)

rcebulko: `TestRunWithJobAndBuild` SGTM

Rafer45: Done, PTAL at https://github.com/ampproject/amp-github-apps/pull/973

Issue #892, opened 2020-06-10
[test-case-reporting] Compute & record pass/fail%


Rafer45:
/cc @rcebulko
Part of #881
Depends on #886

PR #967, opened 2020-08-07
[test-case-reporting] Add endpoint for cron task

Rafer45:
Closes https://github.com/ampproject/amp-github-apps/issues/892

### DIFF
--- a/test-case-reporting/app.ts
+++ b/test-case-reporting/app.ts
@@ -15,6 +15,7 @@
  */
 
 import {Build, PageInfo, TestRun, Travis} from 'test-case-reporting';
+import {TestCaseStats} from './src/test_case_stats';
 import {TestResultRecord} from './src/test_result_record';
 import {dbConnect} from './src/db';
 import Mustache from 'mustache';
@@ -142,12 +143,19 @@ app.post('/report', jsonParser, async (req, res) => {
 });
 
 app.get('/_cron/compute-stats', async (req, res) => {
+  const testCaseStats = new TestCaseStats(db);
   const {count} = req.query;
 
-  const message = `Computing pass/fail % for past ${count} runs`;
-  res.send(message);
+  try {
+    const sampleSize = parseInt(count.toString());
+    await testCaseStats.updateStats(sampleSize);
 
-  console.log(message);
+    const message = `Computed pass/fail % for past ${count} runs`;
+    res.send(message);
+    console.log(message);
+  } catch (error) {
+    handleError(error, res);
+  }
 });
 
 export {app};
--- a/test-case-reporting/app.ts
+++ b/test-case-reporting/app.ts
@@ -147,6 +147,12 @@ app.get('/_cron/compute-stats', async (req, res) => {
   const {count} = req.query;
 
   try {
+    // This header is set by App Engine when running cron tasks, and is
+    // stripped out of external requests.
+    if (!req.header('X-Appengine-Cron') && process.env.NODE_ENV !== 'dev') {
+      throw new Error('Attempted external request to a cron endpoint');
+    }
+
     const sampleSize = parseInt(count.toString());
 
     if (Number.isNaN(sampleSize)) {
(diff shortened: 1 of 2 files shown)


test-case-reporting/app.ts
```diff
@@ -142,12 +143,19 @@ app.post('/report', jsonParser, async (req, res) => {
 });
 
 app.get('/_cron/compute-stats', async (req, res) => {
+  const testCaseStats = new TestCaseStats(db);
```
rcebulko, line 146: This can be created up top outside of the request handler, just like the `TestResultRecord`

test-case-reporting/app.ts
```diff
@@ -142,12 +143,19 @@ app.post('/report', jsonParser, async (req, res) => {
...
 
-  const message = `Computing pass/fail % for past ${count} runs`;
-  res.send(message);
+  try {
+    const sampleSize = parseInt(count.toString());
+    await testCaseStats.updateStats(sampleSize);
 
-  console.log(message);
+    const message = `Computed pass/fail % for past ${count} runs`;
+    res.send(message);
```
rcebulko, line 154: This is a cron job, so there is no receiving browser. Let's `.sendStatus(statusCodes.OK)` on success and similarly send just an error status on failure, but still log the message or error for debugging purposes.

test-case-reporting/app.ts
```diff
@@ -142,12 +143,19 @@ app.post('/report', jsonParser, async (req, res) => {
...
 
 app.get('/_cron/compute-stats', async (req, res) => {
+  const testCaseStats = new TestCaseStats(db);
   const {count} = req.query;
 
-  const message = `Computing pass/fail % for past ${count} runs`;
-  res.send(message);
+  try {
+    const sampleSize = parseInt(count.toString());
+    await testCaseStats.updateStats(sampleSize);
```
rcebulko, line 151: If no value is provided, or an invalid value that produces `NaN`, this will explode. Lets sanitize here, and also gate it so that it can't be invoked by any ol' schmuck who tries to DDoS and overload the DB with stat calculation.

```javascript
// This header is set by App Engine when running cron tasks, and is
// stripped out of external requests.
if (!req.header('X-Appengine-Cron') && process.env.NODE_ENV !== 'dev') {
  throw new Error('Attempted external request to a cron endpoint');
}
```

rcebulko: (See `amp-github-apps/owners/info-server.js:147` to see this in action)

test-case-reporting/app.ts
Rafer45: Done. Why do we want to create it up top, if it's only used by this endpoint?

rcebulko: Because then it's only created once and shared between requests, which is mildly preferable

test-case-reporting/app.ts

test-case-reporting/app.ts
Rafer45: Done.

test-case-reporting/app.ts
```diff
@@ -144,10 +146,26 @@ app.post('/report', jsonParser, async (req, res) => {
...
-  const message = `Computing pass/fail % for past ${count} runs`;
-  res.send(message);
+  try {
+    // This header is set by App Engine when running cron tasks, and is
+    // stripped out of external requests.
+    if (!req.header('X-Appengine-Cron') && process.env.NODE_ENV !== 'dev') {
+      throw new Error('Attempted external request to a cron endpoint');
+    }
+
+    const sampleSize = parseInt(count.toString());
```
danielrozenberg, line 156: Prefer to use `Number(count.toString())` (I'm not even sure if you really need to call `.toString` in this case, you can test it out - `parseInt` is too forgiving of malformed number-strings

rcebulko: `toString` is required by TypeScript since query params can potentially be arrays

Rafer45:
Pushing to unblock.

@estherkim we can address any comments in a future PR

test-case-reporting/app.ts
Rafer45: TypeScript didn't complain about the ToString for the argument to Number() -- looks like it doesn't worry about the type as much.

PR #977, opened 2020-08-13
[test-case-reporting] Add view for test cases, sorted by stat

Rafer45:
Closes https://github.com/ampproject/amp-github-apps/issues/968

### DIFF
--- a/test-case-reporting/app.ts
+++ b/test-case-reporting/app.ts
@@ -128,6 +128,38 @@ app.get('/test-results/build/:buildNumber', async (req, res) => {
   }
 });
 
+app.get('/test-cases/stats/:stat', async (req, res) => {
+  const {stat} = req.params;
+  const {json, count = 10} = req.query;
+
+  try {
+    const sampleSize = Number(count);
+
+    if (Number.isNaN(sampleSize)) {
+      throw new TypeError('sampleSize is not a number');
+    }
+
+    const testCases = await record.getTestCasesSortedByStat(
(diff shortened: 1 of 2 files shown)


test-case-reporting/app.ts
```diff
@@ -128,6 +128,38 @@ app.get('/test-results/build/:buildNumber', async (req, res) => {
...
 
+app.get('/test-cases/stats/:stat', async (req, res) => {
+  const {stat} = req.params;
+  const {json, count = 10} = req.query;
+
+  try {
+    const sampleSize = Number(count);
+
+    if (Number.isNaN(sampleSize)) {
+      throw new TypeError('sampleSize is not a number');
```
rcebulko, line 139:
```suggestion
      throw new TypeError(`Expected a number for parameter 'count'; got ${count}`);
```

test-case-reporting/static/test-case-list.html
```diff
@@ -0,0 +1,24 @@
+<!DOCTYPE html>
```
rcebulko, line 1: Does this not use the same test-case-list template all the other views use?

Rafer45: There was no test-case-list template yet, only test-run-list and build-list templates.

test-case-reporting/app.ts
Rafer45: Done. I'll try to make my error messages more informative in the future :+1:

PR #978, opened 2020-08-17
[error-monitoring] Prepare for automatic issue filing but do not enable cron tasks yet

rcebulko:
Fixes some configuration around automated error monitoring, filing issues in https://github.com/ampproject/error-reporting


rcebulko:
Closing, since we are no longer planning to auto-file issues

PR #981, opened 2020-08-20
[test-case-reporting] Process e2e test reports correctly

Rafer45:


test-case-reporting/src/test_result_record.ts
```diff
@@ -172,17 +178,24 @@ export class TestResultRecord {
...
-        const testCaseName = this.testCaseName(result);
-        return {
-          ...result,
-          testCaseId: md5(testCaseName),
-          testCaseName,
-        };
-      });
+    let resultList: Array<KarmaReporter.TestResult>;
+
+    if (job.testSuiteType === 'e2e') {
```
Rafer45, line 183: We could add a field to reports describing the structure of the reporter (mocha vs karma) -- I think that would be clearer than the (currently implicit) connection between e2e tests and the mocha reporter.

test-case-reporting/src/test_result_record.ts
```diff
@@ -29,6 +31,10 @@ import {QueryBuilder} from 'knex';
 import md5 from 'md5';
 
 type QueryFunction = (q: QueryBuilder) => QueryBuilder;
+interface FormattedResult extends KarmaReporter.TestResult {
```
rcebulko, line 34: Nice

test-case-reporting/src/test_result_record.ts
rcebulko: I agree. We could then make this if(e2e)-elseif(unit|integration)-else(error-unknown)

Rafer45: Oh, I mean changing reports from this:
```js
{
  job: // job info
  build: // build info
  results: // results info in some shape
}
```
to this:
```js
{
  job: // job info
  build: // build info
  results: // results info in some shape
  reportType: // mocha or karma
}
```
It's redundant but it's more readable imo, thoughts?

PR #1005, opened 2020-10-28
[bundle-size] Update all dependencies, fix breaking changes, and modify tests to assert calls to Octokit instead of network calls

danielrozenberg:
Tested using a local database, dev instance of the app, and a separate repo


rcebulko (approved):
Thanks for taking care of this, Octokit/Probot updates are always the worst

estherkim (approved):

PR #1007, opened 2020-10-29
[bundle-size-chart] Move `@octokit/types` from `devDependencies` to `dependencies`

danielrozenberg:


rcebulko (approved):
HeLlO mY nAmE iS oCtOkIt

PR #1009, opened 2020-10-29
[test-status] Update dependencies and modify tests to expect Octokit calls instead of network calls

danielrozenberg:

PR #1011, opened 2020-11-09
[pr-deploy] Update dependencies and tests

estherkim:
- Update dependencies
- Rewrite tests to use nock to intercept HTTP requests instead of mocking octokit, which is an internal dependency of probot
- Linting

### DIFF
--- a/pr-deploy/src/app.ts
+++ b/pr-deploy/src/app.ts
@@ -85,18 +85,17 @@ function initializeDeployment(app: Application) {
       (context.github as unknown) as Octokit,
       context.payload.check_run.head_sha
     );
-    await pr.deploymentInProgress();
-    const travisBuild = await pr.getTravisBuildNumber();
-    await unzipAndMove(travisBuild)
-      .then(async bucketUrl => {
-        await pr.deploymentCompleted(
-          bucketUrl,
-          `${BASE_URL}amp_dist_${travisBuild}/`
-        );
-      })
-      .catch(async e => {
-        await pr.deploymentErrored(e);
-      });
+    try {
+      await pr.deploymentInProgress();
+      const travisBuild = await pr.getTravisBuildNumber();
+      const bucketUrl = await unzipAndMove(travisBuild);
+      await pr.deploymentCompleted(
+        bucketUrl,
+        `${BASE_URL}amp_dist_${travisBuild}/`
+      );
+    } catch (e) {
+      await pr.deploymentErrored(e);
+    };
   });
 }
 
--- a/pr-deploy/src/app.ts
+++ b/pr-deploy/src/app.ts
@@ -88,14 +88,14 @@ function initializeDeployment(app: Application) {
     await pr.deploymentInProgress();
     const travisBuild = await pr.getTravisBuildNumber();
     await unzipAndMove(travisBuild)
-      .then(bucketUrl => {
-        pr.deploymentCompleted(
+      .then(async bucketUrl => {
+        await pr.deploymentCompleted(
           bucketUrl,
           `${BASE_URL}amp_dist_${travisBuild}/`
         );
       })
-      .catch(e => {
-        pr.deploymentErrored(e);
+      .catch(async e => {
+        await pr.deploymentErrored(e);
       });
   });
 }
--- a/pr-deploy/src/app.ts
+++ b/pr-deploy/src/app.ts
@@ -18,6 +18,7 @@ import {Application} from 'probot';
 import express, {IRouter} from 'express';
 import {PullRequest} from './github';
 import {unzipAndMove} from './zipper';
+import {Octokit} from '@octokit/rest';
 
 const BASE_URL = `https://storage.googleapis.com/${process.env.SERVE_BUCKET}/`;
 
@@ -32,8 +33,8 @@ function initializeCheck(app: Application) {
     'pull_request.reopened',
   ], async context => {
     const pr = new PullRequest(
-      context.github,
-      context.payload.pull_request.head.sha,
+      (context.github as unknown) as Octokit,
+      context.payload.pull_request.head.sha
     );
     return pr.createOrResetCheck();
   });
@@ -50,7 +51,9 @@ function initializeRouter(app: Application) {
   router.post('/travisbuilds/:travisBuild/headshas/:headSha/:result',
     async(request, response) => {
       const {travisBuild, headSha, result} = request.params;
-      const github = await app.auth(Number(process.env.INSTALLATION_ID));
+      const github = ((await app.auth(
+        Number(process.env.INSTALLATION_ID)
+      )) as unknown) as Octokit;
       const pr = new PullRequest(github, headSha);
       switch (result) {
         case 'success':
@@ -79,8 +82,8 @@ function initializeDeployment(app: Application) {
     }
 
     const pr = new PullRequest(
-      context.github,
-      context.payload.check_run.head_sha,
+      (context.github as unknown) as Octokit,
+      context.payload.check_run.head_sha
     );
     await pr.deploymentInProgress();
     const travisBuild = await pr.getTravisBuildNumber();
--- a/pr-deploy/src/github.ts
+++ b/pr-deploy/src/github.ts
@@ -14,11 +14,24 @@
  * limitations under the License.
  */
 
+import {GetResponseTypeFromEndpointMethod as GetType} from '@octokit/types';
 import {Octokit} from '@octokit/rest';
-import {GitHubAPI} from 'probot/lib/github';
 
+const octokit = new Octokit();
+type ChecksCreateParams = GetType<typeof octokit.ChecksCreateParams>;
+type ChecksListForRefResponseCheckRunsItem = GetType<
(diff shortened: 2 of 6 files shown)


pr-deploy/src/app.ts
```diff
@@ -79,20 +82,20 @@ function initializeDeployment(app: Application) {
...
+      .then(async bucketUrl => {
+        await pr.deploymentCompleted(
           bucketUrl,
           `${BASE_URL}amp_dist_${travisBuild}/`
         );
       })
-      .catch(e => {
-        pr.deploymentErrored(e);
+      .catch(async e => {
+        await pr.deploymentErrored(e);
```
rcebulko, line 98: Nit: Since you're already tweaking this, consider making it all async

```javascript
try {
  const bucketUrl = await unzipAndMove(travisBuild);
  await pr.deploymentCompleted(bucketUrl, `${BASE_URL}amp_dist_${travisBuild}/`);
} catch (e) {
  await pr.deploymentErrored(e);
}
```

pr-deploy/src/github.ts
```diff
@@ -32,8 +45,8 @@ export class PullRequest {
   private github: Octokit;
   private headSha: string;
 
-  constructor(github: GitHubAPI, headSha: string) {
-    this.github = github as unknown as Octokit;
+  constructor(github: Octokit, headSha: string) {
+    this.github = github as Octokit;
```
rcebulko, line 49: Since `github` is already declared to be of type `Octokit`, is this `as Octokit` clause still necessary?

pr-deploy/src/types/unzip-stream.d.ts
```diff
@@ -14,11 +14,11 @@
  * limitations under the License.
  */
 declare module 'unzip-stream' {
-  import stream = require('stream');
+  import stream from 'stream';
```
rcebulko, line 17:
```suggestion
  import {Writable} from 'stream';
```

pr-deploy/test/app.test.ts
```diff
@@ -17,115 +17,145 @@
...
+      .mockReturnValue(
+        Promise.resolve('gs://serving-bucket/travisBuildNumber')
+      ),
   };
 });
 
-import {Application} from 'probot';
+process.env.GH_CHECK = 'test-check';
+process.env.GH_OWNER = 'test-owner';
+process.env.GH_REPO = 'test-repo';
```
rcebulko, line 30: This may or may not be relevant for this app, but in the past I've used `jest-preload.ts` to stub values for environment variables like `test_org`, `__TOKEN__`, etc. that are consumed across tests. Useful if these env vars are stubbed elsewhere

pr-deploy/test/app.test.ts
```diff
@@ -17,115 +17,145 @@
...
 import prDeployAppFn from '../src/app';
-import express from 'express';
-import request from 'supertest';
-import Webhooks, {
-  WebhookPayloadPullRequest,
-  WebhookPayloadCheckRun} from '@octokit/webhooks';
+import Webhooks from '@octokit/webhooks';
+import {Probot, ProbotOctokit} from 'probot';
+
+const apiUrl = 'https://api.github.com';
```
rcebulko, line 37: nit: maybe capitalize?

pr-deploy/test/app.test.ts
```diff
@@ -17,115 +17,145 @@
...
+      }),
+    });
+    probot.load(prDeployAppFn);
+  });
 
-    server = express();
-    server.use(app.router);
+  afterEach(() => {
+    nock.cleanAll();
+    nock.enableNetConnect();
```
rcebulko, line 58: I've encountered issues in the past where a nocked endpoint is never called (including the reply handler that has `expects` on the response), and as a result the test passes while actually failing silently. `nock.isDone()` can be used to assert that all nocked endpoints were called (example in [On-Duty Bot](https://github.com/ampproject/amp-github-apps/blob/master/onduty/test/e2e.test.ts#L28))

pr-deploy/test/app.test.ts
```diff
@@ -17,115 +17,145 @@
...
-    github.checks.listForRef.mockReturnValue(
-      {data: {total_count: 1, check_runs: [
-        {id: 12345, output: {text: 'Travis build number: 3'}},
-      ]}}
-    );
-
-    const requestedActionEvent:
-    Webhooks.WebhookEvent<WebhookPayloadCheckRun> = {
+    nock(apiUrl)
+      .persist()
```
rcebulko, line 124: I noticed we `persist()` here but not elsewhere. Is that because this endpoint is called more than once?

estherkim: Yep, but I changed it to use an exact number of times

pr-deploy/test/app.test.ts
estherkim: Good call, added.

pr-deploy/src/app.ts
```diff
@@ -85,18 +85,17 @@ function initializeDeployment(app: Application) {
...
-      .then(async bucketUrl => {
-        await pr.deploymentCompleted(
-          bucketUrl,
-          `${BASE_URL}amp_dist_${travisBuild}/`
-        );
-      })
-      .catch(async e => {
-        await pr.deploymentErrored(e);
-      });
+    try {
```
rcebulko, line 88: :tada: :tada: :tada:

rcebulko (approved):

pr-deploy/test/app.test.ts
estherkim: Ah makes sense. I'll come back to this if I add another test file.

pr-deploy/src/app.ts
estherkim: :)

pr-deploy/src/github.ts
```diff
@@ -14,11 +14,24 @@
...
+  typeof octokit.ChecksListForRefResponseCheckRunsItem
+>;
+type ChecksListForRefResponseCheckRunsItemOutput = GetType<
+  typeof octokit.ChecksListForRefResponseCheckRunsItemOutput
+>;
+type ChecksListForRefParams = GetType<typeof octokit.ChecksListForRefParams>;
+type ChecksUpdateParams = GetType<typeof octokit.ChecksUpdateParams>;
+type ChecksUpdateParamsActions = GetType<
+  typeof octokit.ChecksUpdateParamsActions
+>;
```
danielrozenberg, line 32: I'm confused about this - can't you refer to Octokit.SubType directly anymore? :(

pr-deploy/test/app.test.ts
```diff
@@ -17,115 +17,138 @@
...
+      .reply(200, null) // make sure no checks already exist
+      .post('/repos/test-owner/test-repo/check-runs', body => {
+        expect(body).toMatchObject({
+          'head_sha': 'abcde',
+          'name': 'test-check',
+          'status': 'queued',
+        });
+        return true;
+      })
+      .reply(200);
```
danielrozenberg, line 88: Oh no 😱 and just when I did the exact opposite (#1005, #1009)

I'm gonna argue the exact opposite! Octokit is the thing you should use for expectations, the *network calls* are the internal implementation detail. Think about it this way - when you write code, do you write `probot.networkCall('/repos/test-owner/test-repo/commits/abcde/check-runs?check_name=test-check', payload)` or do you write `probot.github.checks.create(payload)`?

I hated mocking the network calls when I first wrote the tests because that's what felt wrong about it, when I saw how you solved it I was super happy... now I'm sad 😭

danielrozenberg (commented):
I know this is already merged but I have some qualms 😣

pr-deploy/test/app.test.ts
rcebulko: I lean heavily in support of Nock for testing GitHub API calls. There are cases where probot or octokit make calls, or in some cases requests are even built manually for the rare thing octokit doesn't do. The API format is the static, external piece that we want to ensure is called, and that rarely changes and is well documented. Mocking out the various calls of the (inconsistent and often changing) octokit/probot libraries seems more fragile to me and closer to an implementation detail, especially when different apps sometimes are on different versions of each library.

For example, something fetching multiple pages of [insert resource here] might do it manually, or might use octokits auto-pagination feature. They would require entirely different test mocking, but the underlying API calls should be the same.

estherkim: I agree with Ryan, because network calls are further from Probot's internal implementation, whereas Octokit is an internal dependency. So if Probot decides to use its own GH client, our tests will not break. Intercepting http requests is what Probot docs suggest as well.

That said, if Octokit decided to use different queries, like graphql instead of REST, then our tests would break. But I'd imagine this would come with a major version upgrade.

And personally I liked mocking Octokit because it reads a lot nicer in tests... imo both ways of testing are fine!

pr-deploy/src/github.ts
estherkim: So this was all wrong and was setting `any` for all types. Found https://github.com/octokit/rest.js/releases/tag/v17.0.0#typescript which explains how to use the official types. 

Fixed in this PR - https://github.com/ampproject/amp-github-apps/pull/1014

PR #1104, opened 2020-12-09
Update linting/TS dependencies across apps

rcebulko:


checklist/package.json
```diff
@@ -32,17 +32,17 @@
     "@types/jest": "26.0.17",
     "@types/node": "13.13.34",
     "@typescript-eslint/eslint-plugin": "4.9.1",
-    "@typescript-eslint/parser": "2.34.0",
-    "eslint": "6.8.0",
-    "eslint-config-prettier": "6.15.0",
+    "@typescript-eslint/parser": "4.9.1",
+    "eslint": "7.15.0",
+    "eslint-config-prettier": "7.0.0",
```
rcebulko, line 37: Note: updating `eslint-plugin-prettier` causes the linter to error on imports that are only used for type declarations, but the `import type` syntax is not supported, and if you don't import anything it'll complain about the type missing, so there's no apparent way to make it happy. Punting on `eslint-plugin-prettier` for now.

danielrozenberg: Is there no way to configure it to ignore those specific errors?

danielrozenberg (approved):

checklist/package.json
rcebulko: It falls under the "no-unused-vars" rule, which we'd like to have in place. Hopefully there's a better solution, I plan to do more experimenting once I've cleared through the bulk of the other updates.

PR #1194, opened 2021-01-15
Make test case reporting CI-agnostic

### LINKED ISSUE #1113: Update test case reporting internals to work on CircleCI
There is a bunch of test case reporting code that relies on Travis: https://github.com/ampproject/amp-github-apps/search?q=travis+f%3Atest-case-reporting

There don't appear to be any major changes needed to `build-system/tasks/test-report-upload.js`, but it's possible that things like default SHA lengths, etc. could change on CircleCI.

This is a tracking bug and can be closed when we're sure report uploading works.

rsimha:
**PR Highlights:**

- Updates all internal implementation to be CI-agnostic
- Renames `buildNumber` and `jobNumber` to `buildId` and `jobId` respectively

**WIP Note:** I see there's a numeric `build_id` field in the `Job` interface that's likely causing this PR to fail. Happy to use a different approach if there's a better one.

Fixes #1113
Companion PR to https://github.com/ampproject/amphtml/pull/31982
Migration steps: https://github.com/ampproject/amphtml/pull/31982#issue-555363290 (will they work?)

### DIFF
--- a/test-case-reporting/README.md
+++ b/test-case-reporting/README.md
@@ -1,19 +1,19 @@
 AMP Test Case Reporting Bot
 ==============
 
-A Google Cloud app that stores the results of tests run by Travis. Used to track flaky tests.
+A Google Cloud app that stores the results of tests run during CI. Used to track flaky tests.
 
 This app runs as a [[something]].
 
 Interface
 ---------
-### API for Travis
+### API for CI
 
-The App has the following API endpoints, which are to be triggered from Travis CI
+The App has the following API endpoints, which are to be triggered during CI
 runs.
 
 * `POST /report`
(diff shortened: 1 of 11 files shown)


rcebulko (changes requested):
Since there is a database backing this, renaming all of these fields is unfortunately not as simple as changing them here. Since the build/job "numbers" are stored as strings, I'd lean toward separating the database schema changes completely, as the rest of the migration can happen independently.

Also, since CircleCI and Github Actions have different build/job data structures, the plan was going to be defining separate data types and tables for different CI platforms (rather than putting them all in one table) and then aggregating appropriately. During development we weren't sure what the end-goal CI platform would be, which is why there's a `Travis` namespace, with the expectation that there will also be CircleCI and GitHubActions namespaces with similar but distinct type definitions (ex. GA jobs may want include browser info in a column, etc.). Would you want to schedule a chat offline to help me get a grasp of where CI is/is headed, and sketch out the best way forward?

rsimha:
Agree. My gut feeling when I stopped working on this last night was that there's a better way. Closing in favor of your idea to create different data types and tables.

PR #1195, opened 2021-01-15
[test-case-reporting] Accept job/build ID fields from CI

rcebulko:
Allows CI to report `jobId` and `buildId` immediately while backend migrations are queued. Part of https://github.com/ampproject/amphtml/issues/31996


rsimha (approved):
Thanks for the quick change!

rcebulko:
@rsimha Redeploying now, should be gtg by the time you merge

PR #1209, opened 2021-01-27
[pr-deploy] Correctly get commit sha when deploying

estherkim:


pr-deploy/src/github.ts
```diff
@@ -201,14 +199,6 @@ export class PullRequest {
     return this.github.checks.update(params);
   }
 
-  async getCommitSha() {
-    const check = await this.getCheck_();
-    if (!check.output || !check.output.text) {
-      return '';
-    }
-    return check.output.text.replace(/Commit SHA: /g, '');
-  }
```
rsimha, line 210: Nice to see this go away!

pr-deploy/src/github.ts
```diff
@@ -145,7 +144,6 @@ export class PullRequest {
           'Please click the `Create a test site` button above to ' +
           'deploy the minified build of this PR along with examples from ' +
           '`examples/` and `test/manual/`. It should only take a minute.',
-        text: `Commit SHA: ${sha}`,
```
rsimha, line 148: For usability / debuggability, would it makes sense to retain this `text` field in `buildCompleted()`, but not actually scrape it `initializeDeployment()`?

pr-deploy/src/github.ts
```diff
@@ -32,12 +32,12 @@ const owner = process.env.GH_OWNER;
...
 
 export class PullRequest {
+  public headSha: string;
   private github: Octokit;
-  private headSha: string;
 
   constructor(github: Octokit, headSha: string) {
-    this.github = github;
     this.headSha = headSha;
+    this.github = github;
```
rsimha, line 40: Curious: does this order matter?

rsimha (approved):
Thanks for the quick fix! LGTM with a couple non-blocking comments.

pr-deploy/src/github.ts
estherkim: the commit sha is displayed as part of the Checks UI on GH

pr-deploy/src/github.ts
estherkim: It was just to make it consistent with above, where I switched the order since public should come before private properties.

PR #1229, opened 2021-02-24
[bundle-size] don't auto-tag @ampproject/wg-infra members on errored checks

### LINKED ISSUE #857: Request: Disable reviewer assignment on PRs with error'd bundle size checks
When the bundle-size check fails, members of the approvers teams can get assigned to the PR, where their only job is to restart the Travis shard. This appears (anecdotally/in my experience) to be a far more frequent occurrence than actual PRs that genuinely needed approval because of a bundle-size change.

It may be reasonable to move this responsibility to the PR author themselves. Rather than automatically assigning a bundle-size approver, we could start the way owners bot did: recommending reviewers in the bot output, and making it the author's responsibility to respond appropriately to the failing checks. When one is needed, the author can request a review explicitly. When the check is just failing because of a race condition with the bundle-size check, the author can handle restarting the shard.

Until the bot can automatically detect error'd checks vs. PRs that actually changed the bundle size, this would cut down on notifications for unrelated PRs that just need a Travis re-run.

/tryassign @danielrozenberg

danielrozenberg:
Fixes #857

### DIFF
--- a/bundle-size/api.js
+++ b/bundle-size/api.js
@@ -88,9 +88,10 @@ function erroredCheckOutput(partialBaseSha) {
       '`https://github.com/ampproject/amphtml-build-artifacts` ' +
       'repository. This can happen due to failed or delayed CI builds on ' +
       'said `master` commit.\n' +
-      `A member of [${superUserTeams}] will be added automatically to review ` +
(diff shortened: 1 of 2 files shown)


rcebulko (approved):
🙏🏻

PR #1235, opened 2021-03-03
[bundle-size] Surface errors encountered while reporting a bundle-size change as HTTP 500

danielrozenberg:


rsimha (approved):

rcebulko (approved):

bundle-size/api.js
```diff
@@ -485,7 +485,21 @@ exports.installApiRouter = (app, router, db, githubUtils) => {
...
+      const superUserTeams =
+        '@' + process.env.SUPER_USER_TEAMS.replace(/,/g, ', @');
+      response
+        .status(500)
+        .end(
+          `The bundle-size bot encountered a server-side error: ${error}\n` +
+            `Contact ${superUserTeams} for help.`
+        );
+      return;
+    }
```
rsimha, line 501: Should this be made into a helper (with retries) and be called on line 512 as well so that any error during a retry is also surfaced?

danielrozenberg: I wrapped the whole thing in the same try/catch clause, so no need for a helper here

PR #1236, opened 2021-03-04
[meta] Setup GitHub Actions

danielrozenberg:


rsimha (approved):

.github/workflows/build.yaml
```diff
@@ -16,6 +16,12 @@ jobs:
...
         uses: actions/setup-node@v2
+      - name: Get npm cache directory
+        uses: actions/cache@v2
+        with:
+          path: ~/.npm
+          key: node-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: node-
       - name: Install dependencies
         run: npm ci
       - name: Build and test
```
rsimha, line 27: Non-blocking suggestion for a later PR: Today, we run `pr-check.js` as one script, which has the benefit of all the build-target logic.

It might be worth splitting the different app tests across multiple steps in one job, or even matrixify them if you think the additional VMs and statuses are worth it. I'd imagine it's possible to do so without losing the efficiencies of the build-target logic.

danielrozenberg: It certainly makes sense at this point. For _most_ PRs though it usually runs just one of the tests, so I'm not sure if it's worth putting in the dev time. Maybe a fixit thing :)

PR #1238, opened 2021-03-04
[meta] Correctly filter changed files for PR checks

danielrozenberg:
GitHub Actions has a different model for how it uses Git compared to Travis and it breaks the "list of changed files" logic. This PR replaces it with a preexisting GitHub Actions that does the heavy-lifting for us


.github/workflows/build.yaml
```diff
@@ -24,7 +24,11 @@ jobs:
           restore-keys: node-
       - name: Install dependencies
         run: npm ci
+      - name: Get changed files
+        uses: lots0logs/gh-action-get-changed-files@2.1.4
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
```
rsimha, line 30: I don't think it's a good idea to use a third party tool like this when we can do things natively.

I think if you use a `fetch-depth` of 0 with the `checkout` action (which is published by GH and not a third party), you'll see the correct set of changes.

Here's how it works for `amphtml`: https://github.com/ampproject/amphtml/blob/59a7b25f4c19b1fe70ac126534a18051fd92d12e/.github/workflows/cross-browser-tests.yml#L18-L21

rsimha (changes requested):
> GitHub Actions has a different model for how it uses Git compared to Travis and it breaks the "list of changed files" logic.

**High level comment:** On `amphtml`, we use the exact same code to determine changed files on Travis, CircleCI, and GH Actions. Specifically, see lines 1-50 in this log file: https://github.com/ampproject/amphtml/runs/2035327441?check_suite_focus=true#step:4:8

It's possible that the logic in this repo has bit-rotted. Instead of rewriting things, can we migrate the code in https://github.com/ampproject/amphtml/blob/master/build-system/pr-check/ci-job.js (and a minimal set of the code it imports) to this repo?

.github/workflows/build.yaml
danielrozenberg: While I don't mind copy-pasting from amphtml, given that this repo only needs this logic to work for one CI system I figured that doing code reuse from another open source project is the better approach.

I also don't see the problem with using a 3rd party tool - this repo doesn't have any secret variables ([`GITHUB_TOKEN`](https://docs.github.com/en/actions/reference/authentication-in-a-workflow#about-the-github_token-secret) is auto-generated and read-only) and everything is world-readable anyways :)

I'm leaning towards reusing a 3rd party Action instead of reinventing the wheel, even if that wheel already exists in our other repo. wdyt?

rsimha: I meant to say that you shouldn't need the third party action if you set the depth.

rsimha: As discussed offline, there are advantages to not fetching the entire commit history, so let's try this approach. (If it works, `ampproject/amphtml` can also benefit from not having to fetch with a depth of 0.

rsimha (approved):
LGTM!

PR #1241, opened 2021-03-10
bundle-size: Use mergeSha when showing code changes

jridgewell:
Uses the [new `mergeSha` field](https://github.com/ampproject/amphtml/pull/33165/files#diff-c141453bcbdbe03ca8414e5b4d425c5b7323e43d524b614455ce13770633509eR191), which resolves https://github.com/ampproject/amphtml/pull/33165#issuecomment-794642367.

### DIFF
--- a/bundle-size/api.js
+++ b/bundle-size/api.js
@@ -116,14 +116,14 @@ function extraBundleSizesSummary(
   bundleSizeDeltasAutoApproved,
   missingBundleSizes
 ) {
-  const compareUrl =
-    mergeSha ||
-    `https://github.com/ampproject/amphtml/compare/${baseSha}..${headSha}`;
   let output =
     '## Commit details\n' +
     `**Head commit:** ${headSha}\n` +
     `**Base commit:** ${baseSha}\n` +
-    `**Code changes:** ${compareUrl}\n`;
+    `**Code changes:** ${
+      mergeSha ||
+      `https://github.com/ampproject/amphtml/compare/${baseSha}..${headSha}`
+    }\n`;
 
   if (bundleSizeDeltasRequireApproval.length) {
     output +=
--- a/bundle-size/api.js
+++ b/bundle-size/api.js
@@ -469,6 +469,7 @@ exports.installApiRouter = (app, router, db, githubUtils) => {
   router.post('/commit/:headSha/report', async (request, response) => {
     const {headSha} = request.params;
     // mergeSha is new, and not all reports will have it.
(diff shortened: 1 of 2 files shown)


bundle-size/api.js
```diff
@@ -110,15 +111,19 @@ function erroredCheckOutput(partialBaseSha) {
...
   headSha,
   baseSha,
+  mergeSha,
   bundleSizeDeltasRequireApproval,
   bundleSizeDeltasAutoApproved,
   missingBundleSizes
 ) {
+  const compareUrl =
+    mergeSha ||
+    `https://github.com/ampproject/amphtml/compare/${baseSha}..${headSha}`;
```
rsimha, line 121: If we're insisting on the presence of `mergeSha` in the report further down, why not just use `mergeSha`? (especially since the constructed URL is wrong more often than not...)

bundle-size/api.js
```diff
@@ -110,15 +111,19 @@ function erroredCheckOutput(partialBaseSha) {
...
   let output =
     '## Commit details\n' +
     `**Head commit:** ${headSha}\n` +
     `**Base commit:** ${baseSha}\n` +
-    `**Code changes:** https://github.com/ampproject/amphtml/compare/${baseSha}..${headSha}\n`;
+    `**Code changes:** ${compareUrl}\n`;
```
rsimha, line 126: Is there a way to explicitly make this a URL with `mergeSha`? (Or does GH linkify it under all circumstances?)

rsimha (commented):
LGTM with two questions. Let's have Daniel review as well.

bundle-size/api.js
jridgewell: Sure.

bundle-size/api.js
jridgewell: Github links it just like it does for `headSha` and `baseSha`.

bundle-size/api.js
```diff
@@ -454,13 +468,21 @@ exports.installApiRouter = (app, router, db, githubUtils) => {
 
   router.post('/commit/:headSha/report', async (request, response) => {
     const {headSha} = request.params;
-    const {baseSha, bundleSizes} = request.body;
+    // mergeSha is new, and not all reports will have it.
```
danielrozenberg, line 471: Add a TODO (you can assign this to me) to make this non-optional after a month or so

danielrozenberg (approved):

rsimha (approved):

PR #1242, opened 2021-03-19
[test-status] Change checkRunId's data type from integer to bigint

danielrozenberg:
* I already changed the database type directly to fix the integer overflow error from the db engine
* knex [returns `bigint`s as strings](http://knexjs.org/#Schema-bigInteger) because it might overflow the safe JavaScript max number value, but we can probably safely assume it won't go over [9,007,199,254,740,991](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER), so explicitly cast all of them to Numbers after the db returns

### DIFF
--- a/test-status/db.js
+++ b/test-status/db.js
@@ -45,7 +45,7 @@ exports.getCheckRunId = async (db, headSha, type, subType) => {
   if (existingCheck === undefined) {
     return null;
   } else {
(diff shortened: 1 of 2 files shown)


rsimha (approved):
👍

PR #1244, opened 2021-03-19
[bundle-size] Change checkRunId's data type from integer to bigint

danielrozenberg:
Same as #1242 but for bundle-size

> * I already changed the database type directly to fix the integer overflow error from the db engine
> * knex [returns `bigint`s as strings](http://knexjs.org/#Schema-bigInteger) because it might overflow the safe JavaScript max number value, but we can probably safely assume it won't go over [9,007,199,254,740,991](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER), so explicitly cast all of them to Numbers after the db returns

### DIFF
--- a/bundle-size/common.js
+++ b/bundle-size/common.js
@@ -20,11 +20,11 @@
  *
  * @param {!Knex} db database connection.
  * @param {string} headSha commit SHA of the head commit of a pull request.
- * @return {!object} GitHub Check object.
+ * @return {?object} GitHub Check object.
  */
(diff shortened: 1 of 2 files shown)


danielrozenberg:
If we ever get there (heavens I hope we won't), we'd have to change the db data type anyways, so for now gonna continue treating it as-is

rcebulko (approved):

PR #1258, opened 2021-03-31
Update test-case-reporting dashboard landing page

rileyajones:
Previously the test-case-reporting dashboards landing page was also the build list page. Because the site has no navigation elements to only ways for a user to discover a subpage was by word of mouth or reading the source code. Some of the stats pages in particular are quite valuable and I would like to expand both their visibility and eventually functionality.
### Old
image
### New
image

I plan to add another stats overview page with graphs etc along with a nav bar in a future PR.

@rsimha @rcebulko @danielrozenberg @estherkim

### DIFF
--- a/test-case-reporting/app.ts
+++ b/test-case-reporting/app.ts
@@ -98,45 +98,24 @@ app.get('/', async (req, res) => {
       // Leaving this in the base path in case there is some automation using this.
       res.json({builds});
     } else {
-      const testCases = await record.getRecentTestCaseStats(
-        10,
-        pageInfo
-      );
-      const testCasesWithStats = testCases.map((testCase: any) => {
-        const total =
-          testCase.stats.pass +
-          testCase.stats.fail +
-          testCase.stats.skip +
-          testCase.stats.error;
-        const statPercent = (stat: number): string =>
-          ((100 * stat) / total).toFixed(0);
-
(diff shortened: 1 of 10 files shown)


rcebulko:
Thanks for doing this! It actually is possible to navigate by clicking any build, then clicking any of the pass/fail/error/skip % boxes to go to that stat's page. Which is extremely non-obvious and convoluted. This is great!

test-case-reporting/app.ts
```diff
@@ -87,7 +87,44 @@ function lowerCaseStatus({status, ...rest}: TestRun): any {
...
+app.get('/', async (req, res) => {
+  const {json} = req.query;
+
+  try {
+    const pageInfo: PageInfo = {limit: 10, offset: 0};
+    const builds = await record.getRecentBuilds(pageInfo);
+
+    if (json) {
+      // Leaving this in the base path in case there is some automation using this.
+      res.json({builds});
```
rcebulko, line 99: Context: This project was built to mainly be a backend/data-collection. In order for Ricardo to have something to show, the front-end was thrown together, with all endpoints supporting JSON, with the expectation (hope?) that down the line it would get a proper frontend, such as the release dashboard, which would poll these API endpoints for JSON directly

rcebulko (approved):

estherkim (approved):

PR #1265, opened 2021-04-13
[pr-deploy] Deploy using build file from CircleCI artifacts instead of a gcloud storage bucket

danielrozenberg:
Should be merged and deployed in conjunction with https://github.com/ampproject/amphtml/pull/33774

Tested on a version of the bot deployed to my fork, see https://github.com/danielrozenberg/amphtml/pull/4/checks?check_run_id=2327965141

Highlights:
* Again we depend on the CI system we use, because CircleCI does not provide direct URLs to the artifacts, but rather we need to get them from their JSON API
* Replaced zip with tar.gz - zip is not supported for streaming and this made it difficult keep it around. .tar.gz files are 1% smaller too 😅
* Fixed package.lock run configs so that `npm run build` will actually use the tsconfig.json file, and `npm run dev` will listen to changes in .ts files, not .js files, and rebuild the js on every change
* Lots of annoying auto-formatter changes


pr-deploy/src/app.ts
```diff
@@ -70,7 +76,8 @@ function initializeRouter(app: Application) {
     response.send({status: 200});
   };
 
-  router.post('/headshas/:headSha/:result', initialize);
+  router.post('/headshas/:headSha/:result(success)/:externalId', initialize);
+  router.post('/headshas/:headSha/:result(errored|skipped)', initialize);
```
estherkim, line 80: Why add this line? No need to initialize when there are no build files?

pr-deploy/src/zipper.ts
```diff
@@ -14,42 +14,68 @@
...
+    externalId
+  );
+  const buildArtifactZipUrl = await fetch(buildArtifactJsonUrl)
+    .then(res => res.json())
+    .then((json: Array<{[key: string]: string}>) =>
+      json.find(item => item['path'].endsWith('/amp_nomodule_build.tar.gz'))
+    )
+    .then(item => item['url']);
+
+  return await new Promise(async (resolve, reject) => {
```
estherkim, line 47: Thoughts on refactoring `new Promise` to an async function? 😬

pr-deploy/test/app.test.ts
```diff
@@ -14,14 +14,12 @@
  * limitations under the License.
  */
 
-// mock unzipAndMove dependency before importing '../src/app'
+// mock decompressAndMove dependency before importing '../src/app'
 jest.mock('../src/zipper', () => {
   return {
-    unzipAndMove: jest
+    decompressAndMove: jest
```
estherkim, line 20: Is this the only non-formatting change in this file? Hard to tell, help

pr-deploy/src/app.ts
```diff
@@ -70,7 +76,8 @@ function initializeRouter(app: Application) {
     response.send({status: 200});
   };
 
-  router.post('/headshas/:headSha/:result', initialize);
+  router.post('/headshas/:headSha/:result(success)/:externalId', initialize);
```
rsimha, line 79: From reading https://github.com/ampproject/amphtml/pull/33774/files#diff-f456adc9b34fa8977425b08583a46b2106b4208392eaa06e7b72b6995137ea16R93-R94, am I correct that `externalId` here corresponds to the CircleCI `jobId` (a.k.a. `CIRCLE_BUILD_NUM`)? If so, would it be clearer to call this section `jobId` instead of `externalId`?

pr-deploy/src/github.ts
```diff
@@ -21,11 +21,13 @@ type ChecksGetResponseData = Types['checks']['get']['response']['data'];
...
 
-const ACTIONS: ChecksUpdateParams['actions'] = [{
-  label: 'Create a test site',
-  description: 'Serves the minified output of this PR.',
-  identifier: 'deploy-me-action',
-}];
+const ACTIONS: ChecksUpdateParams['actions'] = [
+  {
+    label: 'Create a test site',
+    description: 'Serves the minified output of this PR.',
```
rsimha, line 27: Since we now have multiple types of minified output (nomodule, module, experiment), maybe specify that this is nomodule?
```suggestion
    description: 'Serves the nomodule minified output of this PR.',
```

pr-deploy/src/github.ts
```diff
@@ -21,11 +21,13 @@ type ChecksGetResponseData = Types['checks']['get']['response']['data'];
 type ChecksListForRefParams = Types['checks']['listForRef']['parameters'];
 type ChecksUpdateParams = Types['checks']['update']['parameters'];
 
-const ACTIONS: ChecksUpdateParams['actions'] = [{
-  label: 'Create a test site',
-  description: 'Serves the minified output of this PR.',
-  identifier: 'deploy-me-action',
-}];
+const ACTIONS: ChecksUpdateParams['actions'] = [
```
rsimha, line 24: While you're editing this file, it might be worth refining the GitHub status check UI. Here are screenshots of what they look like today, with some suggestions. (Feel free to punt / ignore as you see fit.)

-----------------

**Before deploy:**

Screen Shot 2021-04-13 at 4 20 17 PM

- Is the list of directories still correct?
- Is the time estimate of a minute still accurate? (Or are things faster with CircleCI?)

-----------------

**After deploy:**

Screen Shot 2021-04-13 at 4 23 08 PM

- The paragraph looks kinda crowded and hard to read. Maybe make it a bulleted list of points instead?
- Both the "website" link and the "amp article example" link point to the same page. Is there a better alternative for the "website" link?
- The "view more details" link appears to be broken (see below). Maybe this won't be a problem once the switch is complete?

Screen Shot 2021-04-13 at 4 23 35 PM

rsimha (commented):
This PR LGTM overall (with a few comments below). I'll let Esther provide the approving review.

pr-deploy/src/app.ts
danielrozenberg: `initialize` is the method that handles all three cases, I'm just splitting between the case that now requires the `externalId` and the two other cases that don't

pr-deploy/src/zipper.ts
danielrozenberg: Between the streams and pipes and all those, I thought it's better to keep more-or-less to the same structure for now...

pr-deploy/test/app.test.ts
danielrozenberg: Yep, only non-formatting change

pr-deploy/src/app.ts
danielrozenberg: `externalId` is how this is called in the GitHub API - I don't mind calling it either way, lmk

rsimha: Oh, if it's a well-known part of an API, okay to leave it as-is. (Maybe add a link to `pr-deploy/redacted.env` for documentation?)

pr-deploy/src/github.ts
danielrozenberg: > * Is the list of directories still correct?

Updated

> * Is the time estimate of a minute still accurate? (Or are things faster with CircleCI?)

I believe on CircleCI it would be faster, but we can find out later

> * The paragraph looks kinda crowded and hard to read. Maybe make it a bulleted list of points instead?

Added line breaks in between

> * Both the "website" link and the "amp article example" link point to the same page. Is there a better alternative for the "website" link?

@estherkim?

> * The "view more details" link appears to be broken (see below). Maybe this won't be a problem once the switch is complete?

Yes, this is a result of the manual testing

danielrozenberg:
@estherkim @rsimha bump

estherkim (approved):

danielrozenberg:
Deployment plan:
* Start a new version of this bot on GAE, but give it 0% traffic (keep 100% to the previous one)
* Simultaneously change the traffic from 0% to 100% and change the gating flag on the main repo's CircleCI builds
* Verify everything works with a test PR, if not, undo previous step

PR #1266, opened 2021-04-15
Sort bundle size output by delta instead of binary

rcebulko:
Was going to make a feature request, but figured it would be easier to just offer the change as a PR. Sorts bundle-size output by delta, so it's easy to scan the list for the files with the largest/smallest relative changes.

### DIFF
--- a/bundle-size/test/api.test.js
+++ b/bundle-size/test/api.test.js
@@ -257,7 +257,7 @@ describe('bundle-size api', () => {
       });
 
       baseBundleSizeFixture.content = Buffer.from(
-        '{"dist/v0.js":12.34,"dist/v0/amp-accordion-0.1.js":1.11,"dist/v0/amp-ad-0.1.js":4.53,"dist/v0/amp-anim-0.1.js":5.65}'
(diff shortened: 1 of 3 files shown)


bundle-size/test/api.test.js
```diff
@@ -302,6 +302,68 @@ describe('bundle-size api', () => {
...
+
+      baseBundleSizeFixture.content = Buffer.from(
+        `{
+          "dist/v0.js": 12.34,
+          "dist/v0/amp-accordion-0.1.js":1.11,
+          "dist/v0/amp-ad-0.1.js": 4.53,
+          "dist/v0/amp-anim-0.1.js": 5.65,
+          "dist/v0/amp-date-display-0.1.js": 8.99,
+          "dist/v0/amp-truncate-text-0.1.js": 2.12
+        }`
```
danielrozenberg, line 337: If you're already gonna format it, just wrap an explicit object with `JSON.stringify({...})`

danielrozenberg (approved):

bundle-size/test/api.test.js
rcebulko: Done

PR #1267, opened 2021-04-15
Delete checklist/ 

estherkim:
Because it's not being used or worked on.


danielrozenberg:
@alanorozco intended on creating this but never got around to it. ok to delete? (of course this can always be recovered later)

alanorozco:
Totally fine to delete. Thanks!

- A Github Workflow or a collection thereof would be more appropriate than a bot at this point.
- We're likely reworking PR templates in a way that would make this less useful.

estherkim:
Both sound great. If we can get access to GitHub forms for issues, that might help a lot too!

That said, can I get a ✅ ✅ ✅

alanorozco (approved):

danielrozenberg (approved):
Proxying approval

PR #1286, opened 2021-05-03
[owners] Fix double-call to path helper

rcebulko:
A recent change to the owners cache introduced an accidental double-call to a helper that prepends the repo root to a filepath, resulting in the cache system trying to delete a nonexistent file at `amphtml/amphtml/...` and crashing each time the bot tried to refresh. This PR fixes that, and wraps the cache so errors are logged but do not blow up tree parsing/refreshing.

image


rsimha (approved):
Appreciate the quick fix!

PR #1284, opened 2021-05-03
[owners] Support `reviewerPool` of users and/or teams in place of `reviewerTeam` property

rcebulko:
Closes https://github.com/ampproject/amp-github-apps/issues/1283
Requires an update to amphtml's root OWNERS file, but this should not fail any PRs on that repo in the interim unless they trigger OWNERS build target. Will require pushing an updated deploy tag to trigger a build.

Before: `reviewerTeam` allowed only in repo root `OWNERS` file, must be a string representing a team
Now: `reviewerPool` allowed only in repo root `OWNERShi file, must be an array of strings representing teams or usernames

PR #1288, opened 2021-05-04
Update main branch name used during CI

rsimha:
During build target computation, we look at the set of files changed by a PR when compared with the common ancestor commit with the `main` branch. On Travis, it was sufficient to use `main`, but on GH actions, we need to use `origin/main`.

For evidence, see this screenshot from an `amphtml` GH Actions [build](https://github.com/ampproject/amphtml/runs/2503804538?check_suite_focus=true#step:4:23):

image

Addresses https://github.com/ampproject/amp-github-apps/pull/1284#issuecomment-831404707

### DIFF
--- a/build-system/git.js
+++ b/build-system/git.js
@@ -20,13 +20,16 @@
  */
 
 const {getStdout} = require('./exec');
+const {isCiBuild} = require('./ci');
 
 /**
- * Returns the merge base of the PR branch and the main branch.
+ * Returns the merge base of the PR branch and the main branch. During CI on
+ * GH actions, the main branch identifier is "origin/main".
  * @return {string}
  */
 function gitMergeBaseMain() {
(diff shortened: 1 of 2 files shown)


build-system/git.js
```diff
@@ -26,7 +26,7 @@ const {getStdout} = require('./exec');
  * @return {string}
  */
 function gitMergeBaseMain() {
-  return getStdout(`git merge-base main HEAD`).trim();
+  return getStdout(`git merge-base origin/main HEAD`).trim();
```
rcebulko, line 29: Let's add an explanatory comment here reflective of the PR description

rsimha: Went one step further and added logic to use `main` during local development and `origin/main` during CI. Also added an explanatory comment as suggested.

rcebulko (approved):

PR #1317, opened 2021-05-26
Add a new GitHub app that publishes any webhook it receives to a Cloud Pub/Sub topic

danielrozenberg:
The intention is to install this app on the amphtml repository and listen to `push` webhook events, then a google-internal app will subscribe to the Pub/Sub topic and synchronize any changes to the [client-side experiment config](https://github.com/ampproject/amphtml/issues/34013) file with the CDN - this way the google-internal part is kept to a minimum, since any CDN can listen to the same Pub/Sub topic and perform a similar sync

### DIFF
--- a/build-system/build-targets.js
+++ b/build-system/build-targets.js
@@ -28,13 +28,13 @@ const APPS_TO_TEST = [
   'bundle-size',
   'error-monitoring',
   'invite',
-  'webhook-pubsub-publisher',
   'onduty',
   'owners',
   'pr-deploy',
   'release-calendar',
   'test-case-reporting',
   'test-status',
+  'webhook-pubsub-publisher',
 ];
 
 /**
--- a/build-system/build-targets.js
+++ b/build-system/build-targets.js
@@ -28,6 +28,7 @@ const APPS_TO_TEST = [
   'bundle-size',
   'error-monitoring',
   'invite',
+  'webhook-pubsub-publisher',
   'onduty',
   'owners',
   'pr-deploy',
--- a/webhook-pubsub-publisher/README.md
+++ b/webhook-pubsub-publisher/README.md
@@ -1,0 +1,52 @@
+# Webhook Pub/Sub Publisher
+
+Publishes received [webhooks from GitHub](https://docs.github.com/en/developers/webhooks-and-events/webhooks) to a [Google Cloud Pub/Sub](https://cloud.google.com/pubsub) topic.
+
+## How the app works
+
+This app is a straight-forward [Probot](https://probot.github.io/) application that does exactly what its description says and nothing more.
+
+[insert patrick-push-it-somewhere-else.gif meme](https://knowyourmeme.com/memes/push-it-somewhere-else-patrick)
+
+## Local development
+
+### GitHub Application setup
+
+1. Start a new [Smee channel](https://smee.io/). This can be used to proxy
+   GitHub webhooks to your local machine.
+2. Create a new [GitHub App](https://github.com/settings/apps/new) with the following settings:
+   - General
+     - Set _Homepage URL_ to the Smee channel
+     - Set _Webhook URL_ to the Smee channel
+     - Set _Webhook Secret_ to any pin your choice
+   - Permissions
+     - None
+   - Events
+     - Subscribe to whichever choice events you want to publish
+3. After creating the application, generate and download a private key. Also
+   take note of the App ID.
+4. Install the application on a GitHub repository that you want to use for
+   testing
+
+### Google Cloud setup
+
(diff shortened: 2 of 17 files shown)


estherkim:
Oh dang I wish I said this earlier but I was gonna propose moving all our bots to GitHub Actions, that way we don't need GCP serving infra and less dependencies on internal things (yay).

A GitHub Action could be triggered by push events too - https://docs.github.com/en/actions/reference/events-that-trigger-workflows#push - then you can run a node script to publish!

danielrozenberg:
I thought about something like that, but in this specific case it's easier to use this shim - it's possible to create a push-subscription with GCP that can forward the messages to a Google-internal RPC server, which is probably the easiest way to do something like this

danielrozenberg:
> Oh dang I wish I said this earlier but I was gonna propose moving all our bots to GitHub Actions, that way we don't need GCP serving infra and less dependencies on internal things (yay).

This is a good goal though! It makes sense for many of the apps to run as GitHub Actions. This _specific_ case does require Google dependencies as long as Google backs the CDN :wink:

danielrozenberg:
Now that you've reviewed the CDN-side AMP_EXP syncer, I would appreciate a review of this app :)

rileyajones (approved):

webhook-pubsub-publisher/README.md
```diff
@@ -0,0 +1,52 @@
+# Webhook Pub/Sub Publisher
```
estherkim, line 1: I think @rileyajones would appreciate it if you added deployment steps

webhook-pubsub-publisher/tsconfig.json
```diff
@@ -0,0 +1,25 @@
...
+    "moduleResolution": "node",
+    "skipLibCheck": true,
+    "sourceMap": true,
+    "outDir": "dist",
+    "baseUrl": ".",
+    "paths": {
+      "*": [
+        "node_modules/*",
+        "node_modules/@types/*",
+        "src/types/*"
```
estherkim, line 16: there is no `src/types` dir is there?

webhook-pubsub-publisher/src/app.ts
```diff
@@ -0,0 +1,46 @@
...
+import {PubSub} from '@google-cloud/pubsub';
+import type {Context, Probot} from 'probot';
+
+export default (app: Probot): void => {
+  if (process.env.NODE_ENV !== 'test') {
+    dotenv.config();
+  }
+
+  app.onAny(async (context: Context): Promise<void> => {
+    context.log(`Received "${context.name}" event with ID ${context.id}`);
```
estherkim, line 27: Which webhooks do we care about? Is it worth filtering here

build-system/build-targets.js
```diff
@@ -28,6 +28,7 @@ const APPS_TO_TEST = [
   'bundle-size',
   'error-monitoring',
   'invite',
+  'webhook-pubsub-publisher',
```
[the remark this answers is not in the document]
danielrozenberg: Done

webhook-pubsub-publisher/redacted.env
```diff
@@ -0,0 +1,15 @@
+# From the GitHub App settings page.
+APP_ID=120697
+WEBHOOK_SECRET=[REDACTED]
+GITHUB_CLIENT_ID=Iv1.16971967a499955d
+GITHUB_CLIENT_SECRET=[REDACTED]
+
+# Base64 encoded private key from GitHub.
+PRIVATE_KEY=[REDACTED]
```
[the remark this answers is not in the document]
danielrozenberg: These are used directly by Probot... also that's what these env variables are called in every app so far, including the ones your wrote 😉

webhook-pubsub-publisher/tsconfig.json
danielrozenberg: Done.

webhook-pubsub-publisher/src/app.ts
danielrozenberg: Doesn't matter! I'm keeping this open so that it can forward everything to the pub/sub topic - the subscribers can do the filtering directly via the subscription or in the application itself. Keeping this open to everything would let us pivot quickly if we ever decide to reuse the same pub/sub topic with other subscribers!

webhook-pubsub-publisher/README.md
danielrozenberg: The deployment will use the same method in DEPLOYMENT.md, so I'm just adding a short description like in the readme for pr-deploy or bundle-size. Done.

PR #1368, opened 2021-06-04
Disable auto-rebasing of update PRs

rsimha:
This PR temporarily sets the renovate setting `rebaseWhen` to `never` so there's less noise while the [initial backlog](https://github.com/ampproject/amp-github-apps/pulls?q=is%3Apr+is%3Aopen+author%3Aapp%2Frenovate+sort%3Aupdated-desc) of updates is cleared. If an upgrade PR needs accompanying code changes, the PR branch will have to be manually rebased after the code is fixed.

Once we're caught up, it may be possible to restore `rebsaeWhen` to the default of `conflicted` and still avoid PR noise. Can revisit this bit later.

**Reference:** https://docs.renovatebot.com/configuration-options/#rebasewhen

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -16,6 +16,7 @@
   "packageRules": [{
     "matchFiles": ["package.json"],
     "groupName": "core packages",
+    "rebaseWhen": "never",
     "automerge": true,
     "assignAutomerge": true
   }, {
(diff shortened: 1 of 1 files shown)


estherkim (approved):
thank you!

PR #1369, opened 2021-06-04
Consolidate and standardize all linting at the root level

rsimha:
**Background:**

Until recently, `amp-github-apps` had a [huge backlog](https://gist.github.com/rsimha/a4d1fff10832e7275f0bc2b6b40584c4) of package updates (>100 PRs). Once the grouping structure was optimized and auto-review + auto-merge was enabled, the backlog became [considerably smaller](https://github.com/ampproject/amp-github-apps/issues/1338#issue-910538348) (~15 PRs).

**Problem:**

Since each app maintains its own set of versions for `eslint`, `prettier`, etc., new linting package releases will trigger multiple PRs for the same upgrade. This can be solved by consolidating all linting at the root level.

**PR Highlights:**
- Move all lint dependencies to the root `package.json`
- Create a single linting npm script while continuing to use app-specific rules and settings
- Update the main PR check script to run lint checks before app-specific tests
- Remove app-specific linting dependencies and commands
- Remove lint rules that conflict with prettier (e.g. [here](https://user-images.githubusercontent.com/26553114/120856094-4f636600-c54d-11eb-8032-1984d9e5a20e.png))
- Apply auto-fixes by running `npm run fix` from the root dir

For easy reviewing, look at one commit at a time.

**Future work:**

If this PR yields good results, consolidate versions for other common infrastructure (e.g. testing libraries like `jest`)

Closes #1357


package.json
```diff
@@ -10,27 +10,28 @@
...
+    "lint": "eslint . --ext .js,.jsx,.ts,.tsx",
+    "fix": "npm run lint -- --fix",
+    "test": "npm run lint",
     "replace-secrets": "node bin/replace-secrets.js",
     "unredact-env": "read -p \"This will overwrite existing .env file. Press enter to continue.\n\" && node bin/unredact-env.js"
   },
   "devDependencies": {
-    "@typescript-eslint/eslint-plugin": "2.34.0",
-    "@typescript-eslint/parser": "2.34.0",
+    "@typescript-eslint/eslint-plugin": "4.26.0",
```
rileyajones, line 20: Glad to see you got this to work. I was seeing some strange issues related to a similar version change in [this PR](https://github.com/ampproject/amp-github-apps/pull/1359). Out of curiosity was this related to the `eslint` version?

rileyajones (approved):

package.json
rsimha: Yes, I'm fairly certain this was due to the variety of plugin versions across different apps, and the fact that they might've been incompatible with the variety of `eslint` / `typescript` / `prettier` versions.

rsimha:
**Final updates:**
- Updated lint check + all app-level tests are [green](https://github.com/ampproject/amp-github-apps/pull/1369/checks?check_run_id=2749267404#step:4:80).
- Noise from in-flight and future upgrade PRs should further reduce.
- Can make any other code changes via a follow up.
- Apps can be re-deployed when other changes make it necessary.

Merging now.

bundle-size-chart/package.json
```diff
@@ -11,15 +11,12 @@
...
   "scripts": {
-    "lint": "eslint src --ext .js,.ts",
-    "fix": "npm run -- --fix",
     "dev": "nodemon",
     "deploy": "gcloud --project amp-bundle-size-chart app deploy",
     "deploy-cron": "gcloud --project amp-bundle-size-chart app deploy cron.yaml",
     "deploy-tag": "git tag 'deploy-bundle-size-chart-'`date -u '+%Y%m%d%H%M%S'`",
     "prestart": "tsc",
     "start": "node dist/app.js",
-    "pretest": "npm run lint",
```
rcebulko, line 22: @rsimha IIRC All of the apps relied on `pretest` so that PRs touching that app would trigger those build targets, and each test target would perform a lint check as part of the testing pass (`npm pretest` is run automatically when `npm test` is run). I believe the result of this PR is that linting is now disabled in CI for this repo. Am I missing something?

rsimha: In addition to removing per-app linting, this PR added a new root-level script that runs lint across the entire repo:

https://github.com/ampproject/amp-github-apps/blob/dd5d251e2b31df7bb6264c669cb3f6104862c6d0/package.json#L13-L15

During CI, the check is run no matter which apps are touched by the incoming PR.

https://github.com/ampproject/amp-github-apps/blob/dd5d251e2b31df7bb6264c669cb3f6104862c6d0/build-system/pr-check.js#L37-L44
https://github.com/ampproject/amp-github-apps/blob/dd5d251e2b31df7bb6264c669cb3f6104862c6d0/build-system/pr-check.js#L57-L58

For example, see the root-level testing step in [this build](https://github.com/ampproject/amp-github-apps/runs/2758457826?check_suite_focus=true#step:4:8).

Linting the entire repo like this takes <20s, and saves significant noise / duplication when it comes to package management / upgrades.

Good question to have asked though. 👍

rcebulko: It seems confusing that the root-level `npm run test` task exclusively lints and runs no tests. Perhaps `npm run test` could trigger tests for all apps, while the `pretest target` could lint and the CI config would call `npm run lint` directly since it's manually executing each test target. Currently, the CI collapses the output so all that's visible is `npm run test` so the presence of linting is hidden

rsimha: Fair point. We could change the root-level execution to `npm run lint` instead of hiding it behind `npm run test` to make things clear.

PR #1452, opened 2021-08-12
Fix bug breaking owners bot

rcebulko:
This seems like it should never have worked, and yet it's been running for 2 months. My best guess is that when I added in cache invalidation/refreshing logic, it was broken, but it never had occasion to trigger until a recent OWNERS file change or deletion, and once that code path finally ran it broke.

### DIFF
--- a/owners/src/api/github.js
+++ b/owners/src/api/github.js
@@ -413,7 +413,7 @@ class GitHub {
    * Fetches the contents of a file from GitHub.
    *
    * @param {string} file file ref to fetch.
-   * @return {string} file contents as a string.
+   * @return {{contents: string, sha: string}} file SHA and contents as a string
    */
   async getFileContents(filename) {
     this.logger.info(`Fetching contents of file ${filename}`);
--- a/owners/src/api/github.js
+++ b/owners/src/api/github.js
@@ -418,14 +418,18 @@ class GitHub {
   async getFileContents(filename) {
     this.logger.info(`Fetching contents of file ${filename}`);
 
-    const response = await this.client.repos.getContent({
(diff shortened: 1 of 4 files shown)


owners/src/repo/virtual_repo.js
```diff
@@ -143,11 +143,12 @@ module.exports = class VirtualRepository extends Repository {
     return await this.cache.readFile(repoPath, async () => {
       let contents = '';
       try {
-        contents = await this.github.getFileContents(relativePath);
+        const result = await this.github.getFileContents(relativePath);
```
jridgewell, line 146: Nit: destructure

owners/src/api/github.js
```diff
@@ -418,14 +418,18 @@ class GitHub {
   async getFileContents(filename) {
```
jridgewell, line 418: The `@return` type is incorrect.

owners/src/repo/virtual_repo.js
rcebulko: `contents` needs to exist outside of the try/catch scope. Is it possible to destructure into an already-defined variable?

owners/src/api/github.js
rcebulko: done

owners/src/repo/virtual_repo.js
jridgewell: Technically yes, but it requires you to early declare the `sha` variable too:

```
let sha;
({ contents, sha} = await this.github.getFileContents(relativePath);
```

I find this ugly, so we can skip.

jridgewell (approved):

PR #1454, opened 2021-08-20
[release-calendar] Remove opt-in beta as a channel

estherkim:
Simplifies release calendar to only track % beta as "beta".


rsimha (approved):
Fully agree that this change will make the release calendar easier to consume. 👍

PR #1456, opened 2021-09-01
Wrap filename in backticks to prevent accidental formatting

alanorozco:
It's hard to notice in this screenshot, but the filename includes `__`, which makes a certain section render as bold.

The actual path is:
`.../extensions/__amp-component_name_hyphenated__/__component_version__/storybook/...`

but it looks like:
.../extensions/__amp-component_name_hyphenated__/__component_version__/storybook/...,

image

### DIFF
--- a/owners/src/ownership/owners_check.js
+++ b/owners/src/ownership/owners_check.js
@@ -33,6 +33,17 @@ const CheckRunState = {
 };
 
 /**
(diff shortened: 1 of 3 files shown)


rcebulko (approved):
Love it; I'll redeploy after you merge

PR #1457, opened 2021-09-01
Update project-metrics dashboard to show metrics from CircleCI rather than Travis

rileyajones:
The project metrics dashboard is still showing metrics related to Travis. Because we are no longer using Travis I am removing the code used to show these metrics and reimplementing them to show data from CircleCI.

image


project-metrics/metrics_service/apis/circleci.py
```diff
@@ -0,0 +1,51 @@
...
+    super(CircleCiError, self).__init__(msg)
+
+
+class CircleCiAPI(object):
+  """Interface class for executing queries agains the CircleCI API."""
+  def __init__(self, token = env.get('CIRCLECI_API_ACCESS_TOKEN')):
+    self._token = token
+
+  def _dict_to_params(self, params: Dict):
+    param_string = '&'.join(['%s=%s' % (key, params[key]) for key in params.keys()])
```
rcebulko, line 32:
```suggestion
    param_string = '&'.join([f'{key}={value}' for (key, value) in params.items()])
```

project-metrics/metrics_service/apis/circleci.py
```diff
@@ -0,0 +1,51 @@
...
+    if len(param_string):
```
rcebulko, line 33:
```suggestion
    if param_string:
```

project-metrics/metrics_service/apis/circleci.py
```diff
@@ -0,0 +1,51 @@
...
+      return '?' + param_string
+    return param_string
+
+  def get_workflow_stats(self, reporting_window = models.CircleCiReportingWindow.LAST_90_DAYS) -> models.CircleCiWorkflowStats:
+    params = {
+      'reporting-window': reporting_window.value
+    }
+    endpoint = '%s/%s' % (INSIGHTS_API, 'workflows')
```
rcebulko, line 41: Consider format strings (`f'{INSIGHTS_API}/workflows'`) here and elsewhere

project-metrics/metrics_service/apis/circleci.py
```diff
@@ -0,0 +1,51 @@
...
+    logging.info(endpoint + self._dict_to_params(params))
+    headers = { 'authorization': "Basic %s" % env.get('CIRCLECI_API_ACCESS_TOKEN') }
+    response = requests.get(endpoint + self._dict_to_params(params), headers=headers)
+    parsed = json.loads(response.text)
+    if response.status_code != 200:
```
rcebulko, line 46: Consider using the `HTTPStatus` enum:
```python
if response.status_code == HTTPStatus.OK:
  return models.CircleCiWorkflowStats.from_json(parsed['items'][0])
raise CircleCiError(response.status_code, parsed['message'])

File: project-metrics/metrics_service/cron.yaml [hunk 5]
```diff
@@ -11,10 +11,6 @@ cron:
   url: /_cron/scrape/cherrypick_issues
   schedule: every 72 hours
 
-- description: "Scrape latest builds from Travis"
```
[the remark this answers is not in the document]
rcebulko: I've got some bad news for you then; look up 🙃

project-metrics/metrics_service/database/models.py
```diff
@@ -30,6 +30,84 @@ def _is_last_n_days(
...
+
+class CircleCiWorkflowStats():
+  project_id: str
+  name: str
+  metrics: CircleCiWorkflowMetrics
+  window_start: datetime.datetime
+  window_end: datetime.datetime
+
+  @staticmethod
+  def from_json(dict):
```
rcebulko, line 100: Naming a local variable/arg `dict` shadows the global/keyword `dict()`; maybe call it `json` instead?

project-metrics/metrics_service/database/models.py
```diff
@@ -30,6 +30,84 @@ def _is_last_n_days(
...
+  mttr: int
+  total_credits_used: int
+  failed_runs: int
+  median_credits_used: int
+  success_rate: float
+  total_recoveries: int
+  throughput: float
+
+  @staticmethod
+  def from_json(dict):
```
rcebulko, line 77: see belowe re: naming this `dict`; also, I think static methods' first parameter should always be `cls` (comparable to `self`), but I may be mis-remembering

project-metrics/metrics_service/metrics/__init__.py
```diff
@@ -2,7 +2,7 @@
 from metrics import base
 from metrics import absolute_coverage
 from metrics import cherrypick_issue_count
+from metrics import circleci_flakiness
+from metrics import circleci_greenness
 from metrics import presubmit_latency
-from metrics import release_granularity
-from metrics import travis_greenness
-from metrics import travis_flakiness
+from metrics import release_granularity
```
rcebulko, line 8:
```suggestion
from metrics import release_granularity

```

project-metrics/metrics_service/metrics/base.py
```diff
@@ -1,9 +1,9 @@
 """Base interface for metrics to implement.
 
 To create a new metric:
-1. Subclass `Metric` or `PercentMetric`
+1. Subclass `Metric` or `PercentageMetric`
```
rcebulko, line 4: Nice catch!

project-metrics/metrics_service/metrics/circleci_flakiness.py
```diff
@@ -0,0 +1,24 @@
...
+from apis.circleci import CircleCiAPI
+
+class CircleCiFlakiness(PercentageMetric):
+  def _score_value(self, percentage: float) -> models.MetricScore:
+    if percentage < 0.6:
+      return models.MetricScore.POOR
+    elif percentage < 0.75:
+      return models.MetricScore.MODERATE
+    elif percentage < 0.9:
+
```
rcebulko, line 15:
```suggestion
```

project-metrics/metrics_service/metrics/presubmit_latency.py
```diff
@@ -1,55 +1,27 @@
...
-    Returns:
-      The percentage of passing builds.
-    """
-    session = db.Session()
-    avg_seconds = session.query(sqlalchemy.func.avg(
-        models.Build.duration)).filter(
-            models.Build.is_last_90_days(base_time=self.base_time)).scalar()
-    session.close()
+  def _compute_value(self) -> models.MetricResult:
+    workflow_stats = CircleCiAPI().get_workflow_stats()
```
rcebulko, line 21: FWIW this means each metric will separately be requesting the same stats from the `CircleCiApi`. Unlikely we'd hit rate-limits like this, but something to be aware of. Also, this does not appear to limit the stats to a 90-day window using `self.base_time`. `self.base_time` was added per request from...Raghu? Will? idk, but supporting it allows for back-filling data by some amount of time. I don't know if the CircleCI API supports arbitrary time windows, though. This was also why we were scraping previous build data and computing the stats ourselves; it meant we could always reproduce/recompute the values.

project-metrics/metrics_service/server.py
```diff
@@ -77,8 +77,8 @@ def recompute(metric_cls_name: Text):
   # This header is added to cron requests by GAE, and stripped from any external
   # requests. See
   # https://cloud.google.com/appengine/docs/standard/python3/scheduling-jobs-with-cron-yaml#validating_cron_requests
-  if not flask.request.headers.get('X-Appengine-Cron'):
-    return 'Attempted to access internal endpoint.', status.HTTP_403_FORBIDDEN
+  # if not flask.request.headers.get('X-Appengine-Cron'):
+  #   return 'Attempted to access internal endpoint.', status.HTTP_403_FORBIDDEN
```
rcebulko, line 81:
```suggestion
  if not flask.request.headers.get('X-Appengine-Cron'):
    return 'Attempted to access internal endpoint.', status.HTTP_403_FORBIDDEN
```

rcebulko (commented):
Solid update! Thanks for taking this on. It was my first project on the team, and that really shows 😅 

Last time I tried to update/deploy it, I ran into some hairballs. You may find it easier to first split out a PR that removes all of the Travis metrics, and get yourself set up deploying that update, and then add the new metric(s) separately. I'm unsure if CI in this repo runs the tests for this app.

project-metrics/metrics_service/database/models.py
rileyajones: Fantastic catch

project-metrics/metrics_service/apis/circleci.py
rileyajones: I didn't know about this syntax, that definitely seems better.

project-metrics/metrics_service/metrics/presubmit_latency.py
rileyajones: The function `get_workflow_stats` has a default argument of `last-90-days` so that much will work, however, circleci's api does not currently allow arbitrary time intervals for their insights api. I do not believe that arbirary intervals are a requirement anymore so, while it would be possible to manually compute these values by paginating their workflows api, I don't see that as being necessary. @rsimha

project-metrics/metrics_service/apis/circleci.py
```diff
@@ -0,0 +1,53 @@
...
+  """Interface class for executing queries agains the CircleCI API."""
+  def __init__(self, token = env.get('CIRCLECI_API_ACCESS_TOKEN')):
+    self._token = token
+
+  @staticmethod
+  def _dict_to_params(params: Dict):
+    param_string = '&'.join([f'{key}={value}' for (key, value) in params.items()])
+    if param_string:
+      return '?' + param_string
+    return param_string
```
danielrozenberg, line 37:
```python
    return param_string and '?' + param_string
```
(returns the left side of the `and` if it's false, or the right side if it's true)

project-metrics/metrics_service/apis/circleci.py
```diff
@@ -0,0 +1,53 @@
...
+
+  def get_workflow_stats(self, reporting_window = models.CircleCiReportingWindow.LAST_90_DAYS) -> models.CircleCiWorkflowStats:
+    params = {
+      'reporting-window': reporting_window.value
+    }
+    endpoint = f'{INSIGHTS_API}/workflows'
+    logging.info(endpoint + self._dict_to_params(params))
```
danielrozenberg, line 44: Should probably be something like:
```python
    logging.info('Called %s', endpoint)
```

project-metrics/metrics_service/apis/circleci.py
```diff
@@ -0,0 +1,53 @@
...
+    param_string = '&'.join([f'{key}={value}' for (key, value) in params.items()])
+    if param_string:
+      return '?' + param_string
+    return param_string
+
+  def get_workflow_stats(self, reporting_window = models.CircleCiReportingWindow.LAST_90_DAYS) -> models.CircleCiWorkflowStats:
+    params = {
+      'reporting-window': reporting_window.value
+    }
+    endpoint = f'{INSIGHTS_API}/workflows'
```
danielrozenberg, line 43: Every time you use `endpoint` in this function, it's with `+ self._dict_to_params(params)` - so might as well add it here:
```python
    endpoint = f'{INSIGHTS_API}/workflow{self._dict_to_params(params)}'
```

project-metrics/metrics_service/apis/circleci.py
```diff
@@ -0,0 +1,53 @@
...
+    logging.info(endpoint + self._dict_to_params(params))
+    headers = { 'authorization': 'Basic %s' % env.get('CIRCLECI_API_ACCESS_TOKEN') }
+    response = requests.get(endpoint + self._dict_to_params(params), headers=headers)
+    parsed = json.loads(response.text)
```
danielrozenberg, line 47:
```python
    parsed = response.json()
```

(also drop the `import json` from the top)

danielrozenberg (commented):
This is a pretty big PR, and I had a lot of comments on the Python style of circleci.py - let's work on those and then I'll continue to review the rest

project-metrics/metrics_service/apis/circleci.py
```diff
@@ -0,0 +1,53 @@
+"""Client for querying CircleCI API.
+
+For an overview of the API https://circleci.com/docs/api/v2/
+"""
+import json
+from typing import Dict
+from http import HTTPStatus
+from absl import logging
+import logging
```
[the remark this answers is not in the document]
rileyajones: Good question

project-metrics/metrics_service/apis/circleci.py
```diff
@@ -0,0 +1,53 @@
...
+    super(CircleCiError, self).__init__(msg)
+
+
+class CircleCiAPI(object):
+  """Interface class for executing queries agains the CircleCI API."""
+  def __init__(self, token = env.get('CIRCLECI_API_ACCESS_TOKEN')):
+    self._token = token
+
+  @staticmethod
+  def _dict_to_params(params: Dict):
```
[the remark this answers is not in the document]
rileyajones: I'll just make this a normal method then

project-metrics/metrics_service/apis/circleci.py
```diff
@@ -0,0 +1,53 @@
...
+    param_string = '&'.join([f'{key}={value}' for (key, value) in params.items()])
```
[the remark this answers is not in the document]
rileyajones: That is good to know, my Python is quite rusty

project-metrics/metrics_service/apis/circleci.py
rileyajones: 10/10

project-metrics/metrics_service/apis/circleci.py
rcebulko: I'm usually one to advocate for ternaries & brevity but I find this harder to parse. My suggestion would be instead
```python
return f'?{param_string}' if param_string else param_string
```

rcebulko (commented):
General/high-level note: While I support all of the changes in this PR, I think it would make for a cleaner history and easier rollout/testing/migration etc. if it were split up. ie. modify to support local DB; add CircleCI API + tests; remove Travis metrics + apis; add CircleCI models; add DB migration file to create/delete tables; submit cron-only change; etc. As-written I worry this will be harder to get deployed and all the DB migrations done cleanly, but leave it to your judgement on the best path forward

project-metrics/metrics_service/apis/circleci.py
```diff
@@ -37,7 +37,7 @@ def get_workflow_stats(self, reporting_window = models.CircleCiReportingWindow.L
       'reporting-window': reporting_window.value
     }
     endpoint = f'{INSIGHTS_API}/workflows{_dict_to_params(params)}'
-    logging.info(endpoint)
+    logging.info('Called {endpoint}')
```
rcebulko, line 40:
```suggestion
    logging.info(f'Called {endpoint}')
```

rcebulko (changes requested):

rileyajones:
@rcebulko I'm always a fan of smaller prs and this one has definitely ballooned into more than originally intended. I'll look into refactoring

rileyajones:
Implemented as #1468 #1469 #1470 #1471

PR #1471, opened 2021-09-08
Add cron jobs to compute circleci metrics

rileyajones:
Now that the project-metrics dashboard contains metrics pertaining to circleci it is time to add jobs to start recomputing these metrics on a regular interval.


rcebulko (approved):
In case you're not aware (not sure if you've been unfortunate enough to have had to use Cloud Cron jobs yet) but in the Cron view you can click a Run Now button to test these out immediately

danielrozenberg (approved):

rileyajones:
@rcebulko thanks for the tip on cloud cron jobs, that will be very useful

PR #1557, opened 2021-12-08
Improve usability of `bundle-size-chart`

alanorozco:
- Dark mode.

- Expands chart to fill viewport. 

- Allows zooming in on a time range by clicking and dragging (!)

- Hover indicator dots are clickable. Labels are now nudged when reaching the edge, so they never truncate.

- Sets a hash so that query and time range filters are shareable by URL.

Demo on [canyon-subsequent-sea.glitch.me](https://canyon-subsequent-sea.glitch.me) (data may be stale).

### DIFF
--- a/bundle-size-chart/static/index.html
+++ b/bundle-size-chart/static/index.html
@@ -1,37 +1,227 @@
 <!DOCTYPE html>
 <html lang="en">
-<head>
-  <meta charset="utf-8">
-  <title>AMPHTML bundle-size history</title>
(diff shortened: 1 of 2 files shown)


danielrozenberg (approved):
omg this is awesome! When I created `bundle-size-chart` it was a quick hack to solve a one-time problem, we never assumed it'll become a tool that the team uses on the daily - glad to see someone taking it much further than I did :)

PR #1567, opened 2022-01-27
[Percy Mirror Verifier] Use the fingerprint field when comparing snapshots

danielrozenberg:
From observing the logs, it looks like a snapshot that produces no diffs results in `fingerprint: null`, whereas a snapshot that produces a diff results in `fingerprint: <some-sha1-string>` that is unique to that snapshot:
* If a snapshot results in the exact same pixels it will have the same fingerprint value, which means the PR and the main builds are mirrored
* If one comes with a null fingerprint and the other not-null, then there is a diff
* If they both have a non-null fingerprint but the sha1 values are different, it means the pixels are different - this is a case that wasn't caught using the previous logic - this PR deals with this and makes the code simpler! Woot woot 🥳

### DIFF
--- a/percy-mirror-verifier/src/percy.ts
+++ b/percy-mirror-verifier/src/percy.ts
@@ -21,16 +21,12 @@ export type PercySnapshot = {
   id: string;
   attributes: {
     name: string;
+    fingerprint: string | null;
     'review-state':
       | 'approved'
       | 'unreviewed'
       | 'changes_requested'
       | 'failed'
-      | null;
-    'review-state-reason':
-      | 'no_diffs'
-      | 'auto_approved_branch'
-      | 'user_approved'
       | null;
   };
 };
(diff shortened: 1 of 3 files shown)


percy-mirror-verifier/src/percy.ts
```diff
@@ -21,17 +21,13 @@ export type PercySnapshot = {
...
       | 'approved'
       | 'unreviewed'
       | 'changes_requested'
       | 'failed'
       | null;
-    'review-state-reason':
-      | 'no_diffs'
-      | 'auto_approved_branch'
-      | 'user_approved'
-      | null;
```
rileyajones, line 34: Does this attribute not exist or are you just not using it anymore? I generally like to leave unused attributes in type definitions to help with future refactoring.

rileyajones (approved):

percy-mirror-verifier/src/percy.ts
danielrozenberg: It's just unused. The thing is, the type itself has dozens of fields that aren't used, and I didn't want to "pollute" the type system with the unused one. I can add a comment about how these types are incomplete if you think that makes sense?

PR #1568, opened 2022-02-02
Sort `bundle-size` summary lists

alanorozco:
Sort file lists in `bundle-size` summaries so that they're quicker to read.

1. Groups by filename without extension, so the equivalent `.mjs` file is always next to `.js`
2. Sorts by delta, either `asc` or `desc` depending on pre-approval.

### DIFF
--- a/bundle-size/api.js
+++ b/bundle-size/api.js
@@ -17,7 +17,12 @@
 
 const minimatch = require('minimatch');
 const sleep = require('sleep-promise');
-const {formatBundleSizeDelta, getCheckFromDatabase} = require('./common');
+const {
+  getCheckFromDatabase,
+  formatBundleSizeItem,
+  formatFileItem,
+  sortBundleSizeItems,
+} = require('./common');
 
 const RETRY_MILLIS = 60000;
 const RETRY_TIMES = 60;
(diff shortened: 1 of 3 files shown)


bundle-size/test/common.test.js
```diff
@@ -0,0 +1,94 @@
...
+ * distributed under the License is distributed on an "AS-IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+const {sortBundleSizeItems} = require('../common');
+
+describe('bundle-size common', () => {
+  describe('sortBundleSizeItems', () => {
```
alanorozco, line 20: I added a test for the utility, but I did not change tests for the summary output. 

(The current summary output only has one item per list, and I could not figure out how to add more entries)

bundle-size/common.js
```diff
@@ -49,6 +49,66 @@ exports.getCheckFromDatabase = async (db, headSha) => {
...
+ */
+const noExtension = file => {
+  const parts = file.split('.');
+  parts.pop();
+  return parts.join('.');
+};
+
+/**
+ * @param {{file: string, bundleSizeDelta: number}[]} items
+ * @param {'asc'|'desc'} sizeOrder
```
danielrozenberg, line 85: Nit: I doubt we'll ever really need both options, we can simplify the code by just assuming desc

bundle-size/common.js
```diff
@@ -49,6 +49,66 @@ exports.getCheckFromDatabase = async (db, headSha) => {
...
+
+/**
+ * @param {string} file
+ * @return {string}
+ */
+const noExtension = file => {
+  const parts = file.split('.');
+  parts.pop();
+  return parts.join('.');
+};
```
danielrozenberg, line 81: This is basically "the opposite of `path.extname`", so how about use it in the definition of this function and make it a oneliner:
```suggestion
const noExtension = file => file.slice(0, -path.extname(file).length);
```
(Add `const path = require('path');` at the top, ofc)

Tested this locally to make sure it works, e.g.:
```
> fn = 'a/b/adasds.max.js';
'a/b/adasds.max.js'
> const noExtension = file => file.slice(0, -path.extname(file).length);
undefined
> noExtension(fn);
'a/b/adasds.max'
```

bundle-size/api.js
```diff
@@ -127,12 +132,20 @@ function extraBundleSizesSummary(
...
   }
   if (bundleSizeDeltasAutoApproved.length) {
     output +=
       '\n## Auto-approved bundle size changes\n' +
-      bundleSizeDeltasAutoApproved.join('\n');
+      // If files don't require approval, we're firstly interested in the
+      // smallest delta, so we sort asc
+      sortBundleSizeItems(bundleSizeDeltasAutoApproved, 'asc')
+        .map(formatBundleSizeItem)
+        .join('\n');
```
danielrozenberg, line 148: Eh, I think it's fine to sort both cases as desc - it creates a consistent reading experience 🤷‍♂️

PR #1618, opened 2023-05-24
[webhook pubsub publisher] log entire event

estherkim:


webhook-pubsub-publisher/src/app.ts
```diff
@@ -19,7 +19,7 @@ import type {Probot} from 'probot';
 
 export default (app: Probot): void => {
   app.onAny(async event => {
-    app.log(`Received "${event.name}" event with ID ${event.id}`);
+    app.log(event, 'Received event');
```
danielrozenberg, line 22:
```suggestion
    app.log('Received event', event);
```

(although I'm not sure if `app.log` takes more than one arg?)

estherkim: Yeah I checked, probot uses `pino` so the way I had it was correct

https://getpino.io/#/docs/api?id=loggerinfomergingobject-message-interpolationvalues

danielrozenberg: Looks like it should be something like this then:
```suggestion
    app.log('Received event: %O', event);
```

* https://getpino.io/#/docs/api?id=interpolationvalues-any
* https://nodejs.org/api/util.html#utilformatformat-args

danielrozenberg (approved):

danielrozenberg (approved):

PR #2307, opened 2024-03-19
[bundle-size] Convert to TypeScript

danielrozenberg:
Actually caught a few bugs this way!

Semantic changes:
- Convert to esm code
- Upgraded `probot` to v13
- Database retrieval method now does the type coarsening from string to the relevant types, instead of having the callers do it
- Correctly return from the `/report` endpoint if it sends a 202, without accidentally sending a 500 after the response has closed

Testing changes:
- Changed the signature of `installApiRouter` to make it easier to test without having to raise an entire Probot app (only Express)
- Avoid using `jest.mock`, which has issues with esm modules
- Import JSON fixtures directly instead of doing `fs.readFile`

GitHub's diff algo struggles with connecting the .js and .ts files, I ran [difftastic](https://github.com/Wilfred/difftastic) with `-M10%` to ensure most files get connected correctly and put the result in this zipped HTML file (GitHub won't let me upload an .html file): [difft.zip](https://github.com/ampproject/amp-github-apps/files/14656182/difft.zip)


danielrozenberg:
Tested the happy paths by deploying to gcloud (not promoted to prod)

bundle-size/src/common.ts
```diff
@@ -0,0 +1,87 @@
...
+      'check_run_id',
+      'approving_teams',
+      'report_markdown'
+    )
+    .where('head_sha', headSha);
+
+  return check
+    ? {
+        ...check,
+        check_run_id: Number(check.check_run_id),
```
erwinmombay, line 67: @danielrozenberg is this part how it knows what to coerce to?

danielrozenberg: The data is in a known format, it's just that the database connection returns everything as a string, so I coerce it here to the original data types

erwinmombay (approved):

danielrozenberg:
This version fails on a database connection issue. I reverted to the previous version and am working on a fix

